PR #56, opened 2026-05-18
fix: correct extension entry point path resolution

RNZ01:
To show the correct extension name on pi we need to
Add root index.ts that re-exports the extension with the correct relative path. Update package.json pi.extensions to point to the new root entry file.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,0 +1,1 @@
+export { default } from "./extension/index.ts";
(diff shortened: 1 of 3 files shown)


mfittko (approved):

mfittko:
thanks for spotting this! 👍

Issue #1058, opened 2026-07-01
ci: deduplicate the three Playwright smoke jobs (composite action vs matrix)


mfittko:
## Problem

`.github/workflows/ci.yml` defines three Playwright WebKit smoke jobs — `viewer-smoke`, `deck-smoke`, and `article-smoke` — that are near-identical. They share the same `runs-on: ubuntu-latest`, Node setup, `npm ci`, `PLAYWRIGHT_BROWSERS_PATH` env, the Playwright WebKit cache-restore step, and `npx playwright install --with-deps webkit`. They differ only in (a) the `changes`-based `if:` trigger and (b) the final `npm run test:playwright:*` command. That is copy-pasted setup across three jobs.

## Options considered

- Matrix (`strategy.matrix.include`) over `{name, when, run}`. Collapses the three into one job. Wrinkle: today each job is gated by a job-level `if: needs.changes.outputs.X == 'true'`, so a non-triggered smoke job costs zero runner time. A matrix leg cannot gate its own spawn on the per-leg condition, so every leg starts and you gate the expensive step instead — meaning skipped legs still pay checkout + `npm ci` + Playwright install. For the common single-file PR this is slower, not faster.
- Composite action (`.github/actions/playwright-webkit/action.yml`) for the shared setup, keeping the three jobs and their zero-cost skips. Removes the duplication without regressing the skip-to-zero behavior.

## Recommendation

Prefer the composite-action extraction over a matrix: it deduplicates the shared setup while preserving the current zero-cost skip semantics. Only choose a matrix if multiple smoke categories are expected to fire together often enough that shared parallel scheduling outweighs the always-on setup cost on skipped legs.

## Acceptance criteria

- [x] The shared smoke-job setup (checkout, Node setup, `npm ci`, Playwright WebKit cache-restore, `npx playwright install --with-deps webkit`, `PLAYWRIGHT_BROWSERS_PATH`) is defined once, not copy-pasted across `viewer-smoke` / `deck-smoke` / `article-smoke`.
- [x] A smoke category that is NOT triggered by the `changes` filter still incurs no wasted CI cost (zero-cost skip preserved), OR the PR explicitly documents and justifies the trade-off if a matrix is chosen instead.
- [x] The three smoke commands (`test:playwright:viewer`; `intro-deck`+`deep-dive`; `intro-article`+`deep-dive-article`) and their existing `changes`-output triggers behave identically to today.
[...]

Issue #1056, opened 2026-07-01
refactor(test): consolidate per-slice Playwright configs into one projects-based (registry-driven) config


mfittko:
## Problem

There is one `playwright.*.config.mjs` per UI slice (intro-article, deep-dive-article, intro-deck, dev-loops-deep-dive, inspect-run-viewer). All are byte-for-byte identical except two values — `sliceId` and `testMatch` — and `testMatch` is just `${sliceId}.spec.mjs`. Every new article/deck/viewer forces another copy of the same boilerplate. The count only grows.

## Proposed change

Collapse to a single `playwright.config.mjs` using Playwright's native **projects** model — one project per slice (`name` + `testMatch`), run via `playwright test --project=<slice>`. `createWebkitSmokeConfig` already sets per-slice `outputDir`/reporter folders; move those to per-project settings.

Ideally **generate the projects from the existing registries** (`ARTICLE_REGISTRY` / `DECK_REGISTRY` / `VIEWER_REGISTRY` in `test/playwright/harness/*`) — the same source `ui-e2e-scoping` reads — so adding a UI artifact needs only a registry entry (already required by the scoping gate) and zero new config file.

## Scope / touches

- Delete the 5 `playwright.*.config.mjs` files; add one `playwright.config.mjs` with per-slice projects.
- `package.json`: `test:playwright:<slice>` scripts become `playwright test --project=<slice>`.
- `.github/workflows/ci.yml`: the `article-smoke` / `deck-smoke` / `viewer-smoke` job commands update to the `--project` form. `UI_E2E_CHECK_NAMES` (job names) and the ui-e2e-scoping contract are unaffected.

## Acceptance criteria

- [x] A single `playwright.config.mjs` defines one project per registered slice; the 5 per-slice config files are removed.
- [x] `npx playwright test --project=intro-article` (etc.) runs exactly the slice's spec; per-slice artifact output is preserved (distinct `outputDir`).
- [x] CI `article-smoke` / `deck-smoke` / `viewer-smoke` run the same specs as before and stay green; gate `UI_E2E_CHECK_NAMES` still resolve.
- [x] Adding a new artifact requires only a registry entry — no new config file (verify by adding a throwaway registry entry in a test, or document the path).

## Non-goals

- Changing the assertions/harness behavior; multi-browser expansion. Purely config-topology consolidation.

Issue #1043, opened 2026-07-01
feat(gates): wire lightMode to auto-bypass full fanout for micro-PRs


mfittko:
## Problem

\`.devloops\` has a \`lightMode\` config block that isn't wired to gate dispatch:

\`\`\`yaml
localImplementation:
  lightMode:
    enabled: true
    maxFiles: 2
    maxLines: 100   # too permissive — see below
\`\`\`

Today every PR — including 4-line CSS selector renames — runs the full 15-angle draft_gate fanout + Copilot polling + 14-angle pre_approval_gate fanout.

## Proposal

When a PR falls under the lightMode threshold, collapse both gates to a single inline correctness + no-op check. Draft gate still runs (mandatory for the draft→ready boundary) but as a single-agent inline pass, not a 15-agent fanout. Copilot polling is skipped. pre_approval_gate fanout is replaced with the same inline check.

**Escalation paths (two mechanisms):**

1. **Auto-escalate:** if the inline check finds anything severity ≥ \`worth-fixing-now\`, it escalates to full fanout automatically. Light mode can't suppress real findings.
2. **Label-based override:** adding the \`gate:full\` label to any PR forces full fanout regardless of size. Maintainer-controlled escape hatch for cases where a small diff touches a sensitive surface.

## Threshold

100 LoC default is too permissive (a non-trivial refactor can be under 100 lines). Suggested: **≤ 20 lines changed AND ≤ 3 files**. Keeps it firmly in "typo/selector rename/config tweak" territory.

The existing \`maxLines\` / \`maxFiles\` keys in \`.devloops\` can carry the new values — no schema change needed.

## Acceptance criteria

- [x] Gate dispatch reads \`lightMode.enabled\`, \`lightMode.maxLines\`, \`lightMode.maxFiles\`
- [x] PR under threshold (and no \`gate:full\` label) → inline single-agent correctness+no-op check; full fanout skipped
- [x] If inline check finds severity ≥ \`worth-fixing-now\` → escalates to full fanout
- [x] PR with \`gate:full\` label → always runs full fanout regardless of size
- [x] Copilot review request + polling skipped for light-mode PRs
- [x] \`draft_gate\` boundary (draft→ready) still recorded (required by \`requireDraftFirst\`)
- [x] \`.devloops\` \`maxLines\` updated to 20, \`maxFiles\` stays 2 (or 3)
- [x] Existing full-fanout path unchanged for PRs over threshold
[...]

PR #1059, opened 2026-07-01
chore(config): pin queue.projectNumber to skip board-title lookup

mfittko:
## Summary

Pins `queue.projectNumber: 3` in `.devloops` so queue commands resolve the GitHub Projects V2 board by number instead of by title.

## Why

`queue add` / `sync-status` / `archive-done` resolve the board from `.devloops`. With only `boardTitle` set, they must map the title to a project number via a `gh project list` lookup. `queue.projectNumber` is a first-class, schema-validated config key that the resolvers prefer over `boardTitle`, skipping that lookup. Surfaced during PR #1055's session when enqueueing a follow-up required resolving the board number by hand.

## Change

- `.devloops`: add `queue.projectNumber: 3` (board "dev-loops Queue"). `boardTitle` stays as the human-readable fallback / bootstrap title.

## Validation

Config loads with zero validation errors; `queue.projectNumber` resolves to 3 and `boardTitle` unchanged (`loadDevLoopConfig()`).

## Non-goals

No behavior change to queue commands beyond the resolution shortcut; no other config keys touched.

### DIFF
--- a/.devloops
+++ b/.devloops
@@ -90,6 +90,10 @@ localImplementation:
     maxLines: 100
 
 queue:
+  # projectNumber pins the GitHub Projects V2 board by number so queue commands
+  # skip the boardTitle -> number lookup (a gh project list call). boardTitle is
+  # kept as a human-readable fallback / for board bootstrap.
+  projectNumber: 3
   boardTitle: "dev-loops Queue"
   maxParallel: 3
   maxAutoFiledIssues: 10


mfittko:
Closing per maintainer decision: a standalone full-gate pipeline for a 4-line config change is disproportionate (cf. #1043). The change (`.devloops queue.projectNumber: 3`) is preserved on branch `chore/queue-project-number` and will be folded into the v0.7 queue/CI/config DX work (#1058 / #1056), where the gate cost is amortized. Not lost — just re-vehicled.

PR #1489, opened 2026-07-29
docs(ui-review): axe install is an explicit opt-in; stat trailing-slash files entries in the escaping-imports contract

### LINKED ISSUE #1471: ui-review prerequisites snippet installs the optional axe runner as if required
The Prerequisites section in `skills/docs/ui-review-recipe-contract.md` (mirrored in `skills/ui-review/SKILL.md`) shows:

```
npm install --save-dev @playwright/test @axe-core/playwright
```

while the surrounding prose states that `@axe-core/playwright` is optional and that without it every `axe.json` is a deterministic JSON `null` with the rest of the review still working. Installing both in one command implies axe is required to run the stages. Raised by Copilot on PR #1460 (four of its five comments made this point); deliberately not fixed there to preserve convergence at the round cap.

## Decided approach

Split the snippet: the default install is `@playwright/test` + `npx playwright install webkit`; the axe install is shown separately as an explicit opt-in for computed-accessibility artifacts. Apply to both doc surfaces and regenerate the `.claude/` mirrors.

Folded in while the area is open: `resolveShippedDirs` in `test/contracts/no-package-escaping-imports.test.mjs` takes a trailing-slash `files` entry as a directory without a `stat`, so an entry that ends with `/` but is a file on disk surfaces as a less-clear walk error later. Stat those too, still treating ENOENT as an absent-dir pass.

## Acceptance criteria

- [x] Both doc surfaces show the required install (`@playwright/test` + webkit) separate from the optional axe install, with the opt-in framing matching the prose.
- [x] `.claude/` mirrors regenerated and in sync (doc-guard contract green).
- [x] `resolveShippedDirs` stats trailing-slash `files` entries; a file-on-disk entry produces a clear failure, ENOENT still passes as absent dir.

## Definition of done

- [x] Contract tests (doc-guard + escaping-imports) green.
- [x] Full verify suite green.

## Non-goals

- Any behavior change to the axe-absent `null` fallback itself.
- Broader restructuring of the prerequisites docs.

mfittko:
The Prerequisites snippet in `skills/docs/ui-review-recipe-contract.md` (mirrored in `skills/ui-review/SKILL.md`) installed `@playwright/test` and `@axe-core/playwright` in one command, implying axe is required — while the surrounding prose states axe is optional and an axe-less run emits a deterministic JSON `null` for every `axe.json` with the rest of the review working. The snippet is now split: the default install is `@playwright/test` + `npx playwright install webkit`, with the axe install shown separately as an explicit opt-in for computed-accessibility artifacts. Both doc surfaces updated, `.claude/` mirrors regenerated.

Folded in per the issue: `resolveShippedDirs` in `test/contracts/no-package-escaping-imports.test.mjs` now stats trailing-slash `files` entries — an entry ending with `/` that is a file on disk fails with a clear message naming the entry, instead of a later raw `ENOTDIR` from the dir-exists probe; ENOENT still passes as an absent dir.

## Validation

- doc-guard: 216/216; escaping-imports contract: 11/11; full verify: 0 failures.

Closes #1471

### DIFF
--- a/.claude/skills/docs/ui-review-recipe-contract.md
+++ b/.claude/skills/docs/ui-review-recipe-contract.md
@@ -23,7 +23,7 @@ Playwright is an **optional peer dependency** of `dev-loops`: a repo that never
 runs a UI review carries none of its weight, so nothing installs it for you.
 
 ```
-npm install --save-dev @playwright/test @axe-core/playwright
+npm install --save-dev @playwright/test
 npx playwright install webkit
 ```
 
@@ -31,9 +31,17 @@ Stage 2 (`ui-review-drive`) and the loop-grill's `visual-grill-capture` stage
 launch headless WebKit through it. With either the package or the browser binary
 missing, both stages **stop with the install command as their stated reason** and
 report no failures — a setup gap on your machine is never charged to the PR under
-review. `@axe-core/playwright` is optional in the same way and drives only the
-computed-accessibility artifact: without it every `axe.json` is a deterministic
-JSON `null` and the rest of the review still works.
+review.
+
+`@axe-core/playwright` is a separate opt-in, only needed for the
+computed-accessibility artifact:
+
+```
+npm install --save-dev @axe-core/playwright
+```
+
+Without it every `axe.json` is a deterministic JSON `null` and the rest of the
+review still works.
 
 ## Where the recipe lives
 
--- a/.claude/skills/ui-review/SKILL.md
+++ b/.claude/skills/ui-review/SKILL.md
@@ -34,6 +34,7 @@ nothing downstream and a local setup gap cannot be charged to the PR. Other
 stopped results are not alike — the missing-recipe stop carries a `must-fix` that
 is only surfaced by threading it onward, so read `failures`, not `stopped`, when
 deciding what to report.
+
 `@axe-core/playwright` is a separate opt-in
 (`npm install --save-dev @axe-core/playwright`) that drives only the
 computed-a11y artifact: without it every `axe.json` is a deterministic JSON
--- a/.claude/skills/ui-review/SKILL.md
+++ b/.claude/skills/ui-review/SKILL.md
@@ -25,7 +25,7 @@ The two browser-driving stages (`ui-review-drive`, and the `visual-grill-capture
 stage the loop-grill uses) launch headless WebKit through Playwright, which is an
 **optional** peer dependency: a consumer who never runs a UI review carries none
 of its weight. Where one is run, install it once —
-`npm install --save-dev @playwright/test @axe-core/playwright`, then
+`npm install --save-dev @playwright/test`, then
 `npx playwright install webkit`. When either the package or the browser binary is
 missing, both stages stop with those instructions as the stop reason and carry no
 failure entries. That is deliberate: `ui-review-diagnose` turns every failure into
@@ -34,10 +34,11 @@ nothing downstream and a local setup gap cannot be charged to the PR. Other
 stopped results are not alike — the missing-recipe stop carries a `must-fix` that
 is only surfaced by threading it onward, so read `failures`, not `stopped`, when
 deciding what to report.
-`@axe-core/playwright` is optional in the same
-way and drives only the computed-a11y artifact: without it every `axe.json` is a
-deterministic JSON `null`, so a11y findings lose their grounding while the rest of
-the review still works.
+`@axe-core/playwright` is a separate opt-in
+(`npm install --save-dev @axe-core/playwright`) that drives only the
+computed-a11y artifact: without it every `axe.json` is a deterministic JSON
+`null`, so a11y findings lose their grounding while the rest of the review still
+works.
 
 The route's handoff envelope carries its stop rules and acceptance
 self-validation (defined in `handoff-envelope.mjs`): no product-code writes,
--- a/skills/docs/ui-review-recipe-contract.md
+++ b/skills/docs/ui-review-recipe-contract.md
@@ -23,7 +23,7 @@ Playwright is an **optional peer dependency** of `dev-loops`: a repo that never
 runs a UI review carries none of its weight, so nothing installs it for you.
 
 ```
-npm install --save-dev @playwright/test @axe-core/playwright
+npm install --save-dev @playwright/test
 npx playwright install webkit
 ```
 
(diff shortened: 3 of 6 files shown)


mfittko:
### Gate fan-out findings: draft_gate

Reviewed head: bc3fba2c0cb54a24e37c79d730f61cf91bb60bcd

#### Worth fixing now (3)
- `contract-surface`: The new axe opt-in sentence starts on line 40 with no blank line after line 39, so it renders as the tail of the preceding paragraph about `stopped` vs `failures` semantics. The sibling surface (skills/docs/ui-review-recipe-contract.md:36-44) got a clean blank-line-separated paragraph plus its own f — _accepted-for-fix_
- `contract-surface`: The new trailing-slash failure message states the problem but not the fix: `... has a trailing slash but is a file on disk, not a directory`. The sibling guard message in the same function (line 72, glob entries) ends with an explicit remedy (`teach it the pattern or list the directory explicitly`). — _accepted-for-fix_
- `input-validation`: The new trailing-slash validation `throw` is raised INSIDE the `try` whose `catch` (line 90-93) is an errno filter for `stat`. It escapes only incidentally, because a plain `Error` has no `.code` and therefore fails the `err.code !== "ENOENT"` test. Verified live: it does propagate today and the new — _accepted-for-fix_

#### Defer (7)
- `contract-surface`: The doc install snippet and the runtime guidance (scripts/loop/ui-review-capture.mjs:24 PLAYWRIGHT_MISSING_MESSAGE) agree exactly at this head, but nothing enforces that. test/docs/ui-review-recipe-doc.test.mjs guards only the config-key reference against the schema, not the install command against — _deferred_
- `correctness`: The deliberate trailing-slash-is-a-file Error is thrown INSIDE the try and lands in that same catch; it only escapes because a plain Error has no .code, so `err.code !== "ENOENT"` happens to be true. It works and is test-covered, but the author-error signal is routed through the ENOENT-swallowing ha — _deferred_
- `coverage`: The install snippet itself has no automated pin. doc-guard covers the .claude mirror SYNC of these surfaces (claude-assets-reproducible.test.mjs:84 byte-reproducibility via checkAssets, which includes .claude/skills/docs/ui-review-recipe-contract.md and .claude/skills/ui-review/SKILL.md), and test/d — _deferred_
- `coverage`: Separator asymmetry, untested: the normalization at line 76 strips a trailing `[\\/]+`, i.e. it accepts a backslash-terminated entry, but the new author-error check uses `entry.endsWith("/")` only. A hypothetical `"README.md\\"` entry would skip the new failure and be silently dropped as a plain fil — _deferred_
- `determinism`: The new trailing-slash-on-a-file throw is decided by stat(), so on a case-insensitive filesystem (macOS APFS, Windows NTFS) a MISCASED trailing-slash entry naming a file resolves and throws, while on a case-sensitive filesystem (Linux ext4 CI) the same entry hits ENOENT and is silently accepted as a — _deferred_
- `input-validation`: Asymmetry between the separator set used to normalize the name and the one used to detect the directory claim: line 76 strips `/[\\/]+$/` (both separators) but lines 87 and 92 test `entry.endsWith("/")` (forward slash only). A backslash-terminated entry therefore has its separator stripped for the s — _deferred_
- `input-validation`: Non-ENOENT stat errnos rethrow as bare OS errors with no `files`-entry or contract context, unlike both sibling validation errors in this function (lines 72 and 88), which name the offending entry and — for the glob case — a remedy. Verified: a permission-denied entry surfaces as `EACCES: permission — _deferred_

mfittko:
### Gate review: `draft_gate`

**Reviewed head SHA:** `440973356a266dd208e1a4344781ad6450804b70`
**Verdict:** clean

**Findings summary:** Draft fan-out (9 angles) converged over 2 rounds: r1 4 clean + 3 w-f-n (errno-filter coupling, remedy-less error, glued SKILL paragraph) — all fixed and re-verified structurally closed; r2 clean. 5 defers in ledger. doc-guard 216, escaping-imports 11, links 588 OK; full verify 0 fail.

**Next action:** Mark ready for review and request Copilot review

mfittko:
### Gate fan-out findings: pre_approval_gate

Reviewed head: 440973356a266dd208e1a4344781ad6450804b70

#### Worth fixing now (3)
- `contradiction-lens`: Both new comments (and the PR body) name walk() as the pre-fix failure site, but the referenced code in the same file never reaches walk(). findEscapingImports stats `new URL(`${dir}/`, repoRootUrl)` as a dir-exists probe at line 133 BEFORE walking; for a trailing-slash entry that is a file on disk — _accepted-for-fix_
- `contradiction-lens`: The file now asserts two mutually exclusive models of what a trailing slash MEANS, and uses each to justify opposite behavior. Lines 59 and 77-78 assert npm treats 'scripts' and 'scripts/' identically — i.e. the slash carries no semantic, which is the stated reason to classify by disk. The new sibli — _accepted-for-fix_
- `dry`: The install command is hand-maintained in four surfaces with no pin to its runtime source of truth. PLAYWRIGHT_MISSING_MESSAGE (scripts/loop/ui-review-capture.mjs:24) is what the stages actually emit as their stop reason; the same command text is restated at skills/docs/ui-review-recipe-contract.md: — _accepted-for-fix_

#### Defer (7)
- `contradiction-lens`: The test titled 'resolveShippedDirs classifies a files entry by what it is on disk, not by a trailing slash' asserts the exact opposite for one input: `absent/` is not on disk, and lines 345/348 expect it in the result purely because of its trailing slash. The change reinforces this at line 93 with — _deferred_
- `contradiction-lens`: This PR moves axe from the default install to an opt-in, which makes an all-`null` axe.json the DEFAULT consumer path rather than an edge case — but the vision template has no null carve-out for axe. Line 17 says axePath 'must point to axe.json', line 31 says 'Fail closed when required inputs are mi — _deferred_
- `docs`: The newly added comment justifies the trailing-slash throw with a premise that argues against it: "npm treats \"scripts\" and \"scripts/\" as the same declaration, but a trailing slash on an entry that is actually a FILE on disk is an author error". If the two forms were truly identical to npm, "REA — _deferred_
- `dry`: The two changed doc surfaces each restate the whole prerequisites fact — optional-peer framing, both install commands, and the axe-null fallback — and neither references the other. skills/docs/ui-review-recipe-contract.md links out to ui-artifact-contract.md (line 352) but there is no link in either — _deferred_
- `dry`: The new trailing-slash-is-a-file test builds a second temp fixture whose only content is an identical `README.md`, duplicating the fixture the adjacent test at line 334 already creates for the same purpose (that test writes README.md at line 340 to cover the no-slash file case). The mkdtemp/try/fina — _deferred_
- `yagni`: The justifying comment overstates the guard's payoff: it claims the alternative is "letting it walk() as a bogus dir later", but the prior code never reached walk(). findEscapingImports stats the dir first (line 133) and rethrows anything that is not ENOENT, so a trailing-slash file entry already fa — _deferred_
- `yagni`: The last two comment lines justify the try/catch restructure by a hypothetical future change ("can never be coupled to (or later swallowed by) a widened errno filter"). Rationale for a widening that has not happened is speculative; the thrown Error carries no .code, so it is rethrown by the existing — _deferred_

mfittko:

**Reviewed head SHA:** `dabccc6b6dedc0d1c23c20a47054bdf8871b6d3b`
**Verdict:** clean

**Findings summary:** PA fan-out (7 angles) converged over 2 rounds + a revert-proofing pass at head dabccc6b. r1: 4 clean; dry drove the runtime-message pin, contradiction-lens fixed the failure-site comment + one trailing-slash model; r2 caught the pin's revert-vacuity — now anchored to the exact default-install line and PROVEN to fail on the pre-fix doc. Copilot 2 rounds clean, zero threads. Full verify 0 fail.

**Gate evidence note:** Copilot review rounds exhausted (2/2); current head has zero unresolved threads and green or credibly green CI, so pre_approval_gate fallback is allowed without another Copilot re-request.

**Next action:** All gates green: merge (operator pre-authorized merge on gates passed)

PR #1500, opened 2026-07-30
fix(package): declare zod and yaml; guard shipped bare imports against the root manifest

### LINKED ISSUE #1470: Shipped scripts import zod/yaml undeclared by the root package; bare-specifier escapes have no contract check
`test/contracts/no-package-escaping-imports.test.mjs` catches one of the two ways a shipped file can fail to resolve from the published tarball — a **relative** specifier leaving the shipped `files` set. The sibling class, a **bare** specifier naming a package that is not in the published dependency contract, has no static check.

## The gap

Three shipped root scripts import packages the root `package.json` does not declare:

| File | Bare import |
|---|---|
| `scripts/generate-config-schema.mjs` | `zod` |
| `scripts/loop/detect-internal-only-pr.mjs` | `yaml` |
| `scripts/pages/render-article.mjs` | `yaml` |

The root package declares only `@dev-loops/core` as a dependency; `zod` and `yaml` are `@dev-loops/core`'s dependencies. They resolve today only because npm **hoists** transitive dependencies into the install root — an implicit layout guarantee that breaks under pnpm's default strict layout, Yarn PnP, `npm install --install-strategy=nested`, or a version conflict that forces nesting. Latent, not active — verified working in a real consumer install from both tarballs.

## Decided approach

1. **Declare what is used.** Add `zod` and `yaml` to the root package's `dependencies` — they are genuinely runtime deps of shipped scripts. (Rejected: routing through `@dev-loops/core` re-exports to preserve a single-dependency surface — indirection for aesthetics.)
2. **Guard the class.** Extend the escaping-imports contract to bare specifiers: match real import statements, resolve each to its package name, and assert membership in `dependencies` / `peerDependencies` / Node builtins. Skip list for vendored/minified files (e.g. `scripts/loop/inspect-run-viewer/vendor/`). The matcher must not be a PERMISSIVE regex — a false hit on the same words inside a string or a minified bundle makes the check unusable. An AST or `node:module` resolver satisfies that; so does TIGHTENING the existing shared matcher so prose cannot bridge into a quoted string, which is what shipped (the keyword→`from` span accepts only binding-list characters, measured at one specifier changed across the shipped set: the false positive). Adding a parser dependency to fix an undeclared-dependency defect would be self-undermining; the residual is a bounded false negative, documented at the regex with an upgrade path that forbids re-widening the span.

## Acceptance criteria

- [x] Root `package.json` declares `zod` and `yaml` with versions compatible with `@dev-loops/core`'s ranges.
- [x] The extended contract test fails when a shipped file gains a bare import outside `dependencies`/`peerDependencies`/builtins, and passes on the current tree.
- [x] No false positives on the known tricky files: `generate-claude-assets.mjs`, `validate-state-machine-conformance.mjs`, `build-adjacent-bundle.mjs`, the mermaid vendor bundle, `extract-changelog-section.mjs`.

## Definition of done

- [x] Contract test runs in the standard verify/CI suite.
- [x] `npm install` from the packed tarball still resolves all three scripts (spot-check `detect-internal-only-pr.mjs --help`).
- [x] Full verify suite green.

## Non-goals

- Supporting Yarn PnP/pnpm layouts end-to-end — the fix removes the hoisting dependency; full alternative-installer CI is out of scope.
- Extending the guard to the `@dev-loops/core` package's own imports (its deps are declared; add later if drift appears).

mfittko:
Three shipped root scripts bare-import packages the root manifest never declares — `scripts/generate-config-schema.mjs` imports `zod`; `scripts/loop/detect-internal-only-pr.mjs` and `scripts/pages/render-article.mjs` import `yaml`. They resolve today only because npm hoists `@dev-loops/core`'s own dependencies into the install root, an implicit layout guarantee that breaks under pnpm's strict layout, Yarn PnP, `--install-strategy=nested`, or any version conflict that forces nesting. Latent rather than active, and the third resolution-failure mode beside the two the escaping-imports contract already covers (a relative specifier leaving the shipped set; a bare specifier naming one of this repo's own top-level dirs).

- Root `package.json` declares `yaml: ^2.9.0` and `zod: ^4.4.3`, matching `@dev-loops/core`'s own ranges so a single hoisted copy still satisfies both.
- New contract: every static bare import in the shipped set must resolve to a package in `dependencies`/`peerDependencies`, or be a Node builtin. It reuses the shipped-set walker and the static-import matcher the optional-peer guard already uses, so an offender cannot arrive through a form one of them misses.
- The shared matcher's span between the keyword and `from` now accepts only **binding-list characters** (identifiers, braces, commas, `*`, whitespace). It still crosses newlines, because `import {\n a,\n b\n} from "pkg"` is the dominant multi-name style here and the shape a new dependency arrives in — but excluding parens, slashes and backticks ends the pairing of an `export` line with a quoted string far below it, which is how a thrown template literal (`scripts/release/assert-core-dependency-version.mjs:46`) read as a specifier. This is a tightening of the existing matcher, not the widened regex the issue's approach warned against, and it keeps ONE matcher shared with the optional-peer guard so the two cannot disagree about what an import is. Contents are BOM-stripped before scanning, since the `^` anchor cannot skip U+FEFF. The class is ASCII-only (`\w`), so a non-ASCII binding identifier or a comment inside a binding list is a bounded false negative — the upgrade path is to widen the class (`u` flag, `[\p{L}\p{N}_$*,{}\s]`), never the span back to `[^'"]*?`. Separately, the `^` anchor means only the first import on a physical line is seen, so a compact `import{a}from"x";import{b}from"y";` under-reports; no such line exists outside the skipped vendor bundle.
- Vendored bundles (`**/vendor/**`) are skipped: they inline their dependencies, so their import-looking text is bundled source, not this package's resolution surface.
- The detector is an exported pure function returning `{ offenders, declaredSeen }`, driven by a synthetic-fixture positive control (single-line, multi-line and BOM-prefixed offenders, declared packages, `node:` and bare builtins, a relative specifier, a vendored bundle) — an assert-empty test over the real tree cannot tell "nothing undeclared" from "the walk found nothing".

## Validation

- Guard proven discriminating, not vacuous: the real-tree test asserts `declaredSeen > 0` (263 declared bare imports resolved today), so a broken matcher or walk fails loudly instead of passing green; dropping `yaml` from the manifest flags exactly `detect-internal-only-pr.mjs` and `render-article.mjs`, dropping `zod` flags `generate-config-schema.mjs`.
- No false positives on the tricky files the issue names (`generate-claude-assets.mjs`, `validate-state-machine-conformance.mjs`, `build-adjacent-bundle.mjs`, the mermaid vendor bundle, `extract-changelog-section.mjs`).
- Multi-line `import`/`export … from` binding lists, `* as` forms, the minified form, and the prose-pairing rejection are all pinned in the matcher's own test.
- `no-package-escaping-imports` 16/16; `detect-internal-only-pr.mjs --help` resolves; full verify 0 failures.

Closes #1470

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,9 @@
         "packages/*"
       ],
       "dependencies": {
-        "@dev-loops/core": "^1.0.0-rc.1"
+        "@dev-loops/core": "^1.0.0-rc.1",
+        "yaml": "^2.9.0",
+        "zod": "^4.4.3"
       },
       "bin": {
         "dev-loops": "cli/index.mjs"
--- a/package.json
+++ b/package.json
@@ -93,7 +93,9 @@
     ]
   },
   "dependencies": {
-    "@dev-loops/core": "^1.0.0-rc.1"
+    "@dev-loops/core": "^1.0.0-rc.1",
+    "yaml": "^2.9.0",
+    "zod": "^4.4.3"
   },
   "repository": {
     "type": "git",
--- a/test/contracts/no-package-escaping-imports.test.mjs
+++ b/test/contracts/no-package-escaping-imports.test.mjs
@@ -231,16 +231,11 @@ export async function findEscapingImports({ repoRootUrl, shippedDirs, allowDirs 
 }
 
 // ponytail: relative specifiers only. Bare-specifier escapes are the sibling
-// defect class, answered below by `findUndeclaredBareImports`. The objection
-// that used to stand here — that the EXHAUSTIVE question ("is every bare
-// specifier a declared dependency?") needs real parsing, because a false hit on
-// the same words inside a string or a minified bundle makes the check unusable
-// — was aimed at a PERMISSIVE matcher. Both failure modes it named are closed
-// structurally rather than by parsing: the keyword→`from` span accepts only
-// binding-list characters, so prose can no longer bridge into a quoted string,
-// and vendored bundles are skipped by path. What remains is a bounded
-// false-NEGATIVE (a binding list containing a comment), which leaves the check
-// usable — the direction a guard can afford to be wrong in.
+// defect class, answered below by `findUndeclaredBareImports` — without a
+// parser, because the two false-hit modes that would need one (prose bridging
+// into a quoted string; a minified vendor bundle) are closed by the matcher's
+// binding-list span and the vendor path skip. The residual is a bounded false
+// NEGATIVE, the direction a guard can afford to be wrong in.
 async function rootPackageOffenders() {
   const repoRootUrl = new URL("../../", import.meta.url);
   const pkg = JSON.parse(await readFile(new URL("package.json", repoRootUrl), "utf8"));
@@ -386,8 +381,7 @@ export async function findUndeclaredBareImports({ repoRootUrl, shippedDirs, decl
       // text is bundled source, not this package's resolution surface.
       if (VENDORED_PATH_RE.test(relative)) continue;
 
-      // `^[ \t]*` cannot skip U+FEFF, so a BOM-saved file whose FIRST line is an
-      // import would be scanned as if that import were absent.
+      // BOM-stripped for the same reason as the peer guard above.
       const contents = (await readFile(fileUrl, "utf8")).replace(/^\uFEFF/, "");
       for (const match of contents.matchAll(STATIC_SPECIFIER_RE)) {
         const specifier = match[1];
--- a/test/contracts/no-package-escaping-imports.test.mjs
+++ b/test/contracts/no-package-escaping-imports.test.mjs
@@ -12,6 +12,7 @@ import { tmpdir } from "node:os";
 import path from "node:path";
 import test from "node:test";
 import { fileURLToPath, pathToFileURL } from "node:url";
+import { isBuiltin } from "node:module";
 
 // Static import/export-from, dynamic import(), and side-effect import relative specifiers.
 const RELATIVE_SPECIFIER_RE = /(?:from\s*|import\s*\(\s*|import\s+)["'](\.\.?\/[^"']+)["']/g;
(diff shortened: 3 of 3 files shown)


mfittko:
### Gate review: `draft_gate`

**Reviewed head SHA:** `4c3100653c40de3957017bd1bf3159955bb50f91`
**Verdict:** clean

**Findings summary:** 8 angles reviewed; 3 findings (see per-angle breakdown below).

- `contract-surface` → clean
- `correctness` → findings_present
  - [defer] One bounded over-tightening introduced by the new character class: JS `\w` is ASCII-only without the `u` flag, so a binding list containing a non-ASCII identifier — `import {\n café,\n} from "pkg"` — matches under the old permissive span but NOT u … (`test/contracts/no-package-escaping-imports.test.mjs:292`) — _deferred_
- `coverage` → findings_present
  - [defer] Unchanged from round 3 and explicitly allowed to stay deferred: nothing pins that the guard is WIRED into the verify suite, only that it exists. Removing ` test/contracts/*.test.mjs` from the `test:assets` script leaves the contracts suite delta-id … (`package.json:0`) — _deferred_
- `determinism` → clean
- `gate-evidence` → clean
- `input-validation` → clean
- `pr-description` → clean
- `scope` → findings_present
  - [defer] Carried from r3, unchanged, and now the only inaccurate record left: the deviation note on issue 1470 (2026-07-30T14:33:48Z, still the latest comment) states the guard 'scans line-wise rather than over whole file contents' and that 'every one in th … (`test/contracts/no-package-escaping-imports.test.mjs:283`) — _deferred_

**Next action:** Mark ready for review and request Copilot review

mfittko:

**Reviewed head SHA:** `5ee48b479717712bfc8508bedad372bd7121667f`
**Verdict:** clean

**Findings summary:** 4 angles reviewed; 4 findings (see per-angle breakdown below).

- `acceptance-criteria` → clean
- `contradiction-lens` → findings_present
  - [defer] Recurring gate-harness input gap (not a diff defect; tracked as issue 1496, eighth consecutive round across PRs 1490/1491/1492/1498/1500): the bundle again carries prBody null and scope.acceptanceCriteria null, so the briefing prefix asserts "(no P … (`tmp/gate-context/mfittko-dev-loops/pr-1500/pre_approval_gate-5ee48b479717712bfc8508bedad372bd7121667f.json`) — _deferred_
- `yagni` → findings_present
  - [defer] UNCHANGED at this head (delta was comment-only). Dead branch: `const declaredSet = declared instanceof Set ? declared : new Set(declared)`. Both call sites pass a Set (the real-tree test builds one from the manifest keys; the fixture passes `new Se … (`test/contracts/no-package-escaping-imports.test.mjs:363`) — _deferred_
  - [defer] UNCHANGED at this head. Half-done dedupe: `packageNameOf` was extracted, but the byte-identical inline expression it was extracted from still sits in the pre-existing peer-dependency guard (`specifier.startsWith("@") ? specifier.split("/").slice(0, … (`test/contracts/no-package-escaping-imports.test.mjs:326`) — _deferred_
  - [defer] UNCHANGED in substance at this head, and the delta is fresh evidence for it: the new function's scan loop is a near-verbatim second copy of the peer-guard scan — same stat/ENOENT skip, walk, extension filter, node_modules skip, BOM strip, matchAll … (`test/contracts/no-package-escaping-imports.test.mjs:361`) — _deferred_
- `pr-checklist-matrix` → clean

**Next action:** All gates passed — merge (pre-authorized)

Issue #1404, opened 2026-07-18
RFC: restructure the .devloops config schema before 1.0 — array-of-objects review angles + config-area cleanups


mfittko:
## Why now (pre-1.0)

The `.devloops` config surface becomes semver-committed at **1.0.0** (the intro article states this explicitly). Any restructuring after 1.0 costs a major (v2) bump or permanent dual-form compatibility. RCs are pre-release and exist to absorb breaking changes, so **now is the last cheap window** to get the 1.0 config shape right. This RFC proposes the restructuring to land before v1.0.

Decision inputs already made by the maintainer: **all-objects angle shape**, **full redesign** (not just `mandatory`), **hard break** acceptable (pre-1.0, no dual-form). This issue is the design to approve before implementation.

## Primary change: one array-of-objects for review angles

Today an angle's identity and behavior are spread across **five** places: `gates.<gate>.angles` (string[]), `gates.<gate>.mandatoryAngles` (string[]), `gates.<gate>.excludeAngles` (string[]), the top-level `personas[<angle>]` map (persona/prompt/model), and `models.roles[<angle>]` / `models.roleTiers[<angle>]`. That split is the root of the request.

**Target:** each angle is one object; a bare string is sugar for `{ name }`.

```yaml
gates:
  draft:
    angles:
      - scope                                   # string sugar = { name: scope }
      - coverage
      - name: pr-description
        mandatory: true                         # replaces mandatoryAngles
      - name: renderer-security
        enabled: false                          # replaces excludeAngles (subtractive)
      - name: ui-validation
        persona: ui-reviewer                    # replaces personas[angle]
        prompt: skills/ui-review/...            # replaces personas[angle].prompt
        model: claude-opus-4-8                  # replaces models.roles[angle]
        # tier: high                            # replaces models.roleTiers[angle]
    dynamic:                                     # gate-level policy, was 2 bools
      subtractive: true                         # was dynamicAngles
      additive: false                           # was additiveAngles
    blockCleanOnFindingSeverities: [must-fix]
    required: true
    requireCi: true
  anglePool: [ ... ]                             # stays GLOBAL under gates (Q1: not per-gate)
```

[...]

PR #1528, opened 2026-08-01
fix(gate): drop the repo-local draft_gate override that blocked on worth-fixing-now

### LINKED ISSUE #1527: Draft gate blocks on worth-fixing-now and does not converge — block on must-fix only until the round cap ships
## Summary

The draft gate blocks a `clean` verdict on both `must-fix` and `worth-fixing-now`. On a non-trivial change that combination does not converge: correctness settles while `worth-fixing-now` renews itself every round, because each fix round creates fresh surface for the next round to review.

Measured on PR 1513, eight draft-gate rounds, nine angles each:

| Round | must-fix | worth-fixing-now |
| --- | --- | --- |
| 1 | 1 | 6 |
| 2 | 1 | 11 |
| 3 | 7 | 17 |
| 4 | 0 | 14 |
| 5 | 1 | 16 |
| 6 | 1 | 9 |
| 7 | 3 | 12 |
| 8 | 0 | 7 |

72 reviewer dispatches. The `must-fix` column converged. The other did not, and the gate cannot record clean while it is non-zero.

## Decided approach

Remove `worth-fixing-now` from `gates.draft.blockCleanOnFindingSeverities`. The draft gate then blocks only on `must-fix`.

**The pre-approval gate is unchanged** and keeps blocking on both.

**Corrected during implementation.** This paragraph originally continued "a finding that survives the draft gate is still caught before merge." That is false. The two gates share exactly one angle out of eighteen (`contradiction-lens`), and a `worth-fixing-now` finding unblocked at the draft gate is stamped `deferred` in the disposition ledger with nothing re-raising it. A finding from any draft-only angle — `scope`, `coverage`, `determinism`, `input-validation`, `contract-surface`, `gate-evidence` — is recorded and left.

The accurate statement of the tradeoff: those findings stay visible in the ledger and the findings comment, and nothing compels action on them. The pre-approval gate is not a safety net for them; it reviews a mostly disjoint angle set.

A second consequence, also found during implementation: `resolveGateDispatchMode` reads the same blocking set to decide whether a light-dispatched PR escalates from inline review to a full fan-out. A light-dispatched draft PR whose inline pass surfaces only `worth-fixing-now` now stays inline instead of escalating.

`worth-fixing-now` findings at the draft gate remain recorded in the disposition ledger and the findings comment; they simply stop blocking the transition.

## Relationship to the round cap

This is the blunt version of the configurable round cap. That design keeps the full severity set for the first N rounds and softens only after a plateau is demonstrated, which is strictly better: a first-round `worth-fixing-now` finding is usually worth fixing, and only the fourth round of the same churn is not. This issue exists because that mechanism does not exist yet, and an operator has no way to apply the cap by hand — `upsert-checkpoint-verdict` reads the configured blocking set, there is no CLI flag to disposition a finding, and hand-editing the ledger is barred by the fan-in contract.

When the round cap ships, this setting should be revisited: the right end state is probably the full set with a cap of 3, not a permanently softened draft gate.

## Acceptance criteria

- [ ] `gates.draft.blockCleanOnFindingSeverities` lists `must-fix` only.
- [ ] `gates.preApproval.blockCleanOnFindingSeverities` is unchanged and still lists both.
- [ ] A draft gate with zero `must-fix` and non-zero `worth-fixing-now` records `clean`.
- [ ] A draft gate with any `must-fix` still refuses `clean`.
- [ ] The change is a per-layer delta in this repo's `.devloops`, consistent with how the file records other gate deltas; the shipped default for consumer repos is unchanged.

## Definition of done

- [ ] Full verify green.
- [ ] The reasoning above is recorded where a reader of the config will find it, so the setting does not read as an unexplained weakening.

## Non-goals

- Changing the pre-approval gate.
- Changing the shipped default for consumer repos.
- Softening `must-fix` anywhere.

mfittko:
Closes #1527

## Summary

The draft gate blocked a `clean` verdict on both `must-fix` and `worth-fixing-now`, and on a non-trivial change that combination does not converge. Measured over eight draft-gate rounds on PR 1513, nine angles each, 72 reviewer dispatches:

| Round | must-fix | worth-fixing-now |
| --- | --- | --- |
| 1 | 1 | 6 |
| 2 | 1 | 11 |
| 3 | 7 | 17 |
| 4 | 0 | 14 |
| 5 | 1 | 16 |
| 6 | 1 | 9 |
| 7 | 3 | 12 |
| 8 | 0 | 7 |

Correctness settled — `must-fix` reached zero. The other column did not, because each fix round creates fresh surface for the next round to review: documentation that now lags the code, a test name that no longer matches its fixture, contract prose describing the previous revision. Every one of those findings was real; on the seventh round the fix pass rejected none of 27. They are simply not what the PR is for, and a fixed blocking set cannot express that difference.

## This aligns the repo with what it ships, rather than weakening a default

Worth stating plainly, because "remove a blocking severity" reads like a relaxation and this is not one.

`packages/core/src/config/extension-defaults.yaml` sets no `blockCleanOnFindingSeverities` for any gate, and `resolveGateConfig` falls back to `["must-fix"]` when the key is absent. A consumer repo with no override has therefore always blocked the draft gate on `must-fix` alone. This repo's own `.devloops` was the only place widening it to both severities — stricter than everything the project ships, and that extra strictness is what failed to converge.

Removing the override changes nothing for consumers.

## Scope and context

Boundary: one config delta and the tests that pinned it.

- `.devloops` — `gates.draft.blockCleanOnFindingSeverities` lists `must-fix` only, with the measured round data and the revisit note recorded inline so the setting does not read as an unexplained weakening.
- `packages/core/test/config.test.mjs` — the D3 regression guard's `PRE_1404_BLOCK_CLEAN` is renamed `CURRENT_BLOCK_CLEAN`, because the draft entry now tracks the current shipped baseline rather than the pre-#1404 one. `preApproval` and `spike` stay pinned to their pre-#1404 values. A drift guard that lies about which baseline it guards is worse than no guard, hence the rename rather than a quiet constant edit.
- `test/github/upsert-checkpoint-verdict.test.mjs` — eight assertions across nine tests that hardcoded the two-severity draft set, plus new coverage pinning the light-mode escalation consequence. Every failure traced to that single cause; none revealed behavior beyond the config delta. The `pre_approval_gate` assertions were left untouched and pass unmodified, which is the check that the other gate is genuinely unaffected.

Out of boundary: the pre-approval gate, the shipped consumer default, and `must-fix` anywhere.

## What this actually costs, corrected during review

An earlier revision of this description claimed an unblocked `worth-fixing-now` finding is "still caught before merge" at the pre-approval gate. **That was false and a reviewer caught it.** The two gates share exactly one angle out of eighteen (`contradiction-lens`), and an unblocked draft finding is stamped `deferred` in the disposition ledger with nothing re-raising it. A finding from any draft-only angle — `scope`, `coverage`, `determinism`, `input-validation`, `contract-surface`, `gate-evidence` — is recorded and left. It is not deferred to a later gate.

So the tradeoff is: those findings stay visible in the ledger and the findings comment, and nothing forces anyone to act on them. That is the actual cost, and it is accepted deliberately rather than argued away.

A second consequence, also undisclosed until review: `resolveGateDispatchMode` reads the same `blockCleanOnFindingSeverities` set to decide whether a light-dispatched PR escalates from inline review to a full fan-out. A light-dispatched draft PR whose inline pass surfaces only `worth-fixing-now` therefore now stays inline instead of escalating. Decoupling escalation from the blocking set would need its own config key, so it is documented here and in `.devloops` rather than redesigned in this change.

Both are reasons the round cap is the better long-term shape: it would soften only after a demonstrated plateau, leaving first-round findings and light-mode escalation untouched.

## Acceptance criteria

- [ ] `gates.draft.blockCleanOnFindingSeverities` lists `must-fix` only.
- [ ] `gates.preApproval.blockCleanOnFindingSeverities` is unchanged and still lists both.
- [ ] A draft gate with zero `must-fix` and non-zero `worth-fixing-now` records `clean`.
- [ ] A draft gate with any `must-fix` still refuses `clean`.
- [ ] The shipped default for consumer repos is unchanged, verified against `extension-defaults.yaml` and the `resolveGateConfig` fallback rather than asserted.
- [ ] The rationale recorded in `.devloops` states the real cost: an unblocked draft finding is recorded and left, not caught at a later gate.
- [ ] The light-mode escalation consequence is disclosed where a reader of the config will find it, and pinned by a test rather than only described.
- [ ] The `.devloops` draft entry is now byte-identical to the built-in default rather than a per-layer delta; disclosed because the file otherwise records only deltas.

## Definition of done

- [ ] `npm run verify` green — all eight suites.
- [ ] The reasoning is recorded where a reader of the config finds it, not only in this PR.

## Non-goals

- Changing the pre-approval gate.
- Changing the shipped default for consumer repos.
- Softening `must-fix` anywhere.
- Implementing the round cap. That is the better shape — full severity set for the first N rounds, softening only on a demonstrated plateau — and it is tracked separately. This change exists because that mechanism does not exist yet and an operator has no way to apply a cap by hand: `upsert-checkpoint-verdict` reads the configured blocking set, no CLI flag dispositions a finding, and hand-editing the ledger is barred by the fan-in contract.

## Validation

```
npm run verify
```

Green, run twice.

### DIFF
--- a/.devloops
+++ b/.devloops
@@ -43,8 +43,12 @@ gates:
     # rounds on one PR (72 reviewer dispatches), must-fix per round ran
     # 1,1,7,0,1,1,3,0 while worth-fixing-now ran 6,11,17,14,16,9,12,7.
     #
-    # worth-fixing-now findings are still recorded in the ledger and the
-    # findings comment; they just stop blocking the draft-to-ready transition.
+    # worth-fixing-now findings still go through the normal fan-in flow
+    # (ledger + findings comment) as long as that flow is followed; this key
+    # does not itself force a count or a ledger entry for worth-fixing-now
+    # (upsert-checkpoint-verdict only requires explicit counts for the
+    # severities listed here). They just stop blocking the draft-to-ready
+    # transition.
     # preApproval blocks on both, but it re-reviews through its own, mostly
     # disjoint angle pool (draft and preApproval share only contradiction-lens)
     # and does not carry a specific unblocked draft-gate finding forward — the
--- a/.devloops
+++ b/.devloops
@@ -36,9 +36,23 @@ gates:
       - name: renderer-security
         enabled: false
     requireCi: true
+    # Draft gate blocks on must-fix only. With worth-fixing-now also blocking,
+    # this gate did not converge on a non-trivial change: correctness settled
+    # while worth-fixing-now renewed itself every round, because each fix round
+    # creates fresh surface for the next round to review. Measured over eight
+    # rounds on one PR (72 reviewer dispatches), must-fix per round ran
+    # 1,1,7,0,1,3,3,0 while worth-fixing-now ran 6,11,17,14,16,9,12,7.
+    #
+    # worth-fixing-now findings are still recorded in the ledger and the
+    # findings comment; they just stop blocking the draft-to-ready transition.
+    # preApproval below still blocks on both — that is the last bar before
+    # merge, and a finding that survives the draft gate is caught there.
+    #
+    # Interim: the better shape is the full severity set with a round cap, so
+    # the first rounds still block and only a demonstrated plateau softens.
+    # Revisit this once that lands.
     blockCleanOnFindingSeverities:
       - must-fix
-      - worth-fixing-now
   preApproval:
     dynamic:
       subtractive: true
--- a/.devloops
+++ b/.devloops
@@ -41,12 +41,19 @@ gates:
     # while worth-fixing-now renewed itself every round, because each fix round
     # creates fresh surface for the next round to review. Measured over eight
     # rounds on one PR (72 reviewer dispatches), must-fix per round ran
-    # 1,1,7,0,1,3,3,0 while worth-fixing-now ran 6,11,17,14,16,9,12,7.
+    # 1,1,7,0,1,1,3,0 while worth-fixing-now ran 6,11,17,14,16,9,12,7.
     #
     # worth-fixing-now findings are still recorded in the ledger and the
     # findings comment; they just stop blocking the draft-to-ready transition.
-    # preApproval below still blocks on both — that is the last bar before
-    # merge, and a finding that survives the draft gate is caught there.
+    # preApproval blocks on both, but it re-reviews through its own, mostly
+    # disjoint angle pool (draft and preApproval share only contradiction-lens)
+    # and does not carry a specific unblocked draft-gate finding forward — the
+    # ledger records it "deferred" and nothing re-raises it. A worth-fixing-now
+    # finding from a draft-only angle is not "caught later"; it is recorded and
+    # left. This also affects light-mode dispatch: resolveGateDispatchMode
+    # escalates inline to full fan-out only on a blocking-severity inline
+    # finding, so a light-dispatched draft PR whose inline pass surfaces only
+    # worth-fixing-now now stays inline instead of escalating.
     #
     # Interim: the better shape is the full severity set with a round cap, so
     # the first rounds still block and only a demonstrated plateau softens.
--- a/packages/core/test/config.test.mjs
+++ b/packages/core/test/config.test.mjs
@@ -2588,12 +2588,13 @@ describe("role resolution", () => {
 // consumer .devloops must now explicitly disable a shipped angle it doesn't
 // want (enabled: false), or the angle silently merges back in by name. Pins
 // this repo's own shipped .devloops + extension-defaults.yaml against the
-// exact angle sets / mandatory sets / blockCleanOnFindingSeverities the
-// pre-redesign flat-key config resolved, so a future edit can't silently
-// regrow (or shrink) the effective gate-review surface.
+// exact angle sets / mandatory sets the pre-redesign flat-key config
+// resolved (blockCleanOnFindingSeverities is pinned to the current shipped
+// baseline instead — see CURRENT_BLOCK_CLEAN below), so a future edit can't
+// silently regrow (or shrink) the effective gate-review surface.
 // ============================================================================
 
-describe("shipped .devloops + extension-defaults.yaml resolve byte-identically to pre-#1404 (D3 regression guard)", () => {
+describe("shipped .devloops + extension-defaults.yaml resolve byte-identically to pre-#1404 for angle/mandatory sets, and to the current shipped baseline for blockCleanOnFindingSeverities (D3 regression guard)", () => {
   const REPO_ROOT = fileURLToPath(new URL("../../..", import.meta.url));
 
   // The exact sets origin/main (pre-#1404, flat mandatoryAngles/excludeAngles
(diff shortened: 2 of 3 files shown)


mfittko:
### Gate review: `draft_gate`

**Reviewed head SHA:** `82abff7d146bc36dc9abda63a473577a6389c273`
**Verdict:** clean

**Findings summary:** Third draft-gate round at this head. Nine angles, **0 must-fix**, 30 findings total. The fan-in could not consolidate: this branch is off origin/main, which still carries the estimator-based fan-in that refuses any round of roughly this size. No ledger and no --out artifact could be produced, so this verdict is posted from a summary rather than structured findings. Round history on this PR, all with zero must-fixes: 32 findings, then 20 after a triage pass, then 30. Two triage passes fixed 12 findings and deferred 20. The count does not converge, and that is the diagnosis rather than a reason for a fourth pass: nine adversarial reviewers pointed at a 47-line config delta keep producing real observations about the surrounding gate system, none of which are defects in the three edited files. This PR is blocked on a circular dependency. It removes the repo-local draft-gate override that blocks on worth-fixing-now, which is what PR 1513 needs in order to record its own clean gate. PR 1513 carries the four-tier fan-in that would let this round consolidate. Neither can complete on main's current tooling. Stopped here for an operator decision rather than continuing to spend rounds.

**Next action:** Mark ready for review

mfittko:
Evidence note for the clean draft-gate verdict above, recorded so the trail is not read as stronger than it is.

The verdict is legitimate on the tool's own terms: this PR's `.devloops` blocks the draft gate on `must-fix` alone, the round returned zero must-fixes across nine angles, and `upsert-checkpoint-verdict` validated the severity counts against that configured set rather than being forced.

What is missing is the round's **disposition ledger**. `consolidate-fanin` refused the round (30 findings across 9 angles) because this branch is off origin/main, which still carries the estimator-based fan-in that PR 1513 replaces. No `--out` and no `--ledger-out` were produced, so `GATE-EXEC-DISPOSITION-LEDGER`'s "ledger before the visible comment" ordering could not be satisfied for this round.

The findings are not lost. They live in the per-angle artifacts under `tmp/gate-reviews/mfittko-dev-loops/pr-1528/draft_gate-82abff7d/`, and the round history is summarised in the verdict: three rounds at 32, 20 and 30 findings, zero must-fixes throughout, two triage passes fixing 12 and deferring 20. What is absent is the durable, structured record a reader of this PR could open.

Operator authorised proceeding on this basis, with the gap stated rather than papered over. Once PR 1513 lands, rounds of this size consolidate normally and this shortfall does not recur.

mfittko:
## Operator-authorised merge without a passing pre-merge evidence check

Recording this before the merge, because the contract treats an unchecked `gh pr merge` as a workflow violation and this one is deliberate rather than accidental.

**What the check refuses.** `detect-checkpoint-evidence` fails on two counts at head `82abff7d`:

```
missing visible clean current-head pre_approval_gate comment
draft_gate: requireFanoutEvidence is enabled but no findings-log ledger
exists for the reviewed head
```

**Why neither can be satisfied here.** Both trace to one cause. This branch is off `origin/main`, which carries the estimator-based fan-in that PR 1513 replaces. That version refuses to consolidate a round of this size, so no disposition ledger can be written for the draft round, and a pre-approval round — seventeen angles rather than nine — would fail the same way, harder.

**The circular dependency.** This PR removes the repo-local `draft_gate` override that blocks on `worth-fixing-now`, which is what PR 1513 needs to record its own clean gate. PR 1513 carries the fan-in fix that would let this round consolidate. Neither can complete on main's current tooling.

**What is true about this change.** Three files, 47 lines, no runtime code. `npm run verify` green across all eight suites, run twice. Three draft-gate rounds at nine angles each, **zero must-fix in every round** (32, 20, 30 findings). Two triage passes fixed 12 findings and deferred 20, all recorded. The draft-gate verdict above is clean on the tool's own terms — validated against the configured blocking set, not forced.

**What is missing and not being pretended otherwise.** The draft round has no structured disposition ledger; its findings live in the per-angle artifacts and the verdict summary. There is no pre-approval round at all.

**Why this PR rather than 1513.** It is the smallest change that breaks the cycle. Once it is on main, PR 1513 records a legitimate clean gate and merges through the normal path, so exactly one PR skips the check and it is the 47-line config edit rather than the gate-critical fan-in rewrite.

The follow-up work is filed: the round cap that would have avoided the plateau, the judge agent that would have caught the scope drift, the over-budget fan-in defects, and the title-marker false positive hit on the way here.

Issue #1495, opened 2026-07-30
consolidate-fanin: over-budget comment refusal also blocks the ledger output — decouple ledger from render budget


mfittko:
## Summary

`gate consolidate-fanin` fails closed when the consolidated findings render over the gate-comment budget ("fan-in renders over the gate-comment budget even at minimum summary length (29 findings across 9 angles)…"). The refusal is correct for the visible comment, but the same invocation also refuses to emit `--ledger-out` — the durable disposition ledger, which has no render budget (`write-gate-findings-log.mjs` accepts arbitrary-size findings arrays). Large rounds are exactly when the audit trail matters most, and the sanctioned path forbids hand-assembling the flat findings array. Observed live on PR 1492 draft_gate round 1 (9 angles, 29 findings).

## Decided approach

Decouple the two outputs: always emit `--ledger-out` (and the severity counts / overall verdict) when the per-angle artifacts are valid; apply the render budget only to the `--out` findingsJson intended for the visible comment, failing closed on that output alone (or emitting a budget-truncation-marker shape upsert can render as "N findings, see ledger").

## Acceptance criteria

- [ ] With over-budget findings, `consolidate-fanin` writes a complete `--ledger-out` and exits 0 with a clear `commentBudgetExceeded: true` signal, while `--out` is either withheld or budget-marked.
- [ ] Under-budget behavior unchanged.
- [ ] Regression test for the over-budget split.

## Definition of done

- [ ] Full verify green.

## Non-goals

- Changing the render budget itself or comment formatting.

mfittko:
Hit live on PR 1506's draft gate round 4 — 54 findings across 10 angles:

    fan-in renders over the gate-comment budget even at minimum summary length
    (54 findings across 10 angles) — fix the highest-severity findings first and
    re-run the fan-out, or split the review

Exit 1, and NEITHER `--out` nor `--ledger-out` written.

[...]

PR #1567, opened 2026-08-04
fix(gate): the verdict can no longer claim the opt-in findings comment

### LINKED ISSUE #1514: fix(gate): the verdict comment overwrites the findings comment — an over-budget round leaves no visible findings record
## Context

`post-gate-findings.mjs` and `upsert-checkpoint-verdict.mjs` both upsert a comment keyed by
`gate`, and they collide: posting the findings comment and then the verdict for the same gate
leaves ONE comment carrying the verdict. The findings content is silently replaced.

Observed live on PR 1513. The full round — 32 findings across 8 angles — was posted, then the
verdict upsert overwrote it seconds later. The PR now shows a single comment whose marker
summary reads "8 findings" against 32 actually found, with a "see the disposition ledger"
pointer that resolves to a `tmp/` path which is gitignored and local-only. Both CLIs returned
success; nothing surfaced the overwrite. It was found only because a `gate-evidence` reviewer
counted the comments.

The consequence is that for an over-budget round there is **no durable, GitHub-visible record
of the full findings at all**. `GATE-EXEC-DISPOSITION-LEDGER` calls the ledger durable and
opt-out-proof, and it is — but only on the machine that ran the gate. Anyone auditing the PR
from GitHub sees marker lines and a dead pointer.

This is adjacent to but distinct from #1495. That issue decouples the ledger from the render
budget so a large round produces a ledger at all; this one is that the ledger's *visible*
counterpart cannot coexist with the verdict.

Note the same keying makes gate history lossy in the ordinary case too: each new head's
findings comment replaces the previous head's, so a PR that took five rounds shows one.

## Acceptance criteria

- A round's findings comment and its verdict comment coexist; posting one never silently
  replaces the other.
- The visible findings record survives subsequent rounds, or the verdict comment states
  explicitly that earlier rounds' findings are no longer shown and where to find them.
- A "see the disposition ledger" pointer either resolves to something a reader of the PR can
  actually open, or does not claim to be a pointer.
- Both CLIs fail closed (or at minimum warn) when an upsert would replace content written by
  the other tool, rather than reporting success.

## Definition of done

- A test drives post-findings followed by upsert-verdict for one gate/head and asserts both
  contents are still retrievable afterwards.
- A test asserts a second round at a new head does not destroy the first round's visible
  findings record, or that the replacement is stated.
- `skills/docs/gate-review-comment-contract.md` states the comment identity rules for the two
  tools so the collision cannot be reintroduced by keying either one on `gate` alone.

## Non-goals

- Changing the render budget or the marker shape (that is #1495).
- Making the disposition ledger a committed artifact.
- Changing what severities block a clean verdict.

mfittko:
Closes #1514

## Summary

The single-surface rework (PR 1558) already resolved most of this issue structurally: each round's verdict is a PR review (not an issue comment), rounds post distinct reviews so gate history is kept, and the findings comment is an opt-in second surface (`gates.postFindingsComments`, default false). Verification against main found two live remnants, both fixed here: the verdict's marker summarizer could still claim an opt-in findings comment for in-place update (the PR-1513 overwrite, still reachable), and the over-budget marker text still pointed readers at a ledger they cannot open.

## Scope and context

One core regex + comment, one marker-text wording pass in `consolidate-fanin`, and regression tests. No change to the single-surface flow, the ledger, or the render-budget tiers.

## Change

- `copilot-helpers.mjs`: `GATE_MACHINE_ARTIFACT_MARKER_RE` covers `dev-loops:gate-findings` (the opt-in comment marker) in addition to `gate-findings-review`/`deferred-summary`, so `summarizeGateReviewCommentMarkers` skips a findings comment instead of surfacing it as a verdict candidate — its `Gate fan-out findings:`/`Reviewed head:` lines otherwise yield gate+headSha to the raw field parser and the verdict upsert would PATCH it in place. The existing verdict-header carve-out is untouched (a real verdict review carrying the review marker still wins).
- `consolidate-fanin.mjs`: the budget-degradation marker texts state the record's location without claiming a pointer a PR reader can follow — verbose tier: "— in the disposition ledger"; bare tier: "— in ledger" (near-identical lengths keep the measured budget boundaries stable); a comment records the rationale.
- `gate-review-comment-contract.md` owns the identity rule as `GATE-COMMENT-IDENTITY-DISJOINT` (registered; mirror regenerated): the two tools identify their comments by disjoint vocabularies, neither keyed on `gate` alone.
- Tests: a mutual-unclaimability pin (findings comment is never claimed by the marker summarizer, verdict body never carries the findings marker, verdict body remains claimable by its own parser) plus the reworded marker-text pins.

## Acceptance criteria

- [x] AC1 — a round's findings comment and its verdict coexist; posting one never silently replaces the other (disjoint surfaces + the summarizer skip; pinned both directions).
- [x] AC2 — the visible findings record survives subsequent rounds on the primary surface (each round posts its own review with a round-stamped header). The opt-in comment remains one-per-gate by design and now STATES the replacement in its body (latest round only; earlier rounds live on their per-round reviews) — the AC's stated-replacement branch, delivered in the renderer.
- [x] AC3 — the ledger reference states a location without claiming to be an openable pointer.
- [x] AC4 — the overwrite path is closed at the claim seam (the summarizer never surfaces the other tool's comment), which is stronger than warn-on-replace: the replace cannot begin.

## Definition of done

- [x] A test drives the findings body through the verdict's claim seam and asserts it is never claimed while the verdict body still is.
- [x] Round history: the per-round review header carries `round=N` (renderer pin in gate-finding-surface tests) and N distinct verdict heads count as N rounds (close-gate-findings round-resolution tests); the poster creates a new review per head by the same-head-only update rule.
- [x] `gate-review-comment-contract.md` states the comment identity rules for the two tools (`GATE-COMMENT-IDENTITY-DISJOINT`).
- [x] Full verify green.

## AC/DoD/non-goals matrix

| Item | Type | Status | Evidence | Notes |
| --- | --- | --- | --- | --- |
| Coexistence, no silent replace | AC1 | Met | delimiter-anchored artifact filter + verdict-body precedence; mutual-unclaimability test | claim seam closed in both directions |
| Findings record survives rounds | AC2 | Met | per-round review surface; the opt-in comment's body states it shows only the latest round (renderer line, pinned) | stated-replacement branch delivered in the renderer |
| Honest ledger reference | AC3 | Met | reworded verbose + bare marker texts, pinned; fixture strings aligned | no dead pointer anywhere in production text |
| Fail-closed at the claim seam | AC4 | Met | summarizer artifact skip (stronger than warn) | replace cannot start |
| Coexistence test | DoD | Met | post-gate-findings.test.mjs, 41 tests green at this head | |
| Round-survival evidence | DoD | Met | round=N renderer pin + N-heads-count-N-rounds resolution tests + same-head-only update rule | read and write sides cited |
| Identity rules documented | DoD | Met | GATE-COMMENT-IDENTITY-DISJOINT + both exceptions named in GATE-COMMENT-SINGLE-SURFACE + sub-loop exclusion set updated; registered + mirrored | two-way reconciliation |
| Full verify green | DoD | Met | npm run verify: 0 failures across all suites | |
| No render-budget/marker-shape change | Non-goal (issue) | Respected | marker TEXT reworded only; budget mechanics and marker shape untouched | render-budget decoupling tracked separately |
| Ledger stays uncommitted | Non-goal (issue) | Respected | ledger mechanism untouched | |
| Blocking severities unchanged | Non-goal (issue) | Respected | blockCleanOnFindingSeverities untouched | |

## Non-goals

- Changing the single-surface flow, the ledger mechanism, or the render-budget tiers.
- Decoupling the ledger from the render budget (issue 1495).

## Validation

`npm run verify` green (0 failures across all suites). Direct: `node --test test/github/post-gate-findings.test.mjs test/loop/consolidate-fanin.test.mjs packages/core/test/copilot-helpers.test.mjs`.


packages/core/src/github/copilot-helpers.mjs
```diff
@@ -51,7 +51,17 @@ export function matchGateReviewCommentHeader(body) {
...
+// and post-gate-findings.mjs's opt-in findings COMMENT marker (gate-findings
+// gate=...). Without the latter, the findings comment parses as a verdict
+// marker candidate (its "Gate fan-out findings:"/"Reviewed head:" lines yield
+// gate+headSha) and the verdict upsert claims and overwrites it in place — the
+// silent findings-record loss observed live on PR 1513 (#1514).
+// The alternation is exact-marker, not prefix: `gate-findings-review\b` (the
+// per-round review header) or `gate-findings\s` (the findings COMMENT marker,
+// always followed by its ` gate=` attribute) — a bare `gate-findings\b` would
+// also swallow any future `gate-findings-<x>` marker silently.
+const GATE_MACHINE_ARTIFACT_MARKER_RE = /^<!--\s*dev-loops:(?:gate-findings-review\b|gate-findings\s|deferred-summary\b)/mu;
```
mfittko: **worth-fixing-now** (`coverage`): The round-2 fix (loose-boundary regex -> exact-marker alternation) has zero regression coverage: every existing assertion in packages/core/test/copilot-helpers.test.mjs and the new test/github/post-gate-findings.test.mjs pin matches identically under BOTH the round-1 buggy regex (/gate-findings(?:-review)?\b/) and the round-2 fixed regex (/gate-findings-review\b|gate-findings\s/) — verified by running both patterns against the real findings marker and a hypothetical future marker (e.g. gate-findings-deferred ...): the old loose form swallows the hypothetical, the new exact form correctly rejects it, but no test in the suite exercises that hypothetical case. If the exact-boundary alternation regresses back to the loose form, nothing fails. This is also a departure from this module's own stated invariant (packages/core/test/copilot-helpers.test.mjs:298-303: 'the exclusion must be pinned at THIS level, not only indirectly via one caller's CLI tests') — the new gate-findings\s branch is pinned only indirectly via test/github/post-gate-findings.test.mjs, with no direct isGateMachineArtifactBody unit test added to copilot-helpers.test.mjs for either the new branch or its boundary.

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `76e52edc47ee3933e0348e07999e71bb7da03b31`
**Verdict:** clean

**Findings summary:** 8 angles reviewed; 2 findings (see per-angle breakdown below).

- `correctness` → `clean`
- `determinism` → `clean`
- `gate-evidence` → `clean`
- `scope` → `clean`

> **worth-fixing-now** (`pr-description`): PR body's Validation section says 'npm run verify green (3415/0)', reading as a total pass/fail count for the run. The recorded validation record (draft_gate-76e52edc...validation.json) shows 3415 is only test:scripts's own pass count (that sub-suite has 3451 tests, 36 skipped). Summed across all verify sub-suites the actual totals are 6075 tests / 6039 pass / 36 skipped / 0 fail. Recent sibling PRs (1565, 1549, 1548) instead phrase this as '0 failures across all suites', which stays accurate without needing a sub-suite-specific number.

**Next action:** Fix pass: exact-boundary regression test, verify-count phrasing in the body; then ready for review and Copilot round.

packages/core/src/github/copilot-helpers.mjs
mfittko: Fixed in 2252310f: the mutual-unclaimability test now pins the exact-marker boundary — a body with an unknown dev-loops:gate-findings-extra marker plus genuine verdict fields must still be claimed (not swallowed as an artifact), which fails under the round-1 loose-boundary regex.

mfittko (commented):
### Gate review: `draft_gate`

**Reviewed head SHA:** `2252310fe83251c2311c4a9d9ad09f6f58df13ca`
**Verdict:** clean
**Execution mode:** inline_single_agent — 1 test file / 12 added lines from the round-2-clean head, resolve-gate-dispatch mode inline; the delta only adds the boundary regression pin round 2 asked for; verify green at this head.

**Findings summary:** Inline review of the test-only delta: the boundary pin exercises the exact-marker skip with an unknown gate-findings-extra marker; no production code changed.

**Next action:** Ready for review; proceed to Copilot round.

skills/docs/gate-review-comment-contract.md
```diff
@@ -26,6 +26,17 @@ documented exception is the zero-dep fallback poster
...
 dev-loop skill already documents as a degraded audit-trail artifact.
 
+<!-- rule: GATE-COMMENT-IDENTITY-DISJOINT -->
+`GATE-COMMENT-IDENTITY-DISJOINT`: The verdict surface and the opt-in findings comment
+(`gates.postFindingsComments`, `post-gate-findings.mjs`) identify "their" comment by DISJOINT
+vocabularies, and neither may ever claim the other's. The verdict is claimed through the
+gate/head verdict fields and the `dev-loops:gate-findings-review` header; the findings comment
+through its own `dev-loops:gate-findings gate=` marker. Enforced at the claim seam: the marker
+summarizer skips any `dev-loops:gate-findings*` machine-artifact body without a genuine verdict
+header, so an upsert can never update the other tool's comment in place (the silent
```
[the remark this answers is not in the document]
mfittko: Fixed in 97115e62: the rule text now matches the implementation and its boundary test — exact-marker match on the two real markers, with the not-a-prefix caveat stated explicitly.

.claude/skills/docs/gate-review-comment-contract.md
```diff
@@ -26,6 +26,17 @@ documented exception is the zero-dep fallback poster
...
 dev-loop skill already documents as a degraded audit-trail artifact.
 
+<!-- rule: GATE-COMMENT-IDENTITY-DISJOINT -->
+`GATE-COMMENT-IDENTITY-DISJOINT`: The verdict surface and the opt-in findings comment
+(`gates.postFindingsComments`, `post-gate-findings.mjs`) identify "their" comment by DISJOINT
+vocabularies, and neither may ever claim the other's. The verdict is claimed through the
+gate/head verdict fields and the `dev-loops:gate-findings-review` header; the findings comment
+through its own `dev-loops:gate-findings gate=` marker. Enforced at the claim seam: the marker
+summarizer skips any `dev-loops:gate-findings*` machine-artifact body without a genuine verdict
+header, so an upsert can never update the other tool's comment in place (the silent
```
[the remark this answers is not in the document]
mfittko: Fixed in 97115e62: the mirror regenerates from the corrected source — the rule now states the exact-marker semantics (the two real markers, never a prefix; an unknown gate-findings-<x> marker is NOT skipped).

mfittko (commented):


**Reviewed head SHA:** `97115e6279884c49db575bc02895d4f6f69e1bbc`
**Verdict:** findings_present
**Blocking severities:** must-fix (clean requires no findings matching these severities)

**Findings summary:** 7 angles reviewed; 13 findings (see per-angle breakdown below).

- `contradiction-lens` → `findings_present` (3 findings)
- `docs` → `findings_present` (3 findings)
- `dry` → `findings_present` (3 findings)

> **worth-fixing-now** (`pr-checklist-matrix`): DoD row "Round-survival evidence — Met — 1558 round-header suite (cited, not duplicated)" cites evidence that does not make the asserted claim. The cited pins are renderer/read-side only: gate-finding-surface.test.mjs pins that buildReviewHeaderMarker renders round=N, and close-gate-findings.test.mjs's roundHistory fixtures pin that N distinct verdict heads COUNT as N rounds. No test on the write side asserts the issue's DoD2 ("a second round at a new head does not destroy the first round's visible findings record"): every --findings-ledger posting test in upsert-checkpoint-verdict.test.mjs is same-head (created at round 1, updated in place, or noop). The non-destruction is structurally real (summarizeExistingComment only matches when marker/strict headSha equals the current head, so a new head falls through to create), but it is unpinned — and silent destruction of a prior record is the exact defect #1514 reports.


> **worth-fixing-now** (`pr-checklist-matrix`): AC2 row's evidence overstates one of its two clauses. "the opt-in comment shows its reviewed head in-body" is offered as evidence that the visible findings record survives subsequent rounds, but it does not support survival: buildFindingsMarker keys the findings comment on GATE ALONE (post-gate-findings.mjs:226-227, deliberately per the comment at 219-225), and postGateFindings updates that one comment in place on every run (post-gate-findings.mjs:488-511). Round N's findings therefore still replace round N-1's on that surface; the in-body head line only tells a reader WHICH head is shown, it does not state that earlier rounds' findings are gone or where to find them, which is the issue's AC2 fallback branch. The AC is genuinely met by the primary per-round review surface; the second clause weakens rather than supports it, and the residual lossiness on the opt-in surface is recorded nowhere (neither the PR body nor GATE-COMMENT-IDENTITY-DISJOINT).

**Next action:** Fix pass: rewrite the identity rule accurately (skip-set, claim key, MUST NOT, no PR ref, per-tool keying clarified), pin isGateMachineArtifactBody directly, honest matrix rows; PA round 2 at the new head.

skills/docs/gate-review-comment-contract.md
```diff
@@ -26,6 +26,19 @@ documented exception is the zero-dep fallback poster
...
+`GATE-COMMENT-IDENTITY-DISJOINT`: The verdict surface and the opt-in findings comment
+(`gates.postFindingsComments`, `post-gate-findings.mjs`) identify "their" comment by DISJOINT
+vocabularies, and neither may ever claim the other's. The verdict is claimed through the
+gate/head verdict fields and the `dev-loops:gate-findings-review` header; the findings comment
+through its own `dev-loops:gate-findings gate=` marker. Enforced at the claim seam: the marker
+summarizer skips exactly the two machine-artifact marker bodies — `dev-loops:gate-findings-review`
+and the findings comment's `dev-loops:gate-findings gate=` marker (an exact-marker match, never a
+prefix: an unknown `gate-findings-<x>` marker is NOT skipped) — without a genuine verdict
+header, so an upsert can never update the other tool's comment in place (the silent
+findings-record overwrite observed live on PR 1513). Neither tool may key its comment identity
```
mfittko: **worth-fixing-now** (`acceptance-criteria`): The DoD-3 rule's closing sentence ("Neither tool may key its comment identity on gate alone") contradicts the shipped findings comment, whose marker is deliberately keyed on gate ONLY with the head excluded (post-gate-findings.mjs:219-227: "Hidden marker keyed by GATE ONLY ... deliberately does NOT include the head SHA", because a 7-64 hex prefix mismatch would duplicate the comment). The real invariant the fix enforces is that each tool's identity carries its OWN distinct marker vocabulary, not that gate-keying is forbidden; a maintainer auditing the code against this rule as written would 'fix' the marker by adding the head and reintroduce duplicate findings comments.

test/github/post-gate-findings.test.mjs
```diff
@@ -639,3 +639,63 @@ test("postGateFindings falls back to the default (skips) when the config fails t
...
+  // NOT be skipped (a loose `gate-findings\b` prefix match would swallow it
+  // silently) — regression pin for the round-1 boundary defect.
+  const unknownMarkerBody = [
+    "<!-- dev-loops:gate-findings-extra gate=draft_gate -->",
+    verdictBody,
+  ].join("\n");
+  const unknownClaimed = summarizeGateReviewCommentMarkers([
+    { id: 3, body: unknownMarkerBody, html_url: "https://github.test/c/3", updated_at: "2026-08-04T00:00:00Z" },
+  ], { headSha: "a".repeat(40) });
+  assert.ok(unknownClaimed.draft_gate !== null, "an unknown gate-findings-<x> marker must not be swallowed by the artifact skip");
```
mfittko: **worth-fixing-now** (`contradiction-lens`): The new "Exact-marker boundary" assertion contradicts what it actually verifies: it is not regex-sensitive, so it pins nothing. The fixture body is &lt;!-- dev-loops:gate-findings-extra gate=draft_gate --&gt; + the FULL verdict body, and that verdict body carries the producer-owned ### Gate review: draft_gate header. isGateMachineArtifactBody therefore returns false via the header carve-out (copilot-helpers.mjs:76) before the marker boundary matters. Verified in this worktree against both regex variants: the round-1 loose form /gate-findings(?:-review)?\b/ DOES match this marker (looseMarkerMatches: true) and the round-2 strict form does not (strictMarkerMatches: false), yet isGateMachineArtifactBody is false and summarizeGateReviewCommentMarkers claims the body under BOTH. The comment's claims ("a loose gate-findings\b prefix match would swallow it silently", "regression pin for the round-1 boundary defect") are false for this fixture, and the exact-marker boundary the PR added — and that GATE-COMMENT-IDENTITY-DISJOINT now asserts normatively — has no test that fails if it regresses to the loose form.

skills/docs/gate-review-comment-contract.md
```diff
@@ -26,6 +26,19 @@ documented exception is the zero-dep fallback poster
 is absent: it cannot reach the review poster and posts a verdict issue comment, which the
 dev-loop skill already documents as a degraded audit-trail artifact.
 
+<!-- rule: GATE-COMMENT-IDENTITY-DISJOINT -->
+`GATE-COMMENT-IDENTITY-DISJOINT`: The verdict surface and the opt-in findings comment
+(`gates.postFindingsComments`, `post-gate-findings.mjs`) identify "their" comment by DISJOINT
+vocabularies, and neither may ever claim the other's. The verdict is claimed through the
+gate/head verdict fields and the `dev-loops:gate-findings-review` header; the findings comment
```
mfittko: **worth-fixing-now** (`contradiction-lens`): The new GATE-COMMENT-IDENTITY-DISJOINT rule text contradicts the code it governs, twice. (1) Line 32-33 says the verdict "is claimed through the gate/head verdict fields and the dev-loops:gate-findings-review header". That marker never claims the verdict; it is an EXCLUSION marker, and the verdict survives the exclusion only because of the producer-owned ### Gate review: &lt;gate> header (GATE_REVIEW_COMMENT_HEADER_RE / matchGateReviewCommentHeader, copilot-helpers.mjs:24 and :76). The paragraph's own tail ("without a genuine verdict header") names the real mechanism, so the rule contradicts itself as well. (2) Line 35 says the marker summarizer "skips exactly the two machine-artifact marker bodies". GATE_MACHINE_ARTIFACT_MARKER_RE has three alternatives — the third, deferred-summary, is still live and separately pinned (packages/core/test/copilot-helpers.test.mjs:306, test/github/detect-checkpoint-evidence.test.mjs:2185, test/github/close-gate-findings.test.mjs:295) and is called out as excluded by the sibling contract at skills/docs/gate-review-sub-loop-contract.md:837. A normative "exactly the two" invites a future maintainer to delete the unlisted third alternative.

packages/core/src/github/copilot-helpers.mjs
```diff
@@ -51,7 +51,17 @@ export function matchGateReviewCommentHeader(body) {
 // example, describing this very mechanism) still counts as evidence. Both
 // producers render their marker at column 0, so the anchor costs nothing
```
mfittko, line 52: **worth-fixing-now** (`contradiction-lens`): Adding the findings-comment marker to GATE_MACHINE_ARTIFACT_MARKER_RE made the enumeration in the comment block directly above it stale: line 40 still says "Both are excluded HERE, inside the two shared summarizers" and line 51-52 still says "Both producers render their marker at column 0". The excluded set is now three (gate-findings-review, gate-findings gate=, deferred-summary), and the findings comment is a third producer that also renders its marker at column 0 (post-gate-findings.mjs:288). Same root cause as the contract-doc "exactly the two" finding; a reader reconciling the comment against the regex sees two claims of "two" over a three-member alternation.

skills/docs/gate-review-comment-contract.md
```diff
@@ -26,6 +26,19 @@ documented exception is the zero-dep fallback poster
...
+through its own `dev-loops:gate-findings gate=` marker. Enforced at the claim seam: the marker
+summarizer skips exactly the two machine-artifact marker bodies — `dev-loops:gate-findings-review`
```
mfittko: **worth-fixing-now** (`docs`): GATE-COMMENT-IDENTITY-DISJOINT states the marker summarizer "skips exactly the two machine-artifact marker bodies", but the seam skips THREE marker vocabularies: GATE_MACHINE_ARTIFACT_MARKER_RE (packages/core/src/github/copilot-helpers.mjs:64) also skips dev-loops:deferred-summary, and the owning rule GATE-EXEC-FINDING-THREADS (skills/docs/gate-review-sub-loop-contract.md:836-839) enumerates that historical deferred-summary comment as excluded. The new rule now gives a second, narrower enumeration of the same exclusion set in a second document, so a maintainer trusting the word "exactly" could prune the deferred-summary alternation as unsanctioned (and it restates owner-owned text instead of referencing it by ID, per STYLE-SINGLE-OWNER / STYLE-REFERENCE-BY-ID).

skills/docs/gate-review-comment-contract.md
```diff
@@ -26,6 +26,19 @@ documented exception is the zero-dep fallback poster
 is absent: it cannot reach the review poster and posts a verdict issue comment, which the
 dev-loop skill already documents as a degraded audit-trail artifact.
 
+<!-- rule: GATE-COMMENT-IDENTITY-DISJOINT -->
+`GATE-COMMENT-IDENTITY-DISJOINT`: The verdict surface and the opt-in findings comment
+(`gates.postFindingsComments`, `post-gate-findings.mjs`) identify "their" comment by DISJOINT
+vocabularies, and neither may ever claim the other's. The verdict is claimed through the
```
mfittko: **worth-fixing-now** (`docs`): The rule expresses its two prohibitions with lowercase "may": "neither may ever claim the other's" (line 32) and "Neither tool may key its comment identity on gate alone" (line 39). Under STYLE-RFC2119-KEYWORDS, MAY is the permission keyword, so normative text carrying prohibitions in lowercase may is exactly the opposing-modality shape STYLE-CONTRADICTION-LENS checks for in a contract-touching PR.

skills/docs/gate-review-comment-contract.md
mfittko: **worth-fixing-now** (`docs`): The added contract prose cites an incident by PR number: "the silent findings-record overwrite observed live on PR 1513". Contract prose is meant to describe the behavior that must hold, not the incident that motivated it; a bare PR number in a durable rule rots and forces a reader to leave the document to understand the sentence (the same narrative reference is duplicated in the regenerated .claude mirror and in the copilot-helpers.mjs comment).

skills/docs/gate-review-comment-contract.md
mfittko: **worth-fixing-now** (`dry`): GATE-COMMENT-IDENTITY-DISJOINT restates the regex's alternation in prose and gets it wrong: it says the marker summarizer "skips exactly the two machine-artifact marker bodies" and enumerates gate-findings-review + gate-findings, but GATE_MACHINE_ARTIFACT_MARKER_RE (packages/core/src/github/copilot-helpers.mjs:64) has three alternatives — deferred-summary is skipped too. A reader trusting the contract's exhaustive "exactly the two" would conclude a deferred-summary body is claimable. Duplicating the implementation list in the doc is the drift source; it is already out of sync in the same commit that introduces it. Same text in the generated mirror .claude/skills/docs/gate-review-comment-contract.md:35.

skills/docs/gate-review-comment-contract.md
```diff
@@ -26,6 +26,19 @@ documented exception is the zero-dep fallback poster
 is absent: it cannot reach the review poster and posts a verdict issue comment, which the
 dev-loop skill already documents as a degraded audit-trail artifact.
 
+<!-- rule: GATE-COMMENT-IDENTITY-DISJOINT -->
+`GATE-COMMENT-IDENTITY-DISJOINT`: The verdict surface and the opt-in findings comment
```
mfittko, line 32: **worth-fixing-now** (`dry`): The new rule asserts the opt-in findings comment's status as a coexisting second surface locally, directly beneath GATE-COMMENT-SINGLE-SURFACE, which states a gate round produces "exactly ONE new visible surface" and that "the single documented exception is the zero-dep fallback poster". The opt-in second surface is already owned and sanctioned in skills/docs/gate-review-sub-loop-contract.md (the post-gate-findings.mjs paragraph around line 622). One document now carries two unreconciled statements of how many visible surfaces a round may produce, which is exactly the weaker-restatement case STYLE-CONTRADICTION-LENS asks a contract-touching PR to check.

packages/core/src/github/copilot-helpers.mjs
mfittko: **defer** (`dry`): The findings-comment marker vocabulary is now spelled in a third place. buildFindingsMarker (scripts/github/post-gate-findings.mjs:227) is the producer, _gate-finding-surface.mjs:141/146 owns the review-header literal and its own matcher, and this regex re-spells both plus deferred-summary. The module's own header comment (line 20-24) states the opposite principle for the verdict header — owned once here "instead of restating it". Drift is currently caught only by the new end-to-end assertion in test/github/post-gate-findings.test.mjs, not by construction.

skills/docs/gate-review-comment-contract.md
mfittko: **worth-fixing-now** (`yagni`): The new GATE-COMMENT-IDENTITY-DISJOINT rule restates regex mechanics the code comment already owns, and its completeness claim is already stale: it says the summarizer "skips exactly the two machine-artifact marker bodies" while GATE_MACHINE_ARTIFACT_MARKER_RE has three branches (gate-findings-review, gate-findings\s, deferred-summary). The enumeration buys maintenance debt the rule does not need — a fourth artifact forces a doc edit, and the count is wrong on arrival.

skills/docs/gate-review-comment-contract.md
mfittko: Fixed in a7795afd: the incident reference is replaced with a plain description (silent destruction of the round's visible findings record), no PR number in durable prose.

skills/docs/gate-review-comment-contract.md
mfittko: Fixed in a7795afd: the rule defers the alternation to the regex's own comment instead of duplicating it in prose.

skills/docs/gate-review-comment-contract.md
mfittko: Fixed in a7795afd: the surface policy stays owned by the sub-loop contract; the rule now covers only comment identity.

skills/docs/gate-review-comment-contract.md
mfittko: Fixed in a7795afd: the rule's closing sentence now allows per-tool keying within its own vocabulary (the findings comment's gate-only marker is explicitly named as deliberate); the invariant is disjoint vocabularies, not per-tool key shape.

test/github/post-gate-findings.test.mjs
mfittko: Fixed in a7795afd: the boundary pin sits directly on isGateMachineArtifactBody (true for the three known markers, false for gate-findings-extra), which fails under the loose prefix regex — the summarizer-level probe was masked by the verdict-header carve-out.

packages/core/src/github/copilot-helpers.mjs
mfittko: Deferred (defer severity): consolidating the thrice-spelled marker vocabulary into one exported constant is tracked as a follow-up beyond this fix's scope.

skills/docs/gate-review-comment-contract.md
mfittko: Fixed in a7795afd: the rule states the verdict's claim key precisely (parsed gate+head verdict fields, with the review header on review surfaces) and defers the marker-set enumeration to the artifact filter that owns it.

skills/docs/gate-review-comment-contract.md
mfittko: Fixed in a7795afd: the stale count is gone — the rule names the filter as the owner rather than counting markers.

packages/core/src/github/copilot-helpers.mjs
mfittko: Fixed in a7795afd: the regex comment enumerates all three alternation branches including deferred-summary.

skills/docs/gate-review-comment-contract.md
mfittko: Fixed in a7795afd: the rule no longer enumerates a narrower skip set — it points at the artifact filter as the owner and states only the exact-marker/no-prefix property.

skills/docs/gate-review-comment-contract.md
mfittko: Fixed in a7795afd: the prohibition reads MUST NOT.

packages/core/src/github/copilot-helpers.mjs
```diff
@@ -51,7 +51,19 @@ export function matchGateReviewCommentHeader(body) {
...
+// findings-comment marker, that comment parses as a verdict marker candidate
+// (its "Gate fan-out findings:"/"Reviewed head:" lines yield gate+headSha)
+// and the verdict upsert claims and overwrites it in place, silently
+// destroying the round's visible findings record.
+// The alternation is exact-marker, not prefix: `gate-findings-review\b` (the
+// per-round review header), `gate-findings\s` (the findings COMMENT marker,
+// always followed by its ` gate=` attribute), and `deferred-summary\b` — a
+// bare `gate-findings\b` would also swallow any future `gate-findings-<x>`
+// marker silently.
+const GATE_MACHINE_ARTIFACT_MARKER_RE = /^<!--\s*dev-loops:(?:gate-findings-review\b|gate-findings\s|deferred-summary\b)/mu;
```
mfittko: **worth-fixing-now** (`contradiction-lens`): The added comment claims "The alternation is exact-marker, not prefix" and that exactness is what stops a future gate-findings-&lt;x> marker from being swallowed silently, but two of the three branches use \b, which IS a prefix match against a hyphen-suffixed name. Probed against the actual regex: &lt;!-- dev-loops:gate-findings-review-v2 ... --&gt; -> true and &lt;!-- dev-loops:deferred-summary-v2 --&gt; -> true (only gate-findings-extra -> false, which is the one branch the new test pins). The mirrored contract sentence in skills/docs/gate-review-comment-contract.md:38 states the same falsified claim absolutely ("an exact-marker match, never a gate-findings-&lt;x> prefix"); gate-findings-review-v2 is exactly such a name and it matches. So the stated invariant and the code disagree, and the test pins only the branch where they happen to agree.

skills/docs/gate-review-comment-contract.md
```diff
@@ -26,6 +26,21 @@ documented exception is the zero-dep fallback poster
...
+(`gates.postFindingsComments`, `post-gate-findings.mjs`) identify "their" comment by DISJOINT
+vocabularies, and each tool's upsert MUST NOT ever claim the other's comment. The verdict is
+claimed through its parsed verdict fields (gate name plus reviewed head, with the
+`dev-loops:gate-findings-review` header on review surfaces); the findings comment through its
```
mfittko: **worth-fixing-now** (`contradiction-lens`): Inside one paragraph the word "header" names two tokens with opposite effects. Line 33-34 calls dev-loops:gate-findings-review the verdict's "header on review surfaces"; line 36 says a marker body "with no genuine verdict header" is a non-candidate, where that header is the ### Gate review: line (GATE_REVIEW_COMMENT_HEADER_RE / matchGateReviewCommentHeader). In the code dev-loops:gate-findings-review is a MARKER whose presence DISQUALIFIES a body unless the ### Gate review: header rescues it (copilot-helpers.mjs:66-79). A reader implementing to this rule can conclude that emitting dev-loops:gate-findings-review is what makes a review claimable, which inverts the actual seam.

packages/core/src/github/copilot-helpers.mjs
```diff
@@ -51,7 +51,19 @@ export function matchGateReviewCommentHeader(body) {
 // example, describing this very mechanism) still counts as evidence. Both
 // producers render their marker at column 0, so the anchor costs nothing
```
mfittko, line 52: **defer** (`contradiction-lens`): The comment block the diff extends still counts two producers above the new three-member set: line 40 "Both are excluded HERE" and line 52 "Both producers render their marker at column 0" now under-describe a set that the added lines 54-65 enumerate as three (gate-findings-review, gate-findings, deferred-summary). The block contradicts itself about how many artifacts it covers.

skills/docs/gate-review-comment-contract.md
```diff
@@ -26,6 +26,21 @@ documented exception is the zero-dep fallback poster
...
+own `dev-loops:gate-findings gate=` marker. Enforced at the claim seam: the marker summarizer
+treats a machine-artifact marker body with no genuine verdict header as a non-candidate — the
+artifact filter in `copilot-helpers.mjs` owns the exact marker set (the findings comment's
+marker among them; an exact-marker match, never a `gate-findings-<x>` prefix) — so an upsert
+can never update the other tool's comment in place. That silent replacement previously
+destroyed a full round's visible findings record seconds after it was posted. Within its OWN
+vocabulary each tool keys identity as it needs (the findings comment's marker is deliberately
+gate-only); the invariant is that the vocabularies never overlap.
```
mfittko: **worth-fixing-now** (`docs`): GATE-COMMENT-IDENTITY-DISJOINT's stated invariant ("the vocabularies never overlap") is not what the code holds: the verdict review body itself carries &lt;!-- dev-loops:gate-findings-review ... --&gt; (scripts/github/_gate-finding-surface.mjs:141, emitted per upsert-checkpoint-verdict.mjs:916), and that marker IS a member of the artifact marker set the same rule cites (GATE_MACHINE_ARTIFACT_MARKER_RE, copilot-helpers.mjs:66). The markers overlap; disjointness of CLAIM is preserved only by the precedence carve-out in isGateMachineArtifactBody (a marker body with a genuine ### Gate review: header is the verdict, not an artifact). The same paragraph also attributes dev-loops:gate-findings-review to the verdict's claim vocabulary (line 34) while attributing it to the artifact filter's set two lines later, so the two halves disagree. A maintainer who trusts the literal invariant could drop the verdict-header carve-out and make the gate's own verdict review unclaimable.

skills/docs/gate-review-comment-contract.md
```diff
@@ -26,6 +26,21 @@ documented exception is the zero-dep fallback poster
 is absent: it cannot reach the review poster and posts a verdict issue comment, which the
 dev-loop skill already documents as a degraded audit-trail artifact.
 
+<!-- rule: GATE-COMMENT-IDENTITY-DISJOINT -->
+`GATE-COMMENT-IDENTITY-DISJOINT`: The verdict surface and the opt-in findings comment
```
mfittko, line 32: **worth-fixing-now** (`docs`): Contradiction-lens gap (STYLE-CONTRADICTION-LENS) inside one document: GATE-COMMENT-SINGLE-SURFACE three lines above states "A gate round produces exactly ONE new visible surface", "Each finding's text appears exactly once across the round", and "The single documented exception is the zero-dep fallback poster". The new rule normatively describes a second visible surface (the opt-in gates.postFindingsComments findings comment) that coexists with the verdict for the same round, and gate-review-sub-loop-contract.md:623-628 confirms it renders the same findings a SECOND time. With the toggle on, the SINGLE-SURFACE sentences and its exception count are false, and the two rules now sit adjacent in the same file with no reconciliation.

test/github/post-gate-findings.test.mjs
```diff
@@ -639,3 +639,61 @@ test("postGateFindings falls back to the default (skips) when the config fails t
...
+    { id: 2, body: verdictBody, html_url: "https://github.test/c/2", updated_at: "2026-08-04T00:00:00Z" },
+  ], { headSha: "a".repeat(40) });
+  assert.ok(verdictClaimed.draft_gate !== null);
+  // Exact-marker boundary, pinned on the filter itself (the summarizer-level
+  // probe is masked by the verdict-header carve-out, so only a direct pin is
+  // regex-sensitive): the artifact skip covers exactly the known markers; an
+  // unknown gate-findings-<x> marker is NOT an artifact (a loose
+  // `gate-findings\b` prefix match would swallow it silently).
+  const { isGateMachineArtifactBody } = await import("@dev-loops/core/github/copilot-helpers");
+  assert.equal(isGateMachineArtifactBody("<!-- dev-loops:gate-findings-review draft_gate " + "a".repeat(40) + " round=1 -->\nfindings"), true);
```
mfittko: **worth-fixing-now** (`dry`): The new isGateMachineArtifactBody pins duplicate the module's own owner test and put the new coverage in the wrong file. Lines 695 and 697 repeat, verbatim in substance, assertions that already exist at packages/core/test/copilot-helpers.test.mjs:305-306 (gate-findings-review marker -> true, deferred-summary marker -> true). The genuinely new pins introduced by this PR — the findings-comment marker alternative gate-findings\s and the exact-marker negative gate-findings-extra -> false (line 698) — live only here, in a caller's CLI test, while the owner test's own comment (packages/core/test/copilot-helpers.test.mjs:298-303) states the exclusion "must be pinned at THIS level, not only indirectly via one caller's CLI tests". Net result: two old assertions now exist in two files, and the regex alternative this PR actually added has no pin in the suite that owns the regex.

packages/core/src/github/copilot-helpers.mjs
mfittko: **defer** (`yagni`): The collision this PR closes only exists because the opt-in findings comment surface still exists. gates.postFindingsComments resolves false in every shipped layer (config.mjs:259 built-in, extension-defaults.yaml:148, artifact-authority-contract.md), and the code itself calls that comment "opt-in duplication" of the verdict review's own findings. Retiring post-gate-findings.mjs's comment posting would delete the second claim vocabulary, the artifact-filter alternative added here, and the GATE-COMMENT-IDENTITY-DISJOINT rule that exists to keep the two vocabularies apart. The guard added here is the correct minimal fix for THIS issue's ACs (coexistence), so this is a scope observation, not a blocker.

mfittko (commented):


**Reviewed head SHA:** `a7795afd2340038f2b4cef6b199123c835a65da9`
**Verdict:** findings_present
**Blocking severities:** must-fix (clean requires no findings matching these severities)

**Findings summary:** 7 angles reviewed; 11 findings (see per-angle breakdown below).

- `contradiction-lens` → `findings_present` (3 findings)
- `docs` → `findings_present` (2 findings)
- `dry` → `findings_present` (2 findings)

> **worth-fixing-now** (`acceptance-criteria`): The PR body's AC2 evidence claims the issue's second, stated-replacement branch ("the verdict comment states explicitly that earlier rounds' findings are no longer shown and where to find them"), but nothing states it. The opt-in findings comment is still keyed gate-only and PATCHed in place on every new head, and renderFindingsCommentBody emits only Reviewed head: &lt;sha> — no statement that this comment now shows a single round, and no pointer to where earlier rounds live. AC2 is still met through its FIRST branch (each head posts its own verdict review; upsert-checkpoint-verdict creates rather than updates on a different head, pinned by the different-head-SHA test in test/github/upsert-checkpoint-verdict.test.mjs), so this is an AC-evidence accuracy defect, not a behavior gap.
> Location: `scripts/github/post-gate-findings.mjs:292`


> **defer** (`dry`): Hand-copied duplicates of the budget-marker sentence still carry the pre-PR wording. upsert-checkpoint-verdict.test.mjs:3772, :4444 and :4563 embed ... — see the disposition ledger as fixture text, which no longer matches what buildAngleMarker (scripts/loop/consolidate-fanin.mjs:314) produces after this PR. Harmless (these are arbitrary fixture strings the verdict renderer never parses), but they are stale copies of a producer literal and now mislead a reader who takes them as examples of real marker text.
> Location: `test/github/upsert-checkpoint-verdict.test.mjs:3772`


> **worth-fixing-now** (`pr-checklist-matrix`): The AC/DoD/non-goals matrix covers only one non-goal. The PR body states two non-goals (single-surface/ledger/render-budget-tier changes; decoupling the ledger from the render budget, issue 1495) and the linked issue #1514 states three (changing the render budget or the MARKER SHAPE; making the ledger a committed artifact; changing which severities block a clean verdict). The matrix has exactly one non-goal row ('No single-surface/budget-tier changes'). The marker-shape non-goal is the material omission: this PR does change the marker text ('see the disposition ledger' -> 'in the disposition ledger', 'see ledger' -> 'in ledger', consolidate-fanin.mjs:198-201) and re-baselines the measured length comments in the budget tests, so a reviewer needs a matrix row stating why that is not the #1495 marker-shape work rather than no row at all.


> **worth-fixing-now** (`pr-checklist-matrix`): The AC2 row's evidence overclaims the branch it satisfies. The PR body and matrix say the opt-in findings comment is 'updated in place with its reviewed head shown - the stated-replacement branch of this AC'. Issue AC2's second branch requires an explicit statement that EARLIER rounds' findings are no longer shown and where to find them. renderFindingsCommentBody (scripts/github/post-gate-findings.mjs:285-297) emits only the marker, 'Gate fan-out findings: &lt;gate>' and 'Reviewed head: &lt;sha>' - it never says a previous round's record was replaced or where it went, so nothing on GitHub states the replacement. AC2 still holds on its FIRST branch (per-round reviews keep history, pinned by the round=N renderer pin at test/github/gate-finding-surface.test.mjs:97 and the N-heads-count-N-rounds test at test/github/close-gate-findings.test.mjs:317), so the status stays Met; only the cited evidence is wrong.

**Next action:** Fix pass: delimiter-anchored marker regex (truly exact), precedence-stated invariant reconciled with SINGLE-SURFACE, AC2 stated-replacement line in the findings comment, pins in the owner suite, matrix non-goal rows; PA round 3 at the new head.

packages/core/src/github/copilot-helpers.mjs
mfittko: Fixed in c447176f: every alternation branch is delimiter-anchored ((?=\s|-->)), proven by owner-suite cases including gate-findings-review-extra and deferred-summary-x as non-artifacts — the exact-marker claim now holds.

skills/docs/gate-review-comment-contract.md
mfittko: Fixed in c447176f: the rule reserves "heading" for the producer-owned verdict body heading and calls the review's marker the round marker; no token carries two meanings.

packages/core/src/github/copilot-helpers.mjs
mfittko: Deferred (defer severity): consolidating the marker vocabulary into one exported constant remains tracked as the round-1 deferral.

skills/docs/gate-review-comment-contract.md
mfittko: Fixed in c447176f: the rule states the mechanism honestly — machine-artifact filtering PLUS verdict-body precedence — which is exactly how the round's own marker-bearing review stays claimable.

skills/docs/gate-review-comment-contract.md
mfittko: Fixed in c447176f: the rule names the opt-in findings comment as the sanctioned second visible surface beside GATE-COMMENT-SINGLE-SURFACE's one-per-round review, reconciling the two rules.

test/github/post-gate-findings.test.mjs
mfittko: Fixed in c447176f: the isGateMachineArtifactBody boundary pins live in the filter's owner suite (copilot-helpers.test.mjs) with the delimiter cases; the cross-tool test keeps only the coexistence assertions.

packages/core/src/github/copilot-helpers.mjs
mfittko: Deferred (defer severity): retiring the opt-in duplicate findings-comment surface is a product decision beyond this fix; the invariant protects the surface that ships today.

skills/docs/gate-review-comment-contract.md
```diff
@@ -26,6 +26,22 @@ documented exception is the zero-dep fallback poster
...
+(`gates.postFindingsComments`, `post-gate-findings.mjs` — the sanctioned second visible surface
```
mfittko: **worth-fixing-now** (`docs`): The round-2 SINGLE-SURFACE reconciliation is one-directional and leaves the owning rule's absolute false. The added rule now normatively calls the opt-in findings comment "the sanctioned second visible surface that opt-in adds beside GATE-COMMENT-SINGLE-SURFACE's one-per-round review" (line 31-32), but GATE-COMMENT-SINGLE-SURFACE itself (line 14-21, unchanged by this PR) still asserts without qualification "A gate round produces exactly ONE new visible surface" and "Each finding's text appears exactly once across the round". With gates.postFindingsComments on, both sentences are false, and skills/docs/gate-review-sub-loop-contract.md:623-628 confirms the opt-in comment "renders the same findings a SECOND time" with "duplicated finding text on a second surface". The two rules sit 15 lines apart in one document, and the surface-count owner is the one still stating the absolute, so a reader entering at SINGLE-SURFACE (the rule that owns the count) is told the opposite of what the new rule sanctions. STYLE-PRECEDENCE-CONFLICT cannot resolve it because both rules live in this same document.

test/loop/consolidate-fanin.test.mjs
```diff
@@ -1750,7 +1750,7 @@ test("a narrow angle keeps its real finding instead of a longer marker when a wi
 // findings, its bare marker} renders cheaper in isolation — not "bare
 // always wins". A single-finding angle with no file/line and a summary at or
 // under the shrink floor renders its real form (62 chars, measured) SHORTER
-// than its own bare "N omitted — see ledger" marker (83 chars, measured):
+// than its own bare "N omitted — in ledger" marker (measured):
```
mfittko: **worth-fixing-now** (`docs`): The marker rewording dropped a measurement the surrounding comment's argument depends on. The comment states the seed comparison holds because the real form renders "(62 chars, measured)" SHORTER than its own bare marker, but the bare marker's counterpart figure was replaced with a bare "(measured)" when "see ledger" became "in ledger". One side of the inequality now carries a number and the other does not, so a future maintainer tuning this fixture cannot check the claim (or see how much headroom the fixture has) without re-deriving the cost. The rewording changed the marker by exactly one character ("see" -> "in"), so the real value was mechanically available.

packages/core/test/copilot-helpers.test.mjs
```diff
@@ -889,3 +889,21 @@ test("round-trip: sanitizing a verdict body before posting keeps the anti-summon
   assert.equal(sanitized, "Finding: this comment violates the `/copilot` prohibition rule.");
   assert.equal(containsBareCopilotSummon(sanitized), false, "sanitized text must not arm the anti-summon guard");
 });
+
+test("isGateMachineArtifactBody matches exactly the known marker tokens, delimiter-anchored", async (t) => {
```
mfittko: **worth-fixing-now** (`dry`): The new isGateMachineArtifactBody table duplicates the predicate's existing owner test instead of extending it. Its first and third rows re-assert what lines 305-306 already pin (gate-findings-review => true, deferred-summary => true), and it re-imports isGateMachineArtifactBody via await import() although the symbol is already statically imported at line 8. The split also left the neighbouring title at line 304 stale: it still says 'recognizes both machine-authored artifact markers' when the filter now covers three tokens.

packages/core/src/github/copilot-helpers.mjs
```diff
@@ -51,7 +51,19 @@ export function matchGateReviewCommentHeader(body) {
...
 // against genuine artifacts.
-const GATE_MACHINE_ARTIFACT_MARKER_RE = /^<!--\s*dev-loops:(?:gate-findings-review|deferred-summary)\b/mu;
+// The set covers the per-round review header (gate-findings-review),
```
mfittko: **worth-fixing-now** (`dry`): The comment block added above GATE_MACHINE_ARTIFACT_MARKER_RE enumerates the same three marker tokens twice within twelve lines (lines 54-56 'The set covers the per-round review header ... and the historical deferred-summary comment', then lines 62-65 'matches exactly the three known marker tokens (the per-round review round marker, the findings COMMENT marker ..., and the historical deferred-summary comment)'). Two maintenance sites for one list, in one block, directly above the regex that already states the list.

mfittko (commented):


**Reviewed head SHA:** `c447176f0bd3a4bcdc9cf04004fbef614b4772ae`
**Verdict:** clean

**Findings summary:** 7 angles reviewed; 11 findings (see per-angle breakdown below).

- `contradiction-lens` → `findings_present` (3 findings)
- `docs` → `findings_present` (2 findings)
- `dry` → `findings_present` (3 findings)
- `pr-checklist-matrix` → `findings_present` (3 findings)
- `yagni` → `clean`

> **worth-fixing-now** (`contradiction-lens`): The new GATE-COMMENT-IDENTITY-DISJOINT rule contradicts GATE-COMMENT-SINGLE-SURFACE four lines above it in the same file. SINGLE-SURFACE states a gate round produces "exactly ONE new visible surface" and that "The single documented exception is the zero-dep fallback poster"; the new rule calls the opt-in findings comment "the sanctioned second visible surface that opt-in adds beside GATE-COMMENT-SINGLE-SURFACE's one-per-round review". A reader following SINGLE-SURFACE literally concludes post-gate-findings.mjs is prohibited; a reader following the new rule concludes it is sanctioned. The mirror .claude/skills/docs/gate-review-comment-contract.md carries the same pair.
> Location: `skills/docs/gate-review-comment-contract.md:23`


> **worth-fixing-now** (`contradiction-lens`): gate-review-sub-loop-contract.md still describes the machine-artifact exclusion set as exactly two HISTORICAL artifacts: "Only a marker-bearing body with no genuine verdict header - a historical standalone findings review, or a historical &lt;!-- dev-loops:deferred-summary --&gt; comment - stays excluded". This diff adds a third, current, sanctioned member (post-gate-findings.mjs's dev-loops:gate-findings marker) to GATE_MACHINE_ARTIFACT_MARKER_RE, and the new GATE-COMMENT-IDENTITY-DISJOINT rule depends on that exclusion being current behavior. The two contracts now describe the same seam incompatibly, and the sub-loop text tells a reader the live findings comment is not excluded.
> Location: `skills/docs/gate-review-sub-loop-contract.md:838`


> **defer** (`contradiction-lens`): Cross-suite fixtures still quote the retired marker literal "see the disposition ledger" as simulated consolidate-fanin output (test/github/upsert-checkpoint-verdict.test.mjs:3772, :4444, :4563). consolidate-fanin can no longer emit that string, so these fixtures now assert verdict-render behavior against producer text that does not exist. Harmless today (they are inputs, not producer assertions) but they contradict the reworded producer and will read as the live marker shape to the next maintainer.
> Location: `test/github/upsert-checkpoint-verdict.test.mjs:3772`


> **worth-fixing-now** (`dry`): Three copies of the fan-in marker sentence still carry the retired wording '— see the disposition ledger' (lines 3772, 4444, 4563). consolidate-fanin can no longer emit that string, so these consumer fixtures now stand in for producer output that does not exist, and the wording the issue calls a dead pointer stays alive in the test corpus. The copy-paste of the marker sentence across suites is what let it drift: this PR updated five sites in test/loop/consolidate-fanin.test.mjs and missed these three.
> Location: `test/github/upsert-checkpoint-verdict.test.mjs:3772`


> **worth-fixing-now** (`pr-checklist-matrix`): Matrix evidence cell is stale at this head: the "Coexistence test | DoD | Met" row cites "post-gate-findings.test.mjs (40/40)", but the file carries 41 top-level tests at c447176f (39 on origin/main, 40 at f33576d3, 41 at head). The count was accurate when written and was not updated when the AC2 stated-replacement pin ("renderFindingsCommentBody states that only the latest round is shown", test/github/post-gate-findings.test.mjs:696) landed, so the matrix's own AC2 evidence is not counted by the row a reader would reconcile it against.


> **worth-fixing-now** (`pr-checklist-matrix`): Non-goal traceability is incomplete in both the Non-goals section and the matrix. The issue states three non-goals; the PR body restates only a reworded first one and drops its "or the marker shape" clause, and the matrix carries a single non-goal row ("No single-surface/budget-tier changes — tiers and flow untouched"). The issue's "Making the disposition ledger a committed artifact" and "Changing what severities block a clean verdict" have no row at all. The dropped clause is exactly where this diff comes closest to the boundary: consolidate-fanin.mjs rewords both the verbose and bare per-angle marker summaries and therefore shifts their measured render lengths, and no row states whether that was judged in-scope (AC3 requires the rewording) or a non-goal breach. A reviewer reconciling matrix against issue cannot tell which.


> **defer** (`pr-checklist-matrix`): The PR's DoD item 1 narrows the issue's DoD item 1 without flagging the substitution. The issue asks for "a test &#91;that] drives post-findings followed by upsert-verdict for one gate/head and asserts both contents are still retrievable afterwards"; the delivered test asserts the claim seam only (summarizeGateReviewCommentMarkers skips the findings body; the verdict body carries no findings marker) and never runs the two CLIs in sequence against a shared comment store. The substitution is defensible (the claim seam is the root cause, and the file already has a writeGhStub harness so the sequenced variant is possible but heavy for upsert-checkpoint-verdict's gh surface), but the matrix marks the row Met against the PR's own reworded wording, so the narrowing is invisible when checked against the issue.

**Next action:** Fix pass (last wfn round): both exceptions in SINGLE-SURFACE, sub-loop artifact-set count, merged regex comment, measured bare-marker note, fixture strings, deduped pins, regenerated matrix; PA round 4 at the new head.

packages/core/src/github/copilot-helpers.mjs
mfittko: Fixed in 2e191e03: the regex comment is one merged paragraph with a single token enumeration; the three stale fixture strings in the verdict suite now carry the shipped wording.

skills/docs/gate-review-comment-contract.md
mfittko: Fixed in 2e191e03: GATE-COMMENT-SINGLE-SURFACE itself names both documented exceptions (the opt-in findings comment and the zero-dep fallback), so the reconciliation is two-way.

test/loop/consolidate-fanin.test.mjs
mfittko: Fixed in 2e191e03: the bare-marker measurement is stated with its number again (22 chars at N=30, measured).

packages/core/test/copilot-helpers.test.mjs
mfittko: Fixed in 2e191e03: the boundary pins are merged into the existing owner test (one test, all cases); the standalone duplicate is deleted and the cross-tool test keeps only coexistence.

test/loop/consolidate-fanin.test.mjs
```diff
@@ -1750,7 +1750,7 @@ test("a narrow angle keeps its real finding instead of a longer marker when a wi
 // findings, its bare marker} renders cheaper in isolation — not "bare
 // always wins". A single-finding angle with no file/line and a summary at or
 // under the shrink floor renders its real form (62 chars, measured) SHORTER
-// than its own bare "N omitted — see ledger" marker (83 chars, measured):
+// than its own bare "N omitted — in ledger" marker (22 chars at N=30, measured):
```
mfittko, line 1753: **worth-fixing-now** (`acceptance-criteria`): The AC3 rewording's own measurement evidence is wrong and self-contradicting: the seed-comparison annotation now reads "renders its real form (62 chars, measured) SHORTER than its own bare \"N omitted — in ledger\" marker (22 chars at N=30, measured)". 22 &lt; 62 contradicts the sentence's conclusion, N=30 is not this fixture's N (the narrow angle carries ONE finding), and 22 is the raw summary string length while the compared quantity is angleRenderCost (the rendered per-angle section length, consolidate-fanin.mjs:355). Measured at this head for this fixture: real = 50, bare = 85. The previous 62/83 pair was at least internally consistent as render costs; the round-3 edit replaced the bare number with a different quantity for a different N, so the comment no longer explains why the seed picks real and the PR's "no render-budget change, measured boundaries stable" evidence row rests on a broken annotation.

mfittko, line 1753: **worth-fixing-now** (`contradiction-lens`): The reworded seed-comparison comment now asserts its own opposite: it says the real form (62 chars, measured) renders SHORTER than the bare marker (22 chars at N=30, measured) — 62 is not shorter than 22. The original pair (62 vs 83) were both angleRenderCost values (the rendered angle SECTION length, consolidate-fanin.mjs:355-357); the rewrite swapped the marker side for the raw marker-string length, so the two numbers are no longer the same measure and the inequality the comment exists to explain reads false. 'at N=30' is also wrong for this fixture: the narrow angle carries exactly one finding (line 1767), so its bare marker is '1 omitted — in ledger'.

packages/core/src/github/copilot-helpers.mjs
```diff
@@ -51,7 +51,17 @@ export function matchGateReviewCommentHeader(body) {
...
+// marker (gate-findings-review), post-gate-findings.mjs's opt-in findings
+// COMMENT marker (gate-findings gate=...), and the historical
+// deferred-summary comment. Without the findings-comment marker, that comment
+// parses as a verdict marker candidate (its "Gate fan-out findings:"/
+// "Reviewed head:" lines yield gate+headSha) and the verdict upsert claims
+// and overwrites it in place, silently destroying the round's visible
+// findings record. Every branch is delimiter-anchored — the token must be
+// followed by whitespace or the closing `-->` — so no suffixed `<token>-<x>`
+// variant ever matches.
+const GATE_MACHINE_ARTIFACT_MARKER_RE = /^<!--\s*dev-loops:(?:gate-findings-review|gate-findings|deferred-summary)(?=\s|-->)/mu;
```
mfittko, line 64: **worth-fixing-now** (`contradiction-lens`): The artifact-marker set grew to a third member that is CURRENT, not historical, but three prose enumerations of that same set were left at two members and now contradict the docs this PR rewrote. gate-review-sub-loop-contract.md:836-840 (changed here) lists three excluded bodies including 'the current opt-in findings comment', while copilot-helpers.mjs:33-46 still ends 'Both are excluded HERE', copilot-helpers.mjs:73-75 still narrows the excluded set to '(a historical standalone findings review or deferred-summary comment)', and scripts/github/detect-checkpoint-evidence.mjs:12-14 still calls the whole set 'Historical machine-authored gate artifacts (a standalone findings review, a deferred-summary comment)'. A reader of isGateMachineArtifactBody's own doc comment is told the live opt-in findings comment is not in scope, which is exactly the case this PR added.

scripts/github/post-gate-findings.mjs
```diff
@@ -291,6 +291,8 @@ export function renderFindingsCommentBody({ gate, headSha, findings }) {
     // Plain text head SHA (no backticks) so GitHub autolinks the commit.
     `Reviewed head: ${headSha}`,
     "",
+    "This comment shows only the latest posted round for this gate; earlier rounds' findings are no longer shown here and live on their own per-round gate reviews.",
```
mfittko, line 294: **worth-fixing-now** (`contradiction-lens`): The AC2 stated-replacement sentence redirects readers to a location that is wrong in exactly the case the issue is about. It promises earlier rounds' findings 'live on their own per-round gate reviews', but consolidate-fanin collapses an over-budget round's per-angle findings to 'N omitted — in ledger' markers (buildAngleMarker) and tier 4 withholds them entirely — consolidate-fanin.mjs:1086-1092, reworded in this same diff, states that a degraded round's only durable record is --ledger-out. Issue 1514's motivating case (PR 1513, 32 findings) was over budget, so for that class of earlier round the per-round review does not carry the findings at all, and the AC's 'and where to find them' branch is satisfied with an inaccurate where.

packages/core/src/github/copilot-helpers.mjs
```diff
@@ -51,7 +51,17 @@ export function matchGateReviewCommentHeader(body) {
...
-const GATE_MACHINE_ARTIFACT_MARKER_RE = /^<!--\s*dev-loops:(?:gate-findings-review|deferred-summary)\b/mu;
+// The set covers exactly three marker tokens: the per-round review round
+// marker (gate-findings-review), post-gate-findings.mjs's opt-in findings
+// COMMENT marker (gate-findings gate=...), and the historical
+// deferred-summary comment. Without the findings-comment marker, that comment
+// parses as a verdict marker candidate (its "Gate fan-out findings:"/
+// "Reviewed head:" lines yield gate+headSha) and the verdict upsert claims
+// and overwrites it in place, silently destroying the round's visible
+// findings record. Every branch is delimiter-anchored — the token must be
+// followed by whitespace or the closing `-->` — so no suffixed `<token>-<x>`
```
mfittko, line 62: **defer** (`contradiction-lens`): The new anchoring comment overstates its own rule: 'no suffixed &lt;token>-&lt;x> variant ever matches' is false for a member of the set it guards — gate-findings-review is a suffixed variant of gate-findings and it matches, as its own token. The docs carry the same absolute (gate-review-comment-contract.md:40, 'delimiter-anchored so no suffixed &lt;token>-&lt;x> variant matches'). The pinned test asserts both halves side by side (gate-findings-review true, gate-findings-extra false), so the code is right and only the claim is over-broad.

mfittko (commented):


**Reviewed head SHA:** `2e191e0347f5b2e14c8d2c8b1eeb366a88e8edc8`
**Verdict:** clean

**Findings summary:** 7 angles reviewed; 10 findings (see per-angle breakdown below).

- `contradiction-lens` → `findings_present` (4 findings)
- `docs` → `findings_present` (1 finding)
- `dry` → `findings_present` (3 findings)
- `yagni` → `clean`

> **defer** (`docs`): Enumeration drift between the updated contract text and its in-code mirror: the sub-loop contract's exclusion sentence was updated to name three excluded marker-bearing classes (standalone findings review, deferred-summary, opt-in findings comment), but the carve-out comment inside isGateMachineArtifactBody in packages/core/src/github/copilot-helpers.mjs (the 'Only a marker-bearing body with NO genuine verdict header (a historical standalone findings review or deferred-summary comment) stays excluded' line, just outside this PR's hunk) still lists only the two historical classes. The behavior is correct and the newly added comment above the regex is accurate; only that older sentence is now an incomplete enumeration.


> **worth-fixing-now** (`dry`): The marker-text rewording left three copies of the old vocabulary behind as negative assertions that can no longer fail. test/loop/consolidate-fanin.test.mjs lines 1404, 1789 and 1835 still assert !/omitted.*see (the disposition )?ledger/, but no production string contains "see" any more (buildAngleMarker now emits "— in the disposition ledger" / "— in ledger"). All three are the same duplicated literal the diff updated at line 1736 (to /omitted.*ledger/) and missed elsewhere, so three marker-collapse regression guards now pass unconditionally: an angle that regressed to a marker would still satisfy them.
> Location: `test/loop/consolidate-fanin.test.mjs`

**Next action:** PA clean at round 4; remaining comment-level wfn auto-defers per the round-cap rule; merge (pre-authorized).

test/loop/consolidate-fanin.test.mjs
mfittko, line 1753: **defer** (`dry`): The rewrite of the seed-comparison comment replaced a comparable measurement with an incomparable one, making the comment contradict itself. The original read "real form (62 chars, measured) SHORTER than its own bare marker (83 chars, measured)" — both angleRenderCost values (rendered whole-angle section length, consolidate-fanin.mjs:355). The new text keeps "62 chars" but gives the bare marker as "22 chars at N=30", which is the raw summary-string length at an N this test's fixture does not use (its narrow angle has one finding, summary "x"). As written the comment asserts 62 is SHORTER than 22 and cites an N from a different test.

test/github/post-gate-findings.test.mjs
```diff
@@ -639,3 +639,62 @@ test("postGateFindings falls back to the default (skips) when the config fails t
...
   }
 });
+
+// #1514 coexistence pin: the opt-in findings comment and the verdict surface
+// must never claim each other. Each tool locates "its" comment by its own
+// marker vocabulary; these assertions fail if either vocabulary grows to
+// match the other tool's rendered body (the class of silent overwrite
+// observed live on PR 1513 before the single-surface rework).
+test("findings comment and verdict body are mutually unclaimable (#1514)", async () => {
+  const { renderFindingsCommentBody, buildFindingsMarker } = await import("../../scripts/github/post-gate-findings.mjs");
```
mfittko, line 649: **defer** (`dry`): The new coexistence test re-imports modules the file already has in scope. renderFindingsCommentBody and buildFindingsMarker are statically imported at the top of the same file (lines 10 and 15) yet re-imported dynamically at line 649, shadowing those bindings; @dev-loops/core/github/copilot-helpers is imported twice inside the one test body (lines 651 and 692).

test/loop/consolidate-fanin.test.mjs
mfittko, line 1753: **worth-fixing-now** (`pr-checklist-matrix`): The AC3 matrix row's evidence ("fixture strings aligned") and the PR summary's claim that the reworded marker texts have "near-identical lengths" are contradicted by the alignment edit the head commit actually made: the seed-comparison pin's comment now reads "real form (62 chars, measured) SHORTER than its own bare 'N omitted — in ledger' marker (22 chars at N=30, measured)" and then states "angleRenderCost(real) &lt;= angleRenderCost(bareMarker) is true". 62 &lt;= 22 is false as written. The two numbers measure different things (62 is angleRenderCost, i.e. renderStructuredFindings length; 22 is the raw marker string length) and cite different N (this test's narrow angle carries ONE finding with summary "x", not 30). The pre-head revision (2e191e03^) had the coherent form "(measured)" with no number after an earlier round dropped the old "83 chars" angleRenderCost figure; the head commit re-added a number from the wrong measure, so the pin's stated rationale no longer supports the assertion it guards. Behavior is unaffected and full verify is green at this head.

mfittko: Deferred at gate close (round 4, fingerprint ca700167a38193ff, severity worth-fixing-now, angle acceptance-criteria): stayed open past this gate chain's round-3 worth-fixing-now fix window; the deferral is recorded on this thread's marker and in the durable findings ledger.

mfittko: Deferred at gate close (round 4, fingerprint 868b31a26270baa5, severity worth-fixing-now, angle contradiction-lens): stayed open past this gate chain's round-3 worth-fixing-now fix window; the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/github/copilot-helpers.mjs
mfittko: Deferred at gate close (round 4, fingerprint b0f89c2ea002ad34, severity worth-fixing-now, angle contradiction-lens): stayed open past this gate chain's round-3 worth-fixing-now fix window; the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/post-gate-findings.mjs
mfittko: Deferred at gate close (round 4, fingerprint e11378f8f3db899b, severity worth-fixing-now, angle contradiction-lens): stayed open past this gate chain's round-3 worth-fixing-now fix window; the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/github/copilot-helpers.mjs
mfittko: Deferred at gate close (round 4, fingerprint e85370d04ca1c35c, severity defer, angle contradiction-lens): defer-severity findings are deferred immediately, at the round they are first posted; the deferral is recorded on this thread's marker and in the durable findings ledger.

test/loop/consolidate-fanin.test.mjs
mfittko: Deferred at gate close (round 4, fingerprint 97237a90ab8924b3, severity defer, angle dry): defer-severity findings are deferred immediately, at the round they are first posted; the deferral is recorded on this thread's marker and in the durable findings ledger.

test/github/post-gate-findings.test.mjs
mfittko: Deferred at gate close (round 4, fingerprint 7aeb15f64e0ede4a, severity defer, angle dry): defer-severity findings are deferred immediately, at the round they are first posted; the deferral is recorded on this thread's marker and in the durable findings ledger.

test/loop/consolidate-fanin.test.mjs
mfittko: Deferred at gate close (round 4, fingerprint 9fb3aaf4c50404f9, severity worth-fixing-now, angle pr-checklist-matrix): stayed open past this gate chain's round-3 worth-fixing-now fix window; the deferral is recorded on this thread's marker and in the durable findings ledger.

Issue #1585, opened 2026-08-06
Fixer-triage for all gate findings (incl. nice-to-have) + require resolution of every review comment before gate close


mfittko:
## Context

During the #1583 drive (PR #1584), the draft_gate posted a clean verdict with 2 `nice-to-have` findings as inline review threads, then the dev-loop marked the PR ready (`draftGateSatisfied: true`) and proceeded to Copilot review — **leaving the 2 nice-to-have threads unresolved**. The pre-merge evidence check (0 unresolved threads) only catches this much later, at the merge boundary.

Root cause: nice-to-haves are auto-deferred by `close-gate-findings.mjs` (the disposition pass), but the fixer never sees them as fix targets, and `draftGateSatisfied` / gate-close is computed from the gate *verdict* (clean = no blocking must-fix/WFN) — it does **not** require all gate-authored threads to be resolved. Concretely: all `draftGateSatisfied` computations (`scripts/loop/pre-pr-ready-gate.mjs`, `scripts/github/ready-for-review.mjs`, `scripts/github/detect-checkpoint-evidence.mjs`) route through `fetchDraftGateEvidence` in `scripts/github/_gate-finding-surface.mjs`, which reads only verdict comments and never inspects review threads. The disposition pass is invoked only from skill prose; no script verifies it ran. So if the disposition pass skips or fails to run for a round, nice-to-have threads dangle silently and the gate still reports satisfied.

This violates `GATE-EXEC-THREAD-DISPOSITION` ("a nice-to-have finding is replied to and resolved immediately, at the same round it was first posted") and leaves the PR in a state that blocks merge later.

## Decided approach

1. **Route every gate-authored review comment to the fixer as a fix/triage target** — must-fix, worth-fixing-now, AND nice-to-have. Today the fixer only sees blocking findings; nice-to-haves go straight to silent auto-defer. The fixer gets triage authority over nice-to-haves: fix-if-cheap-in-the-same-commit (free polish when already touching that code), else defer. Defer is allowed from round 1 on for nice-to-haves (no fix window required).
[...]

Issue #891, opened 2026-06-26
Gate deadlock: a non-draft PR can neither post draft_gate ("already satisfied") nor pass the pre-merge check ("missing visible clean draft_gate comment")


mfittko:
## Summary

When a PR is **opened directly as ready-for-review** (not via the draft→ready flow), the dev-loop reaches a deadlock at the draft-gate boundary:

- `upsert-checkpoint-verdict --gate draft_gate` **refuses to post**, throwing:
  > Cannot enter draft_gate on `<repo>#<n>`: draft gate was already satisfied (clean evidence exists, PR is no longer draft). Do not re-post draft_gate. The draft→ready transition was already recorded.
- …but `detect-checkpoint-evidence` **fails the pre-merge gate check**:
  > preMergeGateCheck.ok: false — failures: ["missing visible clean draft_gate comment"]

So the poster believes the draft gate is satisfied and won't write the comment, while the pre-merge evidence check fails closed because that very comment does not exist. The PR cannot merge.

The two checks disagree on what constitutes draft-gate satisfaction.

## Reproduction (observed on release PR #889)

1. `gh pr create … ` **without `--draft`** (PR opens as ready).
2. Post `pre_approval_gate` verdict (succeeds — clean, current head).
3. Attempt `upsert-checkpoint-verdict --gate draft_gate --verdict clean` → throws **"draft gate was already satisfied … Do not re-post"** even though **no `draft_gate` comment exists yet**.
4. `detect-checkpoint-evidence` → `preMergeGateCheck.ok:false`, `failures:["missing visible clean draft_gate comment"]`.

Confirmed artifact state during the deadlock: PR #889 had **zero** `draft_gate` comments; the only gate comment present was the `pre_approval_gate` one. (A `draft_gate` comment only appeared after the manual workaround below.)

## Workaround (what unblocked the release)

Force the PR back through the intended draft→ready transition:

```
gh pr ready <n> --undo            # convert ready → draft
upsert-checkpoint-verdict --gate draft_gate --verdict clean …   # now succeeds
ready-for-review / gh pr ready <n>   # draft → ready
detect-checkpoint-evidence …      # preMergeGateCheck.ok: true
```

This is exactly the toggling the gate machinery should not require.

## Root-cause pointers

- **Poster refusal:** `scripts/github/upsert-checkpoint-verdict.mjs:616` throws when `coordination.draftGateAlreadySatisfied` is true.
[...]

PR #1630, opened 2026-08-09
fix(gate): enforce requireFanoutEvidence at post time, not just at merge time (#1599)

### LINKED ISSUE #1599: Contract breach: requireFanoutEvidence is not enforced at post-time — upsert-checkpoint-verdict accepts inline clean verdicts that fail the merge-time gate
## Summary

`gates.requireFanoutEvidence: true` is enforced **reactively** (merge-time pre-merge evidence check) but **not preventively** (post-time). The gate-verdict producer (`upsert-checkpoint-verdict.mjs`) accepts any `inline_single_agent` verdict with a non-empty `--inline-reason`, without checking whether the PR satisfies the repo's own required-evidence config. This permits a "clean" gate verdict to be **posted as visible merge evidence on the PR** when it does not actually satisfy `requireFanoutEvidence` — a contract breach at the produce step that the outer merge-time net only catches later.

## Observed (live)

PR #1598 (issue #1588, 10 files / 528 changed lines — well over the `lightMode` threshold of 2 files / 20 lines) posted its `draft_gate` verdict as:

> **Execution mode:** inline_single_agent — dev-loop self-drive: inline review by developer subagent + full test suite verification

with `gates.requireFanoutEvidence: true` set in `.devloops`. The post was **accepted** by `upsert-checkpoint-verdict.mjs`. The verdict now sits on the PR as a visible "clean draft_gate" — but it does not satisfy `requireFanoutEvidence` and will be **rejected** by `detect-checkpoint-evidence.mjs` (`buildPreMergeGateCheck`) at the pre-merge evidence check:

```
draft_gate: requireFanoutEvidence is enabled but executionMode is "inline_single_agent" (expected "fanout_fanin"); inline gate verdicts are not accepted
```

## Contract analysis

The canonical rule (skills/docs/gate-review-sub-loop-contract.md, "Execution mode and fan-out evidence enforcement") states:

> Fan-out evidence enforcement is ON by default (`gates.requireFanoutEvidence`): a clean gate verdict requires the gate to run via `--execution-mode fanout_fanin` with a findings-log ledger for the head SHA, and **the pre-merge evidence check fails closed** for a required gate otherwise.

So enforcement is **by design** placed at the pre-merge evidence check (merge-time / `detect-checkpoint-evidence.mjs`), not at post-time. The produce step (`upsert-checkpoint-verdict.mjs`) has **zero** enforcement of `requireFanoutEvidence`:

- Default `executionMode` = `inline_single_agent` (`DEFAULT_EXECUTION_MODE`).
- Accepts `--execution-mode inline_single_agent` with any non-empty `--inline-reason`.
- Does NOT read `requireFanoutEvidence` at all (its only two mentions are comments about the draft-transition re-post path).
- Does NOT re-derive scope, does NOT check `lightMode.enabled`, does NOT check the `gate:full` label.

The only sanctioned inline-acceptance carve-out is **light-mode under-threshold micro-PRs** (#1174): `lightMode.enabled: true` + scope re-derived under `maxFiles`/`maxLines` + no `gate:full` label + non-empty inline reason. PR #1598 meets none of the scope criteria (10 files / 528 lines).

The acceptance predicate already exists as one exported function: `buildFanoutEnforcement` (`scripts/github/detect-checkpoint-evidence.mjs`) resolves `{ required, lightMode, hasFullLabel, gates }` from config + `gate:full` label + fail-closed merge-base scope re-derivation (`detectMergeBaseScope` / `isEligibleForLightMode`, `scripts/loop/detect-change-scope.mjs`) per gate marker. Merge-time consumes it via `buildPreMergeGateCheck`.

## Why this is a breach (spirit) and a gap (letter)

- **By the letter of the contract:** the merge-time fail-closed check will block the merge — the safety net holds, no harmful merge escapes.
- **By the spirit of `requireFanoutEvidence`:** the system PERMITTED a "clean" gate verdict to be PRODUCED and POSTED as visible evidence when it does not satisfy the repo's required-evidence config. Between post and merge, the PR carries a misleading `✅ clean draft_gate` that is not actually mergeable evidence. The dev-loop can post an inline clean verdict and only discover the problem at the merge gate — wasting a cycle (re-post via the #891 draft-transition path) and momentarily misrepresenting gate satisfaction.

"If we have require fan-out, it MUST enforce fan-out" — and "enforce" should mean **refuse to record an inline verdict at post-time** when the PR doesn't qualify for the light-mode exception, not just reject it later at merge.

## Decided approach

Enforce `requireFanoutEvidence` at post-time in `upsert-checkpoint-verdict.mjs` by **reusing the merge-time predicate, not mirroring it**:

1. **Shared predicate.** Call the exported `buildFanoutEnforcement` (`detect-checkpoint-evidence.mjs`) with a candidate marker built from the to-be-posted verdict (`executionMode`, head SHA, gate). One acceptance predicate serves both boundaries, so post-time and merge-time can never drift. If the helper's marker-shaped input needs a small refactor to accept a candidate (not-yet-posted) verdict, that refactor is in scope; duplicating the threshold/label/scope logic is not. `upsert-checkpoint-verdict.mjs` already loads config (`loadDevLoopConfig`/`resolveGateConfig`) and PR context, so the label and base ref are available.
2. **Refusal scope: every inline post for a required gate, regardless of verdict value.** Mode qualification is independent of whether the verdict is `clean`, `findings_present`, or `blocked` — an over-threshold inline round is wrong-mode work whatever it concludes, and failing at the first post saves the whole wasted round. When `requireFanoutEvidence: true` AND `executionMode == inline_single_agent` AND the PR does not meet the light-mode acceptance criteria (per `buildFanoutEnforcement`: `lightMode.enabled`, scope under threshold via fail-closed re-derivation, no `gate:full` label, non-empty `--inline-reason`), the post is refused with an error mirroring the merge-time rejection message.
3. **Fail closed on un-derivable scope** (same as merge-time: no base ref or failed re-derivation → inline not accepted).
4. **No override flag.** The `--force` family is already a removed-flag set in this script; post-time enforcement gets no escape hatch. `requireFanoutEvidence: false` remains the only opt-out.

This makes enforcement **preventive**, not just reactive.

## Acceptance criteria

- [ ] `upsert-checkpoint-verdict.mjs` enforces `gates.requireFanoutEvidence` at post-time by calling the exported `buildFanoutEnforcement` predicate (shared with merge-time); no duplicated threshold/label/scope logic.
- [ ] When `requireFanoutEvidence: true` AND `executionMode == inline_single_agent`, the post is refused unless the PR meets the light-mode acceptance criteria (scope re-derived fail-closed via `detectMergeBaseScope`, `lightMode.enabled`, no `gate:full` label, non-empty `--inline-reason`).
- [ ] Refusal applies to every inline verdict value (`clean`, `findings_present`, `blocked`) for a required gate — not only clean posts.
- [ ] Over-threshold inline post → refused with a clear error naming the failure reason (mirrors the merge-time rejection message).
- [ ] Under-threshold light-mode inline post → accepted (unchanged behavior).
- [ ] `fanout_fanin` posts → unaffected, including the #891 draft-transition re-post path.
- [ ] Tests: post-time refusal on over-threshold inline when `requireFanoutEvidence: true` (each verdict value); acceptance on under-threshold light-mode inline; acceptance of `fanout_fanin` regardless; pass-through when `requireFanoutEvidence: false`; fail-closed refusal on un-derivable scope; post-time and merge-time agree on one shared fixture.
- [ ] `npm run verify` introduces zero new failures vs baseline.

## Definition of done

- Post-time enforcement live in `upsert-checkpoint-verdict.mjs`, reusing `buildFanoutEnforcement`; the PR-1598 scenario (over-threshold inline clean post with `requireFanoutEvidence: true`) is refused at post-time.
- `skills/docs/gate-review-sub-loop-contract.md` "Execution mode and fan-out evidence enforcement" updated: enforcement is post-time AND merge-time (produce step refuses; pre-merge check remains the fail-closed net).
- Tests per the acceptance criteria.
- CHANGELOG entry.

## Non-goals

- Changing the merge-time `buildPreMergeGateCheck` logic (already correct; it stays as the outer net).
- Changing `requireFanoutEvidence`'s default or the light-mode acceptance criteria.
- Affecting `fanout_fanin` posts (unaffected — only the inline path gains post-time enforcement).
- An override/escape flag for the post-time refusal (`requireFanoutEvidence: false` is the only opt-out).

## Context

- Discovered while driving PR #1598 (issue #1588) through the gate path during the rc.4 release cycle.
- Related: #1174 (light-mode inline acceptance — merge-time), #1043 (light-mode dispatch), #891 (draft-transition fanout re-post carve-out), #1585 (fixer-triage + 0-unresolved-before-gate-close).

mfittko:
Closes #1599

## Summary

`gates.requireFanoutEvidence` was enforced reactively at merge time and not at post time. `upsert-checkpoint-verdict.mjs` accepted an `inline_single_agent` verdict of any value regardless of whether the PR qualified for the light-mode carve-out, posting it as visible "clean gate" evidence that the pre-merge check would only reject later. Between post and merge the PR carried a misleading gate verdict that was never valid.

Enforcement is now preventive, reusing the merge-time predicate rather than mirroring it.

## Mechanism

- `detect-checkpoint-evidence.mjs`: the mode-qualification block that lived inline inside `buildPreMergeGateCheck` is extracted into an exported `evaluateInlineFanoutMode(gate, fanoutEnforcement)`. `buildPreMergeGateCheck` now calls it. Behavior is byte-identical; the existing 88 tests pass unchanged.
- `upsert-checkpoint-verdict.mjs`: `enforcePostTimeFanoutMode` runs right after the gate-entry check, before any verdict-specific logic, so it fires uniformly for `clean`, `findings_present`, and `blocked`. It builds a candidate marker for the not-yet-posted verdict, lazily fetches `baseRefOid`/labels only for an inline candidate under active enforcement, and calls the same `buildFanoutEnforcement` + `evaluateInlineFanoutMode` that merge time uses.

One predicate serves both boundaries, so post-time and merge-time cannot drift. No override flag; `requireFanoutEvidence: false` remains the only opt-out.

## Behavior change worth reviewing

`reconcile-draft-gate.mjs` only ever posts inline, so it is now refused for any PR over the light-mode threshold. That is the intended semantics rather than a regression: the tool synthesizes a clean `draft_gate` verdict for a PR that never ran one, which is exactly the unearned evidence this change exists to stop.

The refusal is made actionable rather than left as a bare error. `withFanoutRefusalGuidance` detects the shared predicate's marker and rethrows with the real recovery path — convert to draft and run the gate with `--execution-mode fanout_fanin` and a findings-log ledger — with the original reason preserved verbatim. The existing draft-state rollback still runs, so a refused reconcile does not strand the PR in draft.

The recovery for a large PR is now "run the real gate", which is what it always should have been. `scripts/README.md` and the tool's own `USAGE` state the limitation.

## Test-fixture change

Pre-existing tests that were incidentally posting inline verdicts (now correctly refused) are isolated by two mechanisms. In `upsert-checkpoint-verdict.test.mjs`, tests that relied on this repo's real `.devloops` through ambient `process.cwd()` use a `fanoutDisabledRepoRoot` fixture: a byte-identical copy of the real `.devloops` with only `requireFanoutEvidence` flipped, guarded by an assertion that the flip actually changed the value. In `reconcile-draft-gate.test.mjs` (whose tests already ran in a `tempDir` cwd) and in the subprocess tests, a minimal `disableFanoutEvidence(dir)` helper writes the single flag instead.

This is provably behavior-preserving. The old code never read `requireFanoutEvidence` at all, so no pre-existing assertion could have depended on its value. Verified by diffing those files against the pre-issue base for removed fan-out or light-mode assertions: zero matches. The test that pins "draft_gate blocks on high only" still reads the real `blockCleanOnFindingSeverities` and angle config unchanged.

Also found while doing this: the shipped `extension-defaults.yaml` enables `localImplementation.lightMode` by default, so a bare tempDir is not "no lightMode". A misleading comment was corrected, and `enforcePostTimeFanoutMode` gained a `resolveLightMode(config) != null` guard so its light-facts fetch matches merge-time's own lazy-fetch condition instead of always paying the call.

## Scope and context

- `scripts/github/detect-checkpoint-evidence.mjs` — predicate extraction.
- `scripts/github/upsert-checkpoint-verdict.mjs` — post-time enforcement.
- `scripts/github/reconcile-draft-gate.mjs` — actionable refusal, usage text.
- `scripts/README.md` — the recovery limitation.
- `skills/docs/gate-review-sub-loop-contract.md` (+ generated mirror) — enforcement at both boundaries.
- `test/github/upsert-checkpoint-verdict.test.mjs`, `test/github/reconcile-draft-gate.test.mjs`.
- `CHANGELOG.md` — entry under Unreleased.

## Acceptance criteria

- [x] Post-time enforcement calls the exported `buildFanoutEnforcement` predicate shared with merge time; no duplicated threshold/label/scope logic.
- [x] An over-threshold inline post is refused when `requireFanoutEvidence: true`.
- [x] Refusal applies to `clean`, `findings_present`, and `blocked` alike.
- [x] The refusal error names the failure reason, mirroring the merge-time wording.
- [x] An under-threshold light-mode inline post is still accepted.
- [x] `fanout_fanin` posts are unaffected, including the draft-transition re-post path.
- [x] Scope that cannot be derived fails closed, as at merge time.
- [x] No override flag.

## Definition of done

- [x] Tests: refusal for each verdict value; light-mode acceptance; fail-closed on un-derivable scope; `fanout_fanin` and opt-out pass-through; a shared fixture proving both boundaries reach identical decisions.
- [x] A test pins that a non-light-mode reconcile is refused AND carries the actionable guidance.
- [x] Non-vacuity proven by mutation: forcing the mode check to accept breaks 3 new and 6 existing merge-time tests; neutering the post-time call site breaks 4; neutering the guidance wrapper reverts the error to the bare message.
- [x] `npm run verify` green, `npm run assets:check` green.

## Non-goals

- Changing the light-mode thresholds or the `gate:full` label semantics.
- Adding an override for post-time refusal.
- Changing merge-time behavior (the extraction is behavior-preserving).


scripts/github/reconcile-draft-gate.mjs
```diff
@@ -19,10 +19,18 @@ Optional/manual recovery tool for an already non-draft PR when you want to
 retroactively record clean \`draft_gate\` evidence.
 Converts the PR to draft, validates the head, posts a reconciling clean
 draft_gate comment, then marks the PR ready for review again.
+This posts an INLINE verdict — it can only reconcile a PR that qualifies for
+the light-mode carve-out (gates.localImplementation.lightMode, under
```
mfittko: **low** (`contract-surface`): The new USAGE text names the light-mode config key as gates.localImplementation.lightMode, but the setting is top-level localImplementation.lightMode — resolveLightMode reads config?.localImplementation?.lightMode (packages/core/src/config/config.mjs:1913), no gates.localImplementation alias exists in the schema, and the PR's own test fixtures write localImplementation: at the top level. An operator following this refusal guidance to enable the carve-out would set a key that resolves to nothing. The neighbouring gates.requireFanoutEvidence on line 24 is correctly scoped, which makes the wrong prefix read as deliberate.

scripts/README.md
```diff
@@ -83,7 +83,7 @@ Contract:
 - rejects `--ready` before invoking `gh`; use `gh pr ready` later after draft-gate approval
 - forwards every other argument to `gh pr create` unchanged and in order
 - preserves the underlying `gh pr create` stdout, stderr, and exit code without wrapping success output
-- stays intentionally narrow: this is the prevention layer for draft-first creation, while `scripts/github/reconcile-draft-gate.mjs` remains the separate recovery path for already-open non-draft PRs
+- stays intentionally narrow: this is the prevention layer for draft-first creation, while `scripts/github/reconcile-draft-gate.mjs` remains the separate recovery path for already-open non-draft PRs — that recovery only completes for a PR under the light-mode threshold (it posts an inline verdict); a larger PR must convert to draft and run the real fan-out gate instead
```
mfittko, line 86: **low** (`contract-surface`): The added clause states the reconcile limitation unconditionally — "that recovery only completes for a PR under the light-mode threshold (it posts an inline verdict)". The refusal only fires while gates.requireFanoutEvidence is enabled; in an opted-out repo (requireFanoutEvidence: false, the documented and only opt-out) reconcile still completes for a PR of any size. The tool's own USAGE text carries that qualifier ("When gates.requireFanoutEvidence is enabled and the PR is over that threshold"), so this summary is narrower than the behavior it summarizes and tells opted-out consumers their working recovery path is unavailable.

scripts/github/reconcile-draft-gate.mjs
```diff
@@ -168,6 +176,26 @@ async function checkCiStatus({ repo, pr, headSha }, { env, ghCommand, runChild =
...
+// refusal is correct, not a bug in this tool, but a bare "inline gate
+// verdicts are not accepted" leaves the operator without a recovery path.
+const FANOUT_REFUSAL_MARKER = "inline gate verdicts are not accepted";
+function isFanoutRefusalError(error) {
+  return error instanceof Error && error.message.includes(FANOUT_REFUSAL_MARKER);
+}
+function withFanoutRefusalGuidance(error) {
+  if (!isFanoutRefusalError(error)) return error;
+  return new Error(
+    `${error.message} reconcile-draft-gate only completes for a PR under the light-mode threshold ` +
```
mfittko, line 193: **medium** (`correctness`): The new guidance text, the new USAGE block (lines 22-27) and scripts/README.md:86 all claim reconcile-draft-gate still completes for a PR under the light-mode threshold. It completes for NO PR while requireFanoutEvidence is on: reconcileDraftGate calls upsertCheckpointVerdict (line 238) without executionMode and without inlineReason, so the candidate marker carries inlineReason: null; evaluateInlineFanoutMode's lightAccepted requires typeof gate.inlineReason === "string" && trim().length > 0, which null can never satisfy. Even a 1-file/2-line PR under the shipped default (lightMode enabled 2 files/100 lines, requireFanoutEvidence default true) is refused, and the refusal tells the operator to run a full fan-out gate for a micro-PR. No test covers an under-threshold reconcile succeeding; the one new reconcile test explicitly disables lightMode, so the documented path is untested and unreachable.

scripts/github/upsert-checkpoint-verdict.mjs
```diff
@@ -1491,6 +1554,22 @@ export async function upsertCheckpointVerdict(options, { env = process.env, ghCo
   if (gateActionForbidden) {
     throw new Error(buildGateEntryRefusalError({ options, coordination }));
   }
+  // Post-time fan-out evidence enforcement: refuse an under-qualified inline
+  // verdict here, before it is posted, for every verdict value. No override
+  // flag — requireFanoutEvidence: false is the only opt-out, and that is
+  // already handled inside buildFanoutEnforcement.
+  await enforcePostTimeFanoutMode(
```
mfittko, line 1561: **medium** (`correctness`): enforcePostTimeFanoutMode is placed after the #891 draft-transition branch (line 1525), so an under-qualified inline draft_gate post on a ready PR first converts the PR to draft, then re-enters, then refuses. The refusal is deterministic from options+config alone and needs none of the PR-state facts that branch depends on, yet it opens the exact convert->crash->stuck-in-draft window documented at line 1142 for a post that can never succeed. This is now the main new refusal path (reconcile of a non-light PR), so it is hit routinely, not rarely.

scripts/github/upsert-checkpoint-verdict.mjs
```diff
@@ -1491,6 +1554,22 @@ export async function upsertCheckpointVerdict(options, { env = process.env, ghCo
...
+    {
+      repo: options.repo,
+      pr: options.pr,
+      gate: options.gate,
+      executionMode: options.executionMode ?? DEFAULT_EXECUTION_MODE,
+      inlineReason: options.inlineReason,
+      headSha: canonicalHeadSha,
+      config,
```
mfittko, line 1569: **low** (`correctness`): The config fed to post-time enforcement ignores config-load errors: line 1431 destructures only { config } from loadDevLoopConfig, while merge time nulls the config when errors is non-empty (detect-checkpoint-evidence.mjs:773-774). applyLayer drops an invalid layer and records an error, so a repo whose gates.requireFanoutEvidence: false opt-out lives in a layer that failed schema validation resolves to the default true at post time (refused) while merge time treats the config as unavailable and does not enforce at all. That is a live divergence in the one seam the PR claims cannot drift.

scripts/github/upsert-checkpoint-verdict.mjs
```diff
@@ -1249,6 +1250,68 @@ export function buildCoordinationEvaluatorInput({
...
+// otherwise (fanout_fanin candidates, light-mode-accepted inline candidates,
+// and any candidate while requireFanoutEvidence is off).
+async function enforcePostTimeFanoutMode({ repo, pr, gate, executionMode, inlineReason, headSha, config }, { env, ghCommand, repoRoot, runChild }) {
+  let hasFullLabel = false;
+  let baseRef = null;
+  // Light-mode facts only matter for an inline candidate under active
+  // enforcement; fetched lazily so a fanout_fanin post (the common path) pays
+  // no extra gh call, and so does an inline post once light mode itself is
+  // off (no facts could ever change the outcome).
+  if (executionMode === "inline_single_agent" && resolveRequireFanoutEvidence(config) && resolveLightMode(config) != null) {
```
mfittko, line 1271: **low** (`correctness`): Only the predicate is shared; the light-facts acquisition is copy-pasted. The lazy-fetch guard, the gh pr view --json baseRefOid,labels call, the baseRefOid trim/empty check and the labels/gate:full parse at lines 1271-1287 duplicate detect-checkpoint-evidence.mjs:785-799 verbatim. They agree today (post-time's missing config != null guard is masked because resolveLightMode(null) returns null), but the fact-derivation layer is exactly the drift surface the change exists to close, and the two copies already differ in one respect (merge time passes a restFallback to runGhJson, post time does not).

scripts/github/upsert-checkpoint-verdict.mjs
```diff
@@ -1249,6 +1250,68 @@ export function buildCoordinationEvaluatorInput({
...
+    config,
+    cwd: repoRoot,
+    hasFullLabel,
+    baseRef,
+  });
+  if (!fanoutEnforcement.required) return;
+  for (const gateResult of fanoutEnforcement.gates) {
+    const modeFailure = evaluateInlineFanoutMode(gateResult, fanoutEnforcement);
+    if (modeFailure) {
+      throw new Error(`Cannot post a ${executionMode} verdict for ${repo}#${pr} ${gate}: ${modeFailure}`);
```
mfittko, line 1310: **low** (`correctness`): The refusal message cannot distinguish its four causes. inline gate verdicts are not accepted is emitted identically for over-threshold scope, a gate:full label, lightMode disabled, and un-derivable scope. Un-derivable scope is a new, realistic post-time failure: detectMergeBaseScope shells out to git diff &lt;baseRefOid>...&lt;head> in repoRoot, and a local checkout that has not fetched the base tip does not have that commit object, so a genuinely under-threshold micro-PR is refused with a message that says the opposite (that inline is not accepted for this PR).

scripts/github/reconcile-draft-gate.mjs
mfittko: **high** (`coverage`): The light-mode recovery path this PR documents has ZERO coverage and is in fact unreachable: no reconcile can ever succeed under requireFanoutEvidence: true. reconcileDraftGate calls upsertCheckpointVerdict programmatically (:238) passing neither executionMode nor inlineReason; the inline-reason default and non-empty validation live only in the CLI parser (upsert-checkpoint-verdict.mjs:576), so the candidate marker enforcement evaluates is literally {"visible":true,"executionMode":"inline_single_agent","inlineReason":null,...} (captured end-to-end via an instrumented load-hook run of the reconcile suite). evaluateInlineFanoutMode's typeof gate.inlineReason === "string" && trim().length > 0 clause can therefore never hold, so lightAccepted is always false. Verified against the real predicate with identical light facts (lightMode on, scope 0 files/0 lines, no gate:full): the CLI marker shape is ACCEPTED, the reconcile marker shape is REFUSED, differing only in inlineReason. Consequently a genuine 1-file/2-line PR is refused too, and the new USAGE text here plus scripts/README.md:86 ("that recovery only completes for a PR under the light-mode threshold") assert an outcome no input can produce. This is exactly the gap a test would have caught: every reconcile test either sets requireFanoutEvidence: false (disableFanoutEvidence / the defaults.yaml requireCi test) or, in the one new test, disables lightMode - so not a single test exercises a successful light-mode inline post through reconcile-draft-gate under active enforcement.

scripts/github/upsert-checkpoint-verdict.mjs
```diff
@@ -1249,6 +1250,68 @@ export function buildCoordinationEvaluatorInput({
...
+  if (executionMode === "inline_single_agent" && resolveRequireFanoutEvidence(config) && resolveLightMode(config) != null) {
+    try {
+      const prFacts = await runGhJson(
+        ["pr", "view", String(pr), "--repo", repo, "--json", "baseRefOid,labels"],
+        { env, ghCommand, runChild },
+      );
+      baseRef = typeof prFacts?.baseRefOid === "string" && prFacts.baseRefOid.trim().length > 0
+        ? prFacts.baseRefOid.trim()
+        : null;
+      hasFullLabel = Array.isArray(prFacts?.labels)
```
mfittko, line 1280: **medium** (`coverage`): Mutation survivor: replacing the whole post-time gate:full label read with hasFullLabel = false breaks ZERO of the 240 tests. The label is a first-class term of the acceptance predicate (the contract text this PR edits says the label "always forces the full fan-out - scope is not even measured"), and the post-time side is the only place that resolves it from the PR. The shared predicate's own label clause IS covered, but only at merge time (mutating fanoutEnforcement.hasFullLabel !== true to true kills exactly one merge-time test). So a regression that drops the label from the post-time gh fetch, or mis-spells GATE_FULL_LABEL, ships silently and lets a gate:full-labelled under-threshold PR post an unearned inline verdict as visible PR evidence - the precise harm class #1599 exists to prevent - caught only later at merge.

scripts/github/upsert-checkpoint-verdict.mjs
```diff
@@ -1249,6 +1250,68 @@ export function buildCoordinationEvaluatorInput({
...
+  // .headSha off it). The other gate is marked invisible so only the gate
+  // actually being posted is evaluated this call.
+  const candidateMarker = { visible: true, executionMode, inlineReason: inlineReason ?? null, headSha };
+  const inactiveMarker = { visible: false };
+  const fanoutEnforcement = await buildFanoutEnforcement({
+    repo,
+    pr,
+    currentHeadSha: headSha,
+    draftGateMarker: gate === "draft_gate" ? candidateMarker : inactiveMarker,
+    preApprovalGateMarker: gate === "pre_approval_gate" ? candidateMarker : inactiveMarker,
```
mfittko, line 1300: **medium** (`coverage`): Mutation survivor: replacing preApprovalGateMarker: gate === "pre_approval_gate" ? candidateMarker : inactiveMarker with a bare inactiveMarker - i.e. removing post-time enforcement for pre_approval_gate entirely - breaks ZERO tests. Every new post-time test posts --gate draft_gate. The contract text this PR adds says the produce step refuses "for a required gate" without qualification, and pre_approval_gate is the gate that actually gates merge, so a regression that silently exempts it would defeat the feature on its most important surface with no test signal.

test/github/upsert-checkpoint-verdict.test.mjs
```diff
@@ -4773,6 +4806,247 @@ test("upsert-checkpoint-verdict records executionMode and warns on inline, stays
...
+// carve-out is refused BEFORE it is ever posted, regardless of verdict value.
+
+test("upsert-checkpoint-verdict refuses to post an inline verdict for a required gate, for every verdict value", async () => {
+  const tempDir = await mkdtemp(path.join(os.tmpdir(), "dev-loops-upsert-postgate-refuse-"));
+  try {
+    // requireFanoutEvidence: true; no explicit base-ref/label mock entry is
+    // staged below, so the light-facts fetch (the shipped default enables
+    // lightMode) gets an unusable response and scope stays un-derivable —
+    // fails closed exactly like an over-threshold PR would.
+    await writeFile(path.join(tempDir, ".devloops"), "version: 1\ngates:\n  requireFanoutEvidence: true\n", "utf8");
```
mfittko, line 4823: **medium** (`coverage`): The headline acceptance criterion ("An over-threshold inline post is refused", the PR-1598 scenario) is never actually tested against a threshold. This test's own comment concedes it: it stages no baseRef/label entry so "scope stays un-derivable - fails closed exactly like an over-threshold PR would", which makes it a second copy of the dedicated scope-underivable test rather than an over-threshold test. Confirmed by mutation: replacing scopeUnderThreshold = scope.ok === true && isEligibleForLightMode(scope, lightThreshold) (detect-checkpoint-evidence.mjs:674) with scopeUnderThreshold = true breaks ZERO of 240 tests, so the maxFiles/maxLines comparison is unexercised in the rejecting direction at BOTH boundaries - merge time included, since they share the predicate. The only test that drives a real git repo through detectMergeBaseScope (detect-checkpoint-evidence.test.mjs:1983) asserts the UNDER-threshold true case and a gate:full case that never enters the guard; the merge-time "REJECTS an over-threshold inline verdict" test hand-sets scopeUnderThreshold: false on a synthetic gate. Net: the DoD's "non-vacuity proven by mutation" claim does not hold for the central criterion - a broken threshold comparison would let over-threshold inline verdicts through at post AND merge time undetected.

scripts/github/upsert-checkpoint-verdict.mjs
```diff
@@ -1249,6 +1250,68 @@ export function buildCoordinationEvaluatorInput({
...
+  // enforcement; fetched lazily so a fanout_fanin post (the common path) pays
+  // no extra gh call, and so does an inline post once light mode itself is
+  // off (no facts could ever change the outcome).
+  if (executionMode === "inline_single_agent" && resolveRequireFanoutEvidence(config) && resolveLightMode(config) != null) {
+    try {
+      const prFacts = await runGhJson(
+        ["pr", "view", String(pr), "--repo", repo, "--json", "baseRefOid,labels"],
+        { env, ghCommand, runChild },
+      );
+      baseRef = typeof prFacts?.baseRefOid === "string" && prFacts.baseRefOid.trim().length > 0
```
mfittko, line 1277: **low** (`coverage`): Mutation survivor: changing the baseRef fallback from : null to a bogus : "HEAD" when a SUCCESSFUL gh response carries a missing/empty baseRefOid breaks ZERO tests. The fail-closed test covers only the fetch-THROWS path (gh exit 1); the fetch-succeeds-but-payload-is-unusable path has no coverage, so this non-empty-string guard is untested. In the mutated world detectMergeBaseScope would measure HEAD...head as an empty diff and accept the inline verdict, which is why the guard matters.

scripts/github/upsert-checkpoint-verdict.mjs
```diff
@@ -1249,6 +1250,68 @@ export function buildCoordinationEvaluatorInput({
...
+// check; this refuses an inline verdict at the PRODUCE step instead, before it
+// is ever posted as visible PR evidence, by reusing the exact merge-time
+// acceptance predicate (buildFanoutEnforcement + evaluateInlineFanoutMode) —
+// never mirroring the threshold/label/scope logic — so the two boundaries
+// cannot drift apart. Applies to every verdict value (clean/findings_present/
+// blocked): mode qualification does not depend on the conclusion being
+// recorded. Throws when the candidate does not qualify; resolves silently
+// otherwise (fanout_fanin candidates, light-mode-accepted inline candidates,
+// and any candidate while requireFanoutEvidence is off).
+async function enforcePostTimeFanoutMode({ repo, pr, gate, executionMode, inlineReason, headSha, config }, { env, ghCommand, repoRoot, runChild }) {
```
mfittko, line 1264: **nit** (`coverage`): The DoD's mutation counts are slight undercounts (they appear to count only the upsert test file). Measured via ESM-load-hook mutation on a pristine git archive tree: forcing evaluateInlineFanoutMode to accept kills 10 tests (6 merge-time + 3 new upsert + 1 new reconcile), not "3 new and 6 existing"; neutering the post-time call site kills 5 (4 upsert + 1 reconcile), not 4; neutering the guidance wrapper kills exactly 1, as claimed. All three claims hold directionally - the checks are non-vacuous - the numbers are just floors that omit the reconcile test.

scripts/github/upsert-checkpoint-verdict.mjs
```diff
@@ -1249,6 +1250,68 @@ export function buildCoordinationEvaluatorInput({
...
+      const prFacts = await runGhJson(
+        ["pr", "view", String(pr), "--repo", repo, "--json", "baseRefOid,labels"],
+        { env, ghCommand, runChild },
+      );
+      baseRef = typeof prFacts?.baseRefOid === "string" && prFacts.baseRefOid.trim().length > 0
+        ? prFacts.baseRefOid.trim()
+        : null;
+      hasFullLabel = Array.isArray(prFacts?.labels)
+        && prFacts.labels.some((label) => (typeof label === "string" ? label : label?.name) === GATE_FULL_LABEL);
+    } catch {
```
mfittko, line 1282: **medium** (`determinism`): The light-facts fetch failure is swallowed with a bare catch {}, so a transient gh failure (rate limit, 5xx, auth blip) silently flips an otherwise-accepted light-mode inline post into a refusal whose message names the execution mode rather than the fetch failure. Same PR, same config, same head, different outcome depending on transient network state, with the cause erased from the operator-visible error. The existing test for this path ("fails closed on an inline verdict when the base ref cannot be resolved") stages exactly this case (gh exit 1, "rate limited") and asserts only the generic policy wording, so the misattribution is pinned rather than caught.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko, line 1277: **medium** (`determinism`): Post-time scope re-derivation depends on machine-local git fetch state, not on the PR. baseRefOid is GitHub's CURRENT base-branch tip, but detectMergeBaseScope (scripts/loop/detect-change-scope.mjs:114) runs git diff &lt;base>...&lt;head> against the LOCAL object store with no fetch or existence probe. When the base branch advanced remotely since the last local fetch (routine here: parallel runners, frequent merges to main), the OID is absent locally, git fails, scopeUnderThreshold stays false, and a genuinely under-threshold micro-PR is refused. Verified in a temp git repo: identical 1-file/1-line change, lightMode on, no gate:full label — base OID present gives scopeUnderThreshold: true and failure: null; base OID absent gives scopeUnderThreshold: false and the full refusal, with git leaking a stray fatal: Invalid symmetric difference expression to stderr. Merge time carries the same exposure, but this PR newly puts it on the PRODUCE step, where it blocks recording the verdict at all.

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `4a331a8d487e75af924d75e8884a2180f3080da3`
**Verdict:** findings_present

**Findings summary:** 9 angles reviewed; 16 findings (see per-angle breakdown below).

- `correctness` → `findings_present` (5 findings)
- `coverage` → `findings_present` (6 findings)
- `gate-evidence` → `clean`
- `link-check` → `clean`
- `scope` → `clean`

**Next action:** Stay draft; fix pass for the high and confirmed mediums, then re-gate.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko, line 1271: **low** (`determinism`): The shared predicate covers only the DECISION; each boundary still hand-writes its own copy of the fact-gathering that feeds it, and the two have already drifted. Merge time (detect-checkpoint-evidence.mjs:785-799) issues the identical gh pr view --json baseRefOid,labels call with restFallback: () => restFetchPrView(...) for the gh-less REST mode; post time's local runGhJson (line 1091) has no restFallback parameter at all, and upsert-checkpoint-verdict.mjs does not import _gh-rest-fallback.mjs. Same class of drift the extraction exists to eliminate, one layer down. Impact today is bounded because upsert cannot post without gh at all, but the duplicated block is where the next divergence lands (see the untested label/base-ref plumbing on the coverage angle).

scripts/github/upsert-checkpoint-verdict.mjs
mfittko, line 1561: **low** (`determinism`): Enforcement is placed AFTER the #891 draft-transition branch (lines 1525-1532), which returns early into postDraftGateViaDraftTransition. An under-qualified inline draft_gate post on a ready PR therefore performs a real convertPullRequestToDraft, re-enters, refuses, and rolls back with pr ready — two live PR-state mutations plus their notifications on a path whose outcome was decided before any of them. The outcome is deterministic; the state churn is avoidable.

scripts/github/upsert-checkpoint-verdict.mjs
```diff
@@ -1249,6 +1250,68 @@ export function buildCoordinationEvaluatorInput({
...
+    }
+  }
+  // A candidate marker describing the verdict ABOUT TO BE POSTED — no comment
+  // exists yet, so this is shaped exactly like a posted marker
+  // (buildFanoutEnforcement only reads .visible/.executionMode/.inlineReason/
+  // .headSha off it). The other gate is marked invisible so only the gate
+  // actually being posted is evaluated this call.
+  const candidateMarker = { visible: true, executionMode, inlineReason: inlineReason ?? null, headSha };
+  const inactiveMarker = { visible: false };
+  const fanoutEnforcement = await buildFanoutEnforcement({
```
mfittko, line 1295: **low** (`determinism`): For a fanout_fanin candidate — the common path — enforcePostTimeFanoutMode still calls the full buildFanoutEnforcement, which spawns git worktree list via resolveLedgerCheckouts and then runs readLedgerProvenanceInAny + ledgerExistsInAny across every resolved checkout. None of that is read: post time consults only evaluateInlineFanoutMode, which returns null immediately on executionMode === "fanout_fanin". Every gate post now pays cross-worktree filesystem and git I/O for a discarded result, and couples the produce step to other checkouts' ledger state for no reason.

scripts/github/upsert-checkpoint-verdict.mjs
```diff
@@ -1249,6 +1250,68 @@ export function buildCoordinationEvaluatorInput({
...
+    repo,
+    pr,
+    currentHeadSha: headSha,
+    draftGateMarker: gate === "draft_gate" ? candidateMarker : inactiveMarker,
```
mfittko, line 1299: **low** (`input-validation`): An unrecognized gate name fails OPEN. The candidate marker is attached only when gate === "draft_gate" or gate === "pre_approval_gate"; any other value leaves both markers { visible: false }, so buildFanoutEnforcement's gateSpecs filter yields an empty gates array, the loop body never runs, and enforcePostTimeFanoutMode resolves silently — the inline verdict is then posted with no mode qualification at all. The CLI is safe (normalizeGateName rejects unknown values at line 459), but upsertCheckpointVerdict is an exported programmatic API (reconcile-draft-gate calls it directly), and resolveGateAction treats every non-draft_gate value as the pre-approval action, so a typo'd programmatic gate can pass the entry check and post unenforced.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko, line 1280: **low** (`input-validation`): The labels payload shape is not validated fail-closed. hasFullLabel = Array.isArray(prFacts?.labels) && ... maps a missing, null or non-array labels field to false, i.e. "this PR carries no gate:full label" — the permissive answer. A response that returns baseRefOid but no usable labels array therefore lets the light-mode carve-out apply to a PR that actually carries gate:full, defeating the "the label always forces the full fan-out, scope is not even measured" invariant the contract doc restates in this same PR. The baseRefOid half of the same block is correctly fail-closed (null base means scope is never re-derived); only the label half is not. The identical parse exists at merge time (detect-checkpoint-evidence.mjs:794), so both boundaries share the weakness and there is no post/merge divergence — the fix belongs in whichever shared helper the duplication is collapsed into.

test/github/reconcile-draft-gate.test.mjs
```diff
@@ -44,6 +44,15 @@ async function writeGhStub(tempDir, entries) {
   return { ...env, DEVLOOPS_RUN_ID: "" };
 }
 
+// These tests are about the reconcile tool's draft-transition/CI/error
+// mechanics, not fan-out evidence — the reconciling post is inline by
+// design — and use a bare tempDir as repoRoot (schema default:
+// requireFanoutEvidence: true). Disable it so the reconciling post is not
+// itself refused as under-qualified inline evidence.
+async function disableFanoutEvidence(tempDir) {
```
mfittko, line 52: **low** (`pr-description`): The body's "Test-fixture change" section attributes one mechanism to both test files, but two different mechanisms ship. It says ~35 tests "in upsert-checkpoint-verdict.test.mjs and reconcile-draft-gate.test.mjs relied on this repo's real .devloops through ambient process.cwd()" and "use a fanoutDisabledRepoRoot fixture ... byte-identical copy of the real .devloops". reconcile-draft-gate.test.mjs never used ambient process.cwd() (every test already passed cwd: tempDir) and never uses fanoutDisabledRepoRoot; it uses a new disableFanoutEvidence(tempDir) helper that writes a minimal 'version: 1 / gates: requireFanoutEvidence: false' file. The same minimal-file mechanism is also used inside upsert-checkpoint-verdict.test.mjs (writeSingleSurfaceLedger and the two subprocess-cwd tests), and that minimal file is not byte-identical to the real .devloops. The behavior-preserving claim itself holds: diffing both test files against merge base 29665f07 shows 7 removed lines, all repoRoot plumbing, and zero removed or weakened assertions.

scripts/github/reconcile-draft-gate.mjs
mfittko: Fixed: reconcile-draft-gate now passes executionMode=inline_single_agent and an inlineReason to upsertCheckpointVerdict, so evaluateInlineFanoutMode's light-mode acceptance clause (non-empty inlineReason) can actually hold for an under-threshold PR. The previously-unreachable recovery path is now reachable; over-threshold PRs still refuse (scopeUnderThreshold stays false).

Covered by a new real-git-repo test (reconcile-draft-gate succeeds for a light-mode under-threshold PR under active requireFanoutEvidence) that drives detectMergeBaseScope against a genuine under-threshold diff and asserts the inline verdict is accepted at post time.

Fixing commits: 46b35dc2 (fix) and 94aefbbf (test count tolerance for CI's non-deterministic post-update verification path).

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Addressed: added a post-time test (upsert-checkpoint-verdict refuses an inline verdict at post time when the PR carries the gate:full label) that stages the light-facts fetch returning a gate:full label and asserts the inline post is refused even when the merge-base scope would otherwise be under threshold. Kills the `hasFullLabel = false` mutation on the post-time fetch.

Fixing commit: 5d706ee5.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Addressed: added a post-time test (upsert-checkpoint-verdict refuses an inline pre_approval_gate verdict at post time under requireFanoutEvidence) that stages a Copilot review so coordination allows pre_approval_gate entry, disables lightMode so the lazy fetch does not fire, and asserts the inline pre_approval_gate post is refused — proving the pre_approval_gate candidate marker is evaluated, not silently skipped. Kills the mutation that replaces the pre_approval_gate candidate marker with a bare inactiveMarker.

Fixing commit: 5d706ee5.

scripts/github/reconcile-draft-gate.mjs
mfittko: Addressed: this is the same root cause as the HIGH coverage finding (comment 3743076341). reconcile-draft-gate now passes executionMode=inline_single_agent and an inlineReason to upsertCheckpointVerdict, so the candidate marker carries a non-empty inlineReason and evaluateInlineFanoutMode's light-mode acceptance clause can hold for an under-threshold PR. The reconcile recovery path is now reachable; a 1-file/2-line micro-PR is no longer refused.

Covered by the new real-git-repo test. Fixing commits: 46b35dc2, 94aefbbf.

test/github/upsert-checkpoint-verdict.test.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/github/reconcile-draft-gate.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/README.md
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

test/github/reconcile-draft-gate.test.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred to follow-up: non-blocking finding (low/medium severity below the gate's high-only blockCleanOnFindingSeverities, or a determinism/coverage improvement that does not affect the shipped correctness of post-time enforcement). The post-time enforcement reuses the exact merge-time acceptance predicate (evaluateInlineFanoutMode via buildFanoutEnforcement), so the two boundaries cannot drift on what counts as an accepted inline verdict. Tracked for a future hardening pass; not gating this PR.

mfittko (commented):
### Gate review: `draft_gate`

**Reviewed head SHA:** `5d706ee58019203f1c0c7422cbfa94ebaf3b10b1`
**Verdict:** findings_present

**Findings summary:** 9 angles reviewed; 8 findings (see per-angle breakdown below).

- `contract-surface` → `findings_present`
  - [`medium`] reconcile-draft-gate USAGE cites the wrong config key path gates.localImplementation.ligh (`scripts/github/reconcile-draft-gate.mjs:23`) — _`accepted-for-fix`_
- `correctness` → `findings_present`
  - [`low`] Config-nulling divergence between the two enforcement boundaries. Post-time loads config v (`scripts/github/upsert-checkpoint-verdict.mjs:1431`) — _`deferred`_
- `coverage` → `findings_present`
  - [`low`] Mutation survivor: the negative branch of withFanoutRefusalGuidance's guard (if (!isFanou (`scripts/github/reconcile-draft-gate.mjs:191`) — _`accepted-for-fix`_
- `determinism` → `findings_present`
  - [`low`] Machine-local git object store gates light-mode acceptance at post time, producing same-in (`scripts/github/upsert-checkpoint-verdict.mjs:1302`) — _`deferred`_
  - [`low`] Bare catch swallows ALL gh failures on the light-facts fetch and misattributes a transient (`scripts/github/upsert-checkpoint-verdict.mjs:1282`) — _`deferred`_
  - [`low`] State churn on a deterministic-refusal path: reconcileDraftGate calls convertPrToDraft (li (`scripts/github/reconcile-draft-gate.mjs:235`) — _`deferred`_
- `gate-evidence` → `clean`
- `input-validation` → `findings_present`
  - [`low`] enforcePostTimeFanoutMode does not validate the gate parameter: when gate is neither "dr (`scripts/github/upsert-checkpoint-verdict.mjs:1299`) — _`deferred`_
- `link-check` → `clean`
- `pr-description` → `findings_present`
  - [`low`] PR body 'Test-fixture change' section claims the subprocess tests use a named disableFano (`test/github/upsert-checkpoint-verdict.test.mjs:2960`) — _`deferred`_
- `scope` → `clean`

**Next action:** Stay draft; fix 1 medium config-key doc + 7 low non-blocking findings or accept, re-gate clean, then mark ready.

scripts/github/reconcile-draft-gate.mjs
mfittko: Fixed: the USAGE text now cites the canonical config key `localImplementation.lightMode` (was `gates.localImplementation.lightMode`, which resolves to nothing — no `gates.localImplementation` alias exists in the schema; `resolveLightMode` reads `config?.localImplementation?.lightMode`). Comment-only doc fix on the head that re-gates this round; the previously-deferred note is superseded by this actual fix.

Fixing commit: f4772dbb

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `f4772dbbc49529e3fa7530f9112effd4f5fc0f48`
**Verdict:** clean

**Findings summary:** 9 angles reviewed; no findings (see per-angle breakdown below).

- `correctness` → `clean`
- `coverage` → `clean`
- `determinism` → `clean`
- `gate-evidence` → `clean`
- `link-check` → `clean`
- `pr-description` → `clean`
- `scope` → `clean`

**Next action:** Mark PR ready for review

scripts/github/reconcile-draft-gate.mjs
```diff
@@ -168,6 +176,26 @@ async function checkCiStatus({ repo, pr, headSha }, { env, ghCommand, runChild =
...
+    `(it can only synthesize an inline verdict); this PR needs an actual fan-out gate run — convert it ` +
+    `to draft and post draft_gate with --execution-mode fanout_fanin backed by a findings-log ledger for ` +
+    `the current head, instead of reconciling.`
+  );
```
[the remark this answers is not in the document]
mfittko: Deferred to follow-up: non-blocking finding. The `withFanoutRefusalGuidance` separator is a cosmetic error-message formatting improvement (a run-on error string) that does not affect the correctness of post-time `requireFanoutEvidence` enforcement — the refusal still fires, the guidance still reaches the operator, and the predicate is shared with merge time. Tracked for a future hardening pass alongside the other deferred guidance/formatting findings; not gating this PR.

scripts/github/reconcile-draft-gate.mjs
```diff
@@ -168,6 +176,26 @@ async function checkCiStatus({ repo, pr, headSha }, { env, ghCommand, runChild =
...
   };
 }
+
+// The shared post-time fan-out enforcement predicate (evaluateInlineFanoutMode,
+// detect-checkpoint-evidence.mjs) always includes this phrase in its rejection
+// message. This tool only ever posts an inline verdict, so once a PR is over
+// the light-mode threshold it can never synthesize qualifying evidence — that
+// refusal is correct, not a bug in this tool, but a bare "inline gate
+// verdicts are not accepted" leaves the operator without a recovery path.
+const FANOUT_REFUSAL_MARKER = "inline gate verdicts are not accepted";
```
mfittko, line 186: **low** (`dry`): scripts/github/reconcile-draft-gate.mjs:186 — FANOUT_REFUSAL_MARKER hardcodes the substring "inline gate verdicts are not accepted", duplicating a fragment of the rejection message that evaluateInlineFanoutMode builds at scripts/github/detect-checkpoint-evidence.mjs:323. The two are not linked: if the predicate's message wording ever changes, reconcile-draft-gate's guidance detection silently regresses to the bare error (withFanoutRefusalGuidance no-ops) with no compile-time or test signal. The extraction of the decision predicate itself is byte-identical and genuinely deduplicates the merge/post-time decision; this is the one residual cross-module copy that survives it.

mfittko (commented):


**Reviewed head SHA:** `f4772dbbc49529e3fa7530f9112effd4f5fc0f48`
**Verdict:** clean

**Findings summary:** 7 angles reviewed; 2 findings (see per-angle breakdown below).

- `docs` → `clean`
- `yagni` → `clean`

> **medium** (`pr-checklist-matrix`): PR body has no AC/DoD/non-goals matrix as a markdown table. The 'Acceptance criteria', 'Definition of done', and 'Non-goals' sections exist as separate checkbox/bullet lists, but no table cross-maps each acceptance criterion to its corresponding definition-of-done item(s). The angle requires a markdown table (header row + >=1 content row) mapping AC -> DoD. All AC/DoD checkboxes are checked (&#91;x]) and non-goals are explicit, so the substance is complete; only the required matrix table format is missing.

**Next action:** Await final human approval; merge-ready pending explicit merge authorization

scripts/github/reconcile-draft-gate.mjs
mfittko: Deferred at gate close (round 1, fingerprint e50430cce91f8806, severity low, angle dry): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer; #1585); the deferral is recorded on this thread's marker and in the durable findings ledger.

Issue #1012, opened 2026-06-28
Bug: reply-resolve-review-threads.mjs (plural) hangs indefinitely -> accumulates stale background tasks; add bounded timeout + fail-closed


mfittko:
**Bug (tooling reliability), pre-0.6.** `scripts/github/reply-resolve-review-threads.mjs` (the PLURAL reply-resolve tool) **hangs indefinitely** instead of completing. Every hung invocation becomes a permanent "running" background task — across a long dev-loop drive these accumulate as stale Claude Code background tasks (observed: hung processes from prior sessions still alive after **1+ day**, e.g. on PRs #942 and #961, several stuck for 25+ hours).

## Evidence
`ps` showed multiple `node scripts/github/reply-resolve-review-threads.mjs --repo … --pr <n> --author copilot-pull-request-reviewer --resolve …` processes with `etime` of 1d01h–1d10h. Killing them immediately fired the long-pending Claude Code background-task completion notifications — confirming the hung tool process was what kept those tasks "running" forever.

## Symptom seen repeatedly this session
The plural tool "times out at 2m" on the orchestrator side; the workaround all session was to fall back to the per-thread `scripts/github/reply-resolve-review-thread.mjs` (`--comment-id --thread-id --body-file`), which returns promptly. So the per-thread sibling is reliable; the plural one is not.

## Root cause (confirmed)
With `--message` set, the plural tool read `process.stdin` to detect a conflicting second message source via an unbounded `for await`. A detached/idle stdin pipe never sends EOF, so the read never resolved; even after being raced against a timeout, the abandoned async iterator left stdin referenced and flowing, keeping the event loop alive so the process never exited. The per-thread sibling never touches stdin (it reads `--body-file`), which is why only the plural tool hung.

## Fix
- Bound the conflict-detection stdin read with a timeout and release the stdin handle (`pause()` + `unref()`) so the tool always terminates (fail-closed / clean exit).
- Add a regression test proving the tool exits (does not hang) on multi-thread input with stdin left open.

## Acceptance criteria
- [x] The plural tool either completes promptly or fails closed within a bounded time; it never produces a process that outlives the call.
- [x] The conflict-detection stdin read is bounded by a timeout.
- [x] Existing conflict-detection behavior is preserved (both `--message` and piped stdin still errors "Choose exactly one message source").
- [x] A regression test proves the tool terminates on multi-thread input with stdin left open; repeated invocations leave no lingering `node …reply-resolve-review-threads` processes.

[...]

PR #1688, opened 2026-08-11
feat(docs): add decision-record validator (#1624)

### LINKED ISSUE #1624: Decision records have no validator: numbering and post-acceptance edits unchecked
## Summary

Decision records have no validator at all. Two rules are unenforced, and one has already been violated silently.

**`ADR-PATH-NUMBERING`.** Nothing checks filename shape or duplicate numeric prefixes.

**`ADR-SUPERSEDE-NOT-REWRITE`.** Nothing prevents editing an accepted record in place. Already bitten: `docs/decisions/0047` had `Partially amended by [0048] ...` appended after acceptance. That is neither the sanctioned Superseded-by flip nor a legal `ADR-STATUS-VALUES` value, and nothing objected.

## Decided approach

New `scripts/docs/validate-decision-records.mjs`, chained into the `test:docs` npm script (`package.json:41`), which `.github/workflows/ci.yml:85` already requires.

Refuse:
1. A filename not matching `^\d{4}-[a-z0-9-]+\.md$`, skipping `0000-template.md`.
2. Any repeated four-digit prefix.
3. In base-ref mode (`git merge-base origin/<default> HEAD`; CI checks out with `fetch-depth: 0`), a changed line outside the Status section of a record whose base Status is Accepted or Superseded.

Rule 3 needs the base-ref comparison, so it must degrade gracefully when the base ref is unavailable rather than failing every local run.

## Acceptance criteria

- [x] A malformed decision-record filename fails.
- [x] A duplicate four-digit prefix fails.
- [x] A post-acceptance edit outside the Status section fails.
- [x] A sanctioned Status flip (Accepted to Superseded) passes.
- [x] A new record added at the next free number passes.
- [x] Unavailable base ref degrades gracefully rather than failing closed on every local run.
- [x] The existing `0047` violation is resolved (either reverted or converted into a proper superseding record).
- [x] Refusals name the rule they uphold.

## Definition of done

- [x] Each check has a test that fails when reverted, proven by mutation.
- [x] `npm run verify` green with the new validator wired into `test:docs`.
- [x] CHANGELOG entry under Unreleased.

## Non-goals

- Judging whether a decision record's content is correct.
- Retro-validating the full existing decision-record history beyond the numbering checks.

mfittko:
## Summary

Decision records (`docs/decisions/*.md`) had no validator. This adds one, `scripts/docs/validate-decision-records.mjs`, chained into the `test:docs` npm script (which `ci.yml` already requires). It enforces two rules:

1. **`ADR-PATH-NUMBERING`** — filenames must match `^\d{4}-[a-z0-9-]+\.md$` (`0000-template.md` exempt) and every four-digit prefix must be unique.
2. **`ADR-SUPERSEDE-NOT-REWRITE`** — in base-ref mode, an Accepted or Superseded record must not change any line outside its Status section (status flips pass); deleting or renaming such a record is also refused. Rule 3 now fails closed in PR CI (the `test:docs` leg fetches the base branch so `merge-base` resolves) and fails on the real fail-open cases (deleted/renamed records) rather than degrading silently.

## Spec

Spec-of-record is issue **#1624** (tracker-first: this repo is `inputSource: tracker`). Its Acceptance criteria, Definition of done, and Non-goals are authoritative and mirrored below.

## Scope and context

Adds the validator, wires it into `test:docs`, makes CI's docs leg have the history it needs (fetch-depth 0 + base-branch fetch), and reverts the existing `0047` Status-section violation. No changes to decision-record content rules or existing record history beyond the `0047` revert.

## Acceptance criteria

- [x] A malformed decision-record filename fails.
- [x] A duplicate four-digit prefix fails.
- [x] A post-acceptance edit outside the Status section fails.
- [x] A sanctioned Status flip (Accepted to Superseded) passes.
- [x] A new record added at the next free number passes.
- [x] Unavailable base ref degrades gracefully rather than failing closed on every local run.
- [x] The existing `0047` violation is resolved (reverted).
- [x] Refusals name the rule they uphold.

## Definition of done

- [x] Each check has a mutation test in `test/docs/validate-decision-records.test.mjs` (21 mutation tests).
- [x] `npm run verify` green with the validator wired in (CI `verify-suite` green on this head; the docs changes are exercised by the `test:docs` leg).
- [x] CHANGELOG entry under Unreleased.

## Non-goals

- Judging whether a decision record's content is correct (including Status-section content beyond the sanctioned flip — explicitly out of scope per issue #1624).

## AC/DoD/Non-goals matrix

| Item | Type | Status | Evidence | Notes |
|---|---|---|---|---|
| A malformed decision-record filename fails. | AC | Met | `test/docs/validate-decision-records.test.mjs` "a malformed decision-record filename fails"; validator `ADR-PATH-NUMBERING` | Mutation-tested |
| A duplicate four-digit prefix fails. | AC | Met | "a duplicate four-digit prefix fails" test | Mutation-tested |
| A post-acceptance edit outside the Status section fails. | AC | Met | "a post-acceptance edit outside the Status section fails" test | Mutation-tested |
| A sanctioned Status flip (Accepted to Superseded) passes. | AC | Met | "a sanctioned Status flip (Accepted to Superseded) passes" test | |
| A new record added at the next free number passes. | AC | Met | "a new record added at the next free number passes" test | |
| Unavailable base ref degrades gracefully rather than failing closed on every local run. | AC | Met | "an unavailable base ref degrades gracefully" + CI-degrade fail-closed tests | |
| The existing `0047` violation is resolved (reverted). | AC | Met | `0047` reverted to canonical `Accepted` status; CHANGELOG entry | |
| Refusals name the rule they uphold. | AC | Met | All tests assert the rule constant (`ADR-PATH-NUMBERING` / `ADR-SUPERSEDE-NOT-REWRITE`) | |
| Each check has a mutation test in `test/docs/validate-decision-records.test.mjs`. | DoD | Met | 21 mutation tests in the suite | |
| `npm run verify` green with the validator wired in. | DoD | Met | CI `verify-suite` green on this head | |
| CHANGELOG entry under Unreleased. | DoD | Met | CHANGELOG Unreleased entry (#1624) | |
| Judging whether a decision record's content is correct (incl. Status-section content beyond the sanctioned flip). | Non-goal | N/A | Scoped out per issue #1624 | Declared non-goal |

## Validation

- `npm run test:docs` (runs the validator inline) — green locally and in CI.
- `node --test test/docs/validate-decision-records.test.mjs` — 21 mutation tests pass.
- Full `npm run verify` is green in CI for this head. The only local red is 48 pre-existing, environment-specific failures in untouched `test/loop` suites (run-watch-cycle, resolve-dev-loop-startup-cli-contract, reply-resolve-review-threads), byte-identical on clean `main`; none touch this PR's diff.

Closes #1624


.github/workflows/ci.yml
```diff
@@ -106,6 +108,16 @@ jobs:
       # Only this matrix leg needs the binary; putting it on PATH via
       # GITHUB_PATH lets test:workflows find the plain `actionlint` command
       # exactly like a local install would.
+      # The decision-record validator's rule 3 resolves the base via
+      # `git merge-base origin/<default> HEAD`. actions/checkout with fetch-depth: 0
+      # for a pull_request checks out the PR merge ref but does NOT fetch
+      # remote-tracking refs like origin/main, so without this the guard silently
+      # degrades to skipped instead of enforcing in PR CI.
+      - name: Fetch base branch (decision-record merge-base)
```
mfittko: **low** (`ci-guard`): The 'cannot silently report green without rule 3 running' guarantee rests solely on the fetch step: if the Fetch base branch step is ever removed or its if regresses, resolveBaseRef has no origin/&lt;default> and the validator degrades rule 3 to skipped while still exiting 0, so test:docs reports GREEN with rule 3 silently unenforced. There is no CI assertion that rule 3 actually executed (state == 'ran' vs 'degraded') on the docs leg. The step is fail-closed only on its own fetch failure (default bash -e aborts, which is correct), not on a silent base-resolution regression.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,243 @@
...
+
+async function resolveBaseRef(git) {
+  let defaultBranch = null;
+  try {
+    defaultBranch = (await git.symbolicRef("refs/remotes/origin/HEAD")).replace(/^refs\/remotes\/origin\//, "");
+  } catch {
+    defaultBranch = null;
+  }
+  if (defaultBranch) return git.mergeBase(`origin/${defaultBranch}`, "HEAD");
+  for (const branch of ["main", "master"]) {
```
mfittko: **question** (`config-drift`): resolveBaseRef falls back to hardcoded &#91;'main','master'] (line 119) when refs/remotes/origin/HEAD is unavailable, while the CI fetch step fetches GITHUB_BASE_REF into origin/&lt;base>. In this repo default==main so the fetched ref is found and rule 3 resolves; latent drift only if the default branch ever becomes a non-main/master name, in which case the validator's main/master fallback would miss the fetched origin/&lt;base> and silently degrade rule 3 (skipped) in PR CI. Confidence: answered, non-manifesting at this head.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,243 @@
...
+    rule3 = { state: "degraded", notice: "base ref unavailable; skipping ADR-SUPERSEDE-NOT-REWRITE post-acceptance edit check" };
+  } else {
+    const changed = await git.diffNameOnly(base, "HEAD", DECISIONS_DIR);
+    for (const rel of changed) {
+      const baseName = path.posix.basename(rel);
+      if (baseName === TEMPLATE || !rel.endsWith(".md")) continue;
+      let baseText;
+      try {
+        baseText = await git.show(`${base}:${rel}`);
+      } catch {
```
mfittko: **low** (`correctness`): The git.show catch blanket-continues, treating ANY git error as 'path absent from base'. This contradicts the in-code invariant (lines 163-165: 'Any error inside the per-record loop is a real failure and must surface (fail closed)'). A non-absent git.show failure (corrupt object, rev-parse error) silently disables the ADR-SUPERSEDE-NOT-REWRITE guard for that record instead of failing closed.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,243 @@
...
+ * the next `## ` heading) and everything else. The Status heading stays in `rest`
+ * (it is identical in both versions, so equality is unaffected).
+ */
+export function splitStatus(text) {
+  const lines = text.split(/\r?\n/);
+  const rest = [];
+  const statusLines = [];
+  let inStatus = false;
+  for (const line of lines) {
+    if (/^##\s+Status\s*$/.test(line)) {
```
mfittko: **low** (`correctness`): splitStatus only enters Status mode on the exact heading /^##\s+Status\s*$/. A record whose Status heading deviates (e.g. ## Status: or ## Status&lt;/code>) never sets inStatus, so its whole body lands in rest and isAcceptedOrSuperseded sees an empty status -> treated as Proposed -> rule 3 silently disabled for that record (fail-open). All current records use the canonical heading so no live impact, but the guard fails open on harmless heading drift.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,243 @@
+#!/usr/bin/env node
+import { execFile } from "node:child_process";
+import { readdir, readFile } from "node:fs/promises";
+import path from "node:path";
+import { promisify } from "node:util";
+import { fileURLToPath } from "node:url";
+
+import { isDirectCliRun } from "../_core-helpers.mjs";
+
+const FILENAME_RE = /^\d{4}-[a-z0-9-]+\.md$/;
```
mfittko, line 10: **nit** (`correctness`): FILENAME_RE &#91;a-z0-9-]+ permits a trailing or repeated hyphen (e.g. 0047-foo-.md, 0047--x.md) while still passing ADR-PATH-NUMBERING.

test/docs/validate-decision-records.test.mjs
```diff
@@ -0,0 +1,266 @@
...
+  }, makeGit({ "docs/decisions/0047-something.md": PROPOSED_BASE }));
+  try {
+    const result = await validateDecisionRecords({ root, git });
+    assert.equal(result.ok, true);
+  } finally {
+    await rm(root, { recursive: true, force: true });
+  }
+});
+
+test("deleting an accepted record fails, naming ADR-SUPERSEDE-NOT-REWRITE", async () => {
```
mfittko, line 218: **medium** (`coverage`): Rename-protection is not mutation-anchored. The documented rule that 'renaming an accepted/superseded record is refused' lives entirely in createGitClient's --no-renames flag (scripts/docs/validate-decision-records.mjs:101). The delete test (line 199) exercises the ENOENT/deletion branch through the fake git client, which never emulates rename-detection collapse, so removing --no-renames from the real git arg breaks no test; rename would silently evade rule 3 (destination path absent from base -> git.show throws -> continue, unblocked). The delete mutation anchor does not extend to the rename claim the PR body/DoD explicitly document.

.github/workflows/ci.yml
```diff
@@ -106,6 +108,16 @@ jobs:
       # Only this matrix leg needs the binary; putting it on PATH via
       # GITHUB_PATH lets test:workflows find the plain `actionlint` command
       # exactly like a local install would.
```
mfittko, line 115: **low** (`gate-evidence`): PR Validation section claims the only local verify red is '48 pre-existing environment-specific gh-mock suite failures, byte-identical on clean main', but this round's recorded local run (validation.json, headSha matching) shows test:scripts red with 1 hang failure in reply-resolve-review-threads (regression #1012), not 48 gh-mock failures. The affected suite test:docs is green with rule 3 'ran', and the CI wiring is honest (fetch-depth 0 + conditional base-branch fetch is fail-closed on fetch failure), so this does not defeat or mislead the gate verdict, but the description's validation claim does not match the recorded run and should be reconciled so a reviewer does not trust a different failure class.

scripts/docs/validate-decision-records.mjs
mfittko: **medium** (`input-validation`): FAIL-OPEN at a trust boundary: the rule-3 per-record git.show catch { continue } swallows ANY git failure as 'path absent from base' (lines 179-184), while the enclosing comment (163-165) promises per-record errors fail closed. Base-resolution failures do degrade correctly, but a real git failure inside the loop silently disables ADR-SUPERSEDE-NOT-REWRITE for that record rather than aborting. This is precisely the 'silently-degraded guard that should fail closed' the contract forbids.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,243 @@
...
+ * retro-validation.
+ */
+export async function validateDecisionRecords({ root, git = createGitClient(root) }) {
+  const dir = path.join(root, DECISIONS_DIR);
+  const errors = [];
+  let names;
+  try {
+    const entries = await readdir(dir);
+    names = entries.filter((e) => e.endsWith(".md")).sort();
+  } catch {
```
mfittko: **low** (`input-validation`): FAIL-OPEN: the readdir(dir) catch sets names = &#91;], silently disabling ALL index checks (ADR-PATH-NUMBERING filename shape + duplicate prefix) if docs/decisions is missing, unreadable, or readdir fails. A genuine filesystem error passes validation with a '0 records scanned' success and no notice - it should fail closed on an internal path that is always expected to exist.

scripts/docs/validate-decision-records.mjs
mfittko: **low** (`input-validation`): Silent guard-disable on input format drift: splitStatus enters Status mode only on the exact /^##\s+Status\s*$/ heading. A record using a non-canonical Status heading (e.g. ## Status: suffix) gets an empty status, isAcceptedOrSuperseded returns false, and rule 3 stops guarding it - the edit-protection guard fails open without any refusal naming a rule.

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `e491c2b4b44ae802fdbeb5e4a067d48d24932591`
**Verdict:** clean

**Findings summary:** 11 angles reviewed; 10 findings (see per-angle breakdown below).

- `ci-guard` → `findings_present` (1 finding)
- `config-drift` → `findings_present` (1 finding)
- `correctness` → `findings_present` (3 findings)
- `determinism` → `clean`
- `input-validation` → `findings_present` (3 findings)
- `link-check` → `clean`
- `pr-description` → `clean`
- `scope` → `clean`

**Next action:** Draft gate CLEAN (0 high). Non-blocking: 1 question (config-drift: resolveBaseRef fallback), 2 in-window medium (coverage: rename-protection mutation anchor; input-validation: git.show blanket-continue fail-open). Fixer triage: 6 low + 1 nit to be deferred. Then close gate and mark PR ready for review.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,272 @@
...
+}
+
+async function main() {
+  const root = resolveDefaultRepoRoot();
+  const result = await validateDecisionRecords({ root });
+  if (result.rule3.notice) process.stdout.write(`${result.rule3.notice}\n`);
+  if (result.ok) {
+    // In CI, rule 3 must actually have run: a silent degrade must not report
+    // green without the post-acceptance-edit guard executing (ci-guard).
+    if (result.rule3.state === "degraded" && process.env.CI) {
```
mfittko: **medium** (`ci-guard`): The fail-closed guarantee that rule 3 actually runs in CI is the only thing making the docs leg red when the base ref is unavailable (main(): degraded && process.env.CI -> exit 1), but it is not mutation-anchored and not unit-testable: main() hardcodes resolveDefaultRepoRoot()/createGitClient, and none of the 18 tests call the CLI path. A regression reverting this guard to the old fail-open behavior (degraded -> exit 0) would leave the docs leg green while ADR-SUPERSEDE-NOT-REWRITE silently never ran in CI — exactly the silent-degrade the guard exists to prevent — with no test and no CI signal catching it.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,272 @@
...
+}
+
+async function resolveBaseRef(git, env = process.env) {
+  let defaultBranch = null;
+  try {
+    defaultBranch = (await git.symbolicRef("refs/remotes/origin/HEAD")).replace(/^refs\/remotes\/origin\//, "");
+  } catch {
+    defaultBranch = null;
+  }
+  if (defaultBranch) return git.mergeBase(`origin/${defaultBranch}`, "HEAD");
```
mfittko, line 144: **low** (`config-drift`): resolveBaseRef resolves origin/HEAD's default branch FIRST and short-circuits before consulting GITHUB_BASE_REF. In CI the new fetch step fetches the PR's actual base (GITHUB_BASE_REF) into origin/&lt;base>, but if actions/checkout creates refs/remotes/origin/HEAD (default), that path wins and the validator compares rule 3 against the DEFAULT branch, not the PR's actual base. For a PR targeting a non-default branch the fetched base ref is dead and rule 3 may diff against the wrong ancestor.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,272 @@
...
+      // old path (-> deletion guard fires) and an added new path (not blocked),
+      // instead of collapsing to just the destination path and evading rule 3.
+      const { stdout } = await run(["diff", "--no-renames", "--name-only", a, b, "--", dir]);
+      return stdout.split(/\r?\n/).filter(Boolean);
+    },
+    async show(spec) {
+      const { stdout } = await run(["show", spec]);
+      return stdout;
+    },
+    async pathExistsIn(rev, rel) {
```
mfittko, line 123: **low** (`correctness`): Residual fail-open from prior (a): pathExistsIn catches ALL git errors and returns false, so the rule-3 loop continues and silently skips the guard for a record when the cat-file -e probe fails for a NON-absence reason (partial clone with missing blob, corrupt object DB) even though the path genuinely exists in base. This contradicts the code's own declared invariant at line 203-205 ('Any error inside the per-record loop is a real failure and must surface (fail closed)'). The blanket git.show continue was correctly removed (git.show now propagates, line 207 + fail-closed test), but the fail-open class was relocated to the cat-file pre-check rather than eliminated.

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `5afee632c59ee79149153595e1f1dee8907be042`
**Verdict:** clean

**Findings summary:** 11 angles reviewed; 11 findings (see per-angle breakdown below).

- `ci-guard` → `findings_present` (1 finding)
- `config-drift` → `findings_present` (1 finding)
- `correctness` → `findings_present` (3 findings)
- `determinism` → `clean`
- `scope` → `clean`
- `link-check` → `clean`

> **nit** (`pr-description`): PR body carries a "Validation" section asserting test/CI results (18 tests pass, npm run verify green in CI, known flaky red) — evidence content that per the review contract belongs in the review verdict, not the PR description. All required sections are otherwise present and complete (Summary, Scope and context, 8 AC, 3 DoD, Non-goals, Closes #1624, spec-of-record stated); no verdict/evidence tables or changelog content present.
> Location: `CHANGELOG.md:1`

**Next action:** Draft gate CLEAN (0 high) at head 5afee632. 3 in-window medium (pathExistsIn fail-open; main() CI-degrade not mutation-anchored), 4 low + 1 nit to defer, 3 questions answered. Fixer triage the mediums, then re-gate at the fixed head before marking ready for review.

scripts/docs/validate-decision-records.mjs
mfittko, line 144: **question** (`correctness`): resolveBaseRef returns the unprotected git.mergeBase(origin/&lt;defaultBranch>, HEAD) immediately when origin/HEAD resolves. If the default branch resolves but that branch is not fetched (origin/HEAD -> origin/develop while only GITHUB_BASE_REF origin/main is fetched), mergeBase throws out of resolveBaseRef and the candidate fallback (GITHUB_BASE_REF/main/master, lines 135-144) is never tried, forcing rule-3 degradation instead of using an available candidate. In CI this can spuriously trip the new fail-closed CI gate (main(): 'degraded in CI' -> exit 1) for a repo whose default branch differs from the PR base. Answered: exotic setup (PRs normally target the default branch), and it is fail-safe (degrades, never a wrong guard), so not a blocker.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,272 @@
...
+    }
+    if (inStatus && /^##\s+\S/.test(line)) inStatus = false;
+    if (inStatus) statusLines.push(line);
+    else rest.push(line);
+  }
+  return { status: statusLines.join("\n"), rest: rest.join("\n") };
+}
+
+/** First non-empty, non-HTML-comment line of the Status section. */
+export function firstMeaningfulLine(text) {
```
mfittko, line 43: **question** (`correctness`): firstMeaningLine only skips lines STARTING with '&lt;!--', so a multi-line leading HTML comment in a base Status section yields the comment's interior first line as the 'meaningful' line; if that interior is not Accepted/Superseded, isAcceptedOrSuperseded returns false and rule 3 silently fails open for the record. Answered: not triggered by the current record set — the template's Status comment is single-line and 0000-template.md is exempt; real base records have no leading Status HTML comment. This is a latent robustness gap in a content-adjacent path, not an active defect.

scripts/docs/validate-decision-records.mjs
mfittko: **medium** (`coverage`): The new CI degrade fail-closed guard (main(): if rule3.state === 'degraded' && process.env.CI -> return 1) is not mutation-anchored by any test. All 18 tests only exercise validateDecisionRecords() directly; none invoke main() / the CLI path. Reverting this branch (back to fail-open degraded -> exit 0) breaks zero tests. It is also not unit-testable as written: main() hardcodes resolveDefaultRepoRoot() and createGitClient(root), so the CI-degrade decision can only be hit in a real repo with CI set — there is no injected-git fixture path to it.

test/docs/validate-decision-records.test.mjs
```diff
@@ -0,0 +1,334 @@
...
+});
+
+test("0000-template.md is exempt from numbering checks", () => {
+  const errors = detectIndexErrors(["0000-template.md"]);
+  assert.deepEqual(errors, []);
+});
+
+// --- Rule 3: post-acceptance edit outside Status section (ADR-SUPERSEDE-NOT-REWRITE) ---
+test("a post-acceptance edit outside the Status section fails, naming ADR-SUPERSEDE-NOT-REWRITE", async () => {
+  const { root, git } = await fixture({
```
mfittko, line 165: **low** (`coverage`): The 'new record at next free number passes' test does not actually exercise the pathExistsIn -> false skip branch (scripts/docs/validate-decision-records.mjs:206). Its fake git has an empty baseFiles map, so diffNameOnly returns &#91;] and the loop body (including the pathExistsIn guard) never runs; the test passes trivially. If pathExistsIn regressed to always-return-true (treating every new record as base-present and thus failing rule 3), this test would not catch it.

scripts/docs/validate-decision-records.mjs
mfittko, line 123: **medium** (`gate-evidence`): Newly added pathExistsIn (this fixer bump) uses a catch-all that swallows ALL git cat-file -e non-zero exits and returns false. The comment claims a non-zero exit always means the path is absent from base, but cat-file also exits non-zero (128-class) on corrupt objects / invalid revs — exactly the real git-failure class the bump claims to fail closed on. Because pathExistsIn runs BEFORE the now-unprotected git.show (line 206 continue), a corrupt object makes it return false first and rule 3 is silently skipped for that record, re-opening the git.show fail-open this bump removes at a NEW call site. The new git.showfailurefailsclosed test only exercises the show path when pathExistsIn returns true, so it gives false confidence; the pathExistsIn catch-all is itself untested for the corrupt-object case.

mfittko, line 123: **low** (`input-validation`): Residual fail-open at the trust boundary between 'base unavailable' and 'real git failure': pathExistsIn (line 110-118) swallows ALL cat-file -e errors and returns false, so a genuine git probe failure (partial clone missing blob, corrupt object DB, git command error) is silently treated as 'path absent from base' and rule 3 is skipped for that record. This widens the documented degradation class ('only base-resolution failures degrade rule 3', per the validateDecisionRecords comment at 178-179 and the PR narrative) — a per-record in-loop git error defeats the guard without surfacing, exactly the class the round-2 fix aimed to fail closed on. Input-validation angle: degradation vs real failure is not cleanly separated at this boundary.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,272 @@
...
+  // Only base-resolution failures degrade rule 3 (the legitimate no-history case).
+  // Any error inside the per-record loop is a real failure and must surface
+  // (fail closed) rather than silently disabling the guard.
+  let base = null;
+  try {
+    base = await resolveBaseRef(git);
+  } catch {
+    base = null;
+  }
+  if (!base) {
```
mfittko, line 208: **question** (`input-validation`): readdir now fails closed (throws on any error, incl. docs/decisions absent) regardless of base availability; a missing directory is treated as an error even in a no-base local run that otherwise degrades rule 3. This means a local repo without docs/decisions (or with an unreadable dir) always fails instead of degrading. Answered: intended per the fail-closed readdir finding (c) and correct — docs/decisions is an internal always-present path; a missing dir is a real repo error, not a degradation case. Not a finding.

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `34915f37f8fe0cfa2990202725d26b9773266cf9`
**Verdict:** findings_present

**Findings summary:** 11 angles reviewed; 5 findings (see per-angle breakdown below).

- `ci-guard` → `clean`
- `config-drift` → `clean`
- `coverage` → `clean`
- `determinism` → `clean`
- `gate-evidence` → `findings_present` (2 findings)
- `pr-description` → `clean`
- `scope` → `clean`
- `link-check` → `clean`

> **medium** (`gate-evidence`): Recorded validation artifact captures only one coarse 'verify' suite (exitCode 1, allPassed false) with null output and stats. It records no per-suite test:docs result, no 18-mutation-test count, and no CI-green result, yet the PR body asserts all three ('npm run test:docs green', '18 mutation tests pass', 'Full npm run verify is green in CI'). The assertions are consistent with the known posture but are not backed by the recorded artifact.
> Location: `tmp/gate-context/mfittko-dev-loops/pr-1688/draft_gate-34915f37f8fe0cfa2990202725d26b9773266cf9.validation.json:1`


> **low** (`gate-evidence`): PR body attributes the only local red to the 'reply-resolve-review-threads run-id coordination test, regression #1012', while the briefing/known posture attributes local red to 'environment gh-mock suites'; the recorded artifact has null output so neither causal attribution is verifiable from evidence, and no explicit statement connects #1012 to the gh-mock suite.
> Location: `tmp/gate-context/mfittko-dev-loops/pr-1688/draft_gate-34915f37f8fe0cfa2990202725d26b9773266cf9.validation.json:1`

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,284 @@
...
+      // exit 0 iff <rev>:<rel> resolves. `git cat-file -e` exits 1 only for a
+      // genuinely absent object/path — a newly added record, the only legitimate
+      // 'not in base' case. Any OTHER non-zero exit (128-class: corrupt object,
+      // invalid rev) is a real git failure and must fail closed rather than
+      // silently disable rule 3 for the record.
+      try {
+        await run(["cat-file", "-e", `${rev}:${rel}`]);
+        return true;
+      } catch (err) {
+        if (err.code === 1) return false; // genuinely absent from base
```
mfittko: **high** (`correctness`): pathExistsIn's 'absent' branch keyed on err.code === 1, but real git cat-file -e &lt;rev>:&lt;path> exits 128 (fatal) for a path absent from a valid rev — verified empirically in this worktree. So the only legitimate 'not in base' case (a newly added record) throws instead of returning false, which propagates out of the rule-3 loop and rejects the entire validator, failing test:docs/CI on ANY PR that adds a decision record. This directly breaks AC 'A new record added at the next free number passes' in real-git execution and inverts the intended fail-closed semantics into a fail-closed false positive. The mutation tests only exercise a fake exec/git that returns code 1, so they validate the branch logic but never git's real exit contract (128-for-absent).

mfittko: **high** (`input-validation`): Trust-boundary defect at the real-git edge: pathExistsIn returns false only on err.code === 1, but git cat-file -e &lt;rev>:&lt;path> exits 128 (fatal) for a path absent from a valid rev (verified empirically). The genuine 'new record not in base' case therefore throws instead of returning false, so the 'absent from base -> skip rule 3' trust boundary is mis-keyed; every added record surfaces as a 'real git failure' and rejects the whole validator, failing CI test:docs. The geniune-error-surfaces fail-closed path now clasifies the legitimate absent case as an error too — a false-positive, not the intended graceful skip.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,284 @@
...
+    ...(env.GITHUB_BASE_REF ? [`origin/${env.GITHUB_BASE_REF}`] : []),
+    ...["main", "master"].map((b) => `origin/${b}`),
+  ];
+  for (const branch of candidates) {
+    try {
+      const base = await git.mergeBase(branch, "HEAD");
+      if (base) return base;
+    } catch {
+      // try next candidate
+    }
```
mfittko, line 158: **low** (`input-validation`): Residual local fail-open at resolveBaseRef: the candidate loop (lines 142-149) and the outer catch (line 194) collapse ALL merge-base failures — including real git failures (corrupt object, invalid rev) — into 'base unavailable', degrading rule 3 to skip. Outside CI run() exits 0 on a degraded rule 3, so a genuine git failure during base resolution silently disables ADR-SUPERSEDE-NOT-REWRITE locally with a green result (CI still fail-closes via run()'s CI guard). This is the one trust boundary where a real git failure does not surface.

.github/workflows/ci.yml
mfittko: Resolved: the CI rule-3 guard now fails closed in CI via run() (`degraded && env.CI` -> exit 1) and is mutation-anchored by the test "run(): degraded rule 3 exits 1 in CI (fail closed), 0 outside CI". No longer rests on the fetch step alone.

scripts/docs/validate-decision-records.mjs
mfittko: Answered: agree the main/master fallback only matters if the default branch stops being main. Non-manifesting at this head; the base branch is fetched into origin/main and rule 3 resolves. Latent drift only.

scripts/docs/validate-decision-records.mjs
mfittko: Resolved: git.show now propagates real failures (no blanket continue); the absent-vs-failure class is handled by pathExistsIn via `git ls-tree` (commit 3d9cafb0), covered by the test "a real git.show failure fails closed instead of silently skipping rule 3".

scripts/docs/validate-decision-records.mjs
mfittko: Resolved: splitStatus now tolerates benign heading drift (e.g. `## Status:` / `## Status;`) so a deviating heading no longer silently fails rule 3 open; covered by "splitStatus tolerates benign trailing punctuation on the Status heading".

scripts/docs/validate-decision-records.mjs
mfittko: Acknowledged nit: FILENAME_RE still permits a trailing/repeated hyphen. Cosmetic; no live record uses it. Deferred as a non-blocking nit.

test/docs/validate-decision-records.test.mjs
mfittko: Resolved: rename-protection is now mutation-anchored — the test "rename-protection is mutation-anchored: diffNameOnly passes --no-renames" asserts the --no-renames flag passes to git diff, so removing it fails the test. No longer lives only in the delete-branch test.

.github/workflows/ci.yml
mfittko: Resolved: the PR Validation section is reconciled — it now states the only local red is the known flaky reply-resolve-review-threads hang (regression #1012), matching the recorded run, and no longer claims the 48 gh-mock failures.

scripts/docs/validate-decision-records.mjs
mfittko: Resolved: git.show's blanket catch/continue was removed — a real git failure inside the rule-3 loop now surfaces (fails closed) instead of silently disabling the guard for that record; covered by the git.show-fails-closed test.

scripts/docs/validate-decision-records.mjs
mfittko: Resolved: readdir now fails closed — a missing/unreadable docs/decisions throws instead of silently setting names=[], covered by "a missing docs/decisions directory fails closed instead of passing with 0 records".

scripts/docs/validate-decision-records.mjs
mfittko: Resolved: splitStatus tolerates a non-canonical Status heading so rule 3 no longer fails open for that record; covered by the splitStatus punctuation test.

scripts/docs/validate-decision-records.mjs
mfittko: Resolved: pathExistsIn now uses `git ls-tree <rev> -- <path>` which exits 0 regardless of presence (non-empty stdout iff the path exists), cleanly distinguishing 'absent from base' (new record -> false) from a real git failure (which still rejects/fails closed). Verified against real git: adding a record passes and rule 3 runs (commit 3d9cafb0).

mfittko: Resolved: same mis-keying of pathExistsIn on git exit 128 fixed via `git ls-tree` (3d9cafb0); the genuine 'new record not in base' case now returns false instead of throwing, so the trust boundary no longer treats a new record as a real git failure.

scripts/docs/validate-decision-records.mjs
mfittko: Deferred: in a local non-CI run, resolveBaseRef collapses real merge-base failures into 'base unavailable' and degrades rule 3. CI is the trust boundary and fail-closes (run() degraded+CI -> exit 1); local degrade is the documented graceful behavior for shallow/no-origin. Low, accepted residual.

scripts/docs/validate-decision-records.mjs
mfittko: Resolved: the CI-degrade fail-closed guard is now mutation-anchored — run() is injectable (root/git/env/out) and covered by "run(): degraded rule 3 exits 1 in CI (fail closed), 0 outside CI", so reverting to fail-open breaks a test.

scripts/docs/validate-decision-records.mjs
mfittko: Answered: origin/HEAD short-circuits before GITHUB_BASE_REF. This repo targets the default branch, so the fetched base equals origin/HEAD; for a non-default-base PR rule 3 would diff the default ancestor. Exotic and fail-safe (degrades, never a wrong guard); documented residual.

scripts/docs/validate-decision-records.mjs
mfittko: Resolved: pathExistsIn no longer swallows all `git cat-file -e` non-zero exits — it uses `git ls-tree` which exits 0 whether the path is absent or present, so a real git failure (corrupt object / invalid rev) rejects and fails closed at the new call site too (commit 3d9cafb0); covered by "pathExistsIn: absent path -> false (new record), real git failure fails closed".

mfittko: Resolved: same pathExistsIn catch-all closed via `git ls-tree`; a genuine git probe failure now surfaces instead of being treated as 'absent from base' and skipping rule 3.

scripts/docs/validate-decision-records.mjs
mfittko: Answered: readdir fails closed intentionally — docs/decisions is an internal always-present path; a missing/unreadable directory is a real repo error, not a degradation case. Agreed, not a finding.

scripts/docs/validate-decision-records.mjs
mfittko: Resolved: the fail-open class that was relocated to the cat-file pre-check is eliminated — pathExistsIn uses `git ls-tree` (non-empty stdout iff present), so a non-absence probe failure rejects and fills closed rather than skip the guard for the record.

scripts/docs/validate-decision-records.mjs
mfittko: Answered: exotic setup (PRs normally target the default branch) and fail-safe — it degrades, never applies a wrong guard. Not a blocker; accepted.

scripts/docs/validate-decision-records.mjs
mfittko: Answered: a multi-line leading HTML comment in a base Status section is not triggered by the current record set (template comment is single-line, 0000-template.md exempt). Latent robustness gap in a content-adjacent path, not an active defect; accepted.

scripts/docs/validate-decision-records.mjs
mfittko: Resolved: main()/run() is now injectable (root/git/env/out) so the CI-degrade decision is unit-testable and mutation-anchored — reverting the degraded->exit 1 branch fails "run(): degraded rule 3 exits 1 in CI".

test/docs/validate-decision-records.test.mjs
mfittko: Covered: the pathExistsIn->false skip branch is exercised directly by the dedicated "pathExistsIn: absent path -> false (new record), real git failure fails closed" unit test, and the integration semantics for a new record at the next free number are covered by "a new record added at the next free number passes".

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,281 @@
...
+ */
+export function splitStatus(text) {
+  const lines = text.split(/\r?\n/);
+  const rest = [];
+  const statusLines = [];
+  let inStatus = false;
+  for (const line of lines) {
+    // Tolerate benign trailing punctuation (e.g. `## Status:`); a deviating
+    // heading must not silently fail open and disable rule 3 for the record.
+    if (/^##\s+Status\s*[:;]?\s*$/.test(line)) {
```
mfittko, line 29: **low** (`correctness`): splitStatus only recognizes the Status heading as /^##\s+Status\s*&#91;:;]?\s*$/; any data-valid deviating heading (e.g. '## Status.' or '## Status note') silently yields an empty status section, so isAcceptedOrSuperseded() returns false and rule 3 (ADR-SUPERSEDE-NOT-REWRITE) is silently disabled for that record. The code's own comment (lines 25-26) promises 'a deviating heading must not silently fail open and disable rule 3', but only the exact form plus trailing ':'/';' are tolerated; every other deviation fails open. Confirmed via runtime check: splitStatus('## Status.') -> isAccepted false, statusLen 0. Not currently live (all 49 existing records use canonical '## Status'), but a future record that drifts loses post-acceptance protection with no warning.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,281 @@
...
+  // No origin/HEAD. Try an env-provided base first (CI sets GITHUB_BASE_REF and
+  // fetches it into origin/<base>, so a non-main/master default branch still
+  // resolves), then conventional default names.
+  const candidates = [
+    ...(env.GITHUB_BASE_REF ? [`origin/${env.GITHUB_BASE_REF}`] : []),
+    ...["main", "master"].map((b) => `origin/${b}`),
+  ];
+  for (const branch of candidates) {
+    try {
+      const base = await git.mergeBase(branch, "HEAD");
```
mfittko, line 154: **low** (`coverage`): resolveBaseRef's candidate-fallback loop (origin/HEAD missing -> try origin/GITHUB_BASE_REF -> origin/main -> origin/master, and 'first candidate rejects then a later one resolves') has no mutation test. All 21 tests stub mergeBase to either always resolve or always throw, so the try-next-candidate path and the env-provided-base path are untested -- a regression that swapped candidate order or broke the env fallback would go undetected.

scripts/docs/validate-decision-records.mjs
mfittko, line 29: **low** (`coverage`): splitStatus fails open when a record puts its Status value inline on the heading line (e.g. '## Status: Accepted -- ...' or '## Status -- Accepted'): the start regex /^##\s+Status\s*&#91;:;]?\s*$/ requires the line to END after the optional punctuation, so trailing content keeps inStatus false, the Status section is never collected, status is empty, isAcceptedOrSuperseded returns false, and rule 3 (ADR-SUPERSEDE-NOT-REWRITE) is silently disabled for that record. This contradicts the code's own stated invariant that 'a deviating heading must not silently fail open and disable rule 3 for the record', and is untested. Given the 0047 incident shows this file family already receives nonstandard Status content, the fail-open is a real latent gap.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,281 @@
...
+ * pre-existing 0047 incident of that class was resolved by reverting the record, not by
+ * retro-validation.
+ */
+export async function validateDecisionRecords({ root, git = createGitClient(root) }) {
+  const dir = path.join(root, DECISIONS_DIR);
+  const errors = [];
+  let names;
+  try {
+    const entries = await readdir(dir);
+    names = entries.filter((e) => e.endsWith(".md")).sort();
```
mfittko, line 188: **low** (`input-validation`): readdir input is pre-filtered with entries.filter((e) => e.endsWith('.md')).sort(), so any non-.md file dropped into docs/decisions/ (e.g. a new '0047-something.txt') is silently ignored: it never reaches detectIndexErrors (ADR-PATH-NUMBERING) and, being absent from base, never hits rule 3. This is an inconsistent validation surface against the documented 'filename must match NNNN-&lt;slug>.md' contract; a stray non-record file in the decisions dir passes with zero signal.

scripts/docs/validate-decision-records.mjs
mfittko, line 29: **low** (`input-validation`): Status-heading normalization/trimming is overly strict and inconsistent: only '## Status', '## Status:', '## Status;' end the section-detection tolerance, so any other punctuation drift ('## Status.', '## Status: More') is treated as 'not a Status section' and isAcceptedOrSuperseded() returns false — silently dropping rule-3 protection for that record rather than normalizing it or surfacing an error. The base-vs-current comparison therefore depends on the heading matching an exact rigid form, an inconsistent-normalization gap across the accepted input surface.

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `3d9cafb04056103ea0607825ee1fae8a16ddf2f6`
**Verdict:** clean

**Findings summary:** 11 angles reviewed; 6 findings (see per-angle breakdown below).

- `ci-guard` → `clean`
- `config-drift` → `clean`
- `determinism` → `clean`
- `link-check` → `clean`
- `pr-description` → `clean`
- `scope` → `clean`

> **low** (`gate-evidence`): No draft_gate checkpoint verdict comment references the current head SHA 3d9cafb040. The most recent draft_gate verdict (round 3, 2026-08-11T22:33:07Z) references prior head 34915f37f8; head then moved to 3d9cafb040 (fix commit 'pathExistsIn distinguishes absent-vs-failure via git ls-tree') with no new verdict on record. PR is still draft (OPEN, isDraft true) and is correctly NOT transitioned to ready/merge.
> Location: `tmp/gate-context/mfittko-dev-loops/pr-1688/draft_gate-3d9cafb04056103ea0607825ee1fae8a16ddf2f6.json:1`

**Next action:** Move the PR from draft to ready-for-review

scripts/docs/validate-decision-records.mjs
mfittko: Deferred at gate close (round 4, fingerprint 854f366bea4460c3, severity low, angle correctness): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer; #1585); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/docs/validate-decision-records.mjs
mfittko: Deferred at gate close (round 4, fingerprint f7c43996c8b0e305, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer; #1585); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/docs/validate-decision-records.mjs
mfittko: Deferred at gate close (round 4, fingerprint 719a95d48098336f, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer; #1585); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/docs/validate-decision-records.mjs
mfittko: Deferred at gate close (round 4, fingerprint 6de6dfa63f25239d, severity low, angle input-validation): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer; #1585); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/docs/validate-decision-records.mjs
mfittko: Deferred at gate close (round 4, fingerprint 84c28dcf754657af, severity low, angle input-validation): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer; #1585); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,281 @@
...
+}
+
+/** First non-empty, non-HTML-comment line of the Status section. */
+export function firstMeaningfulLine(text) {
+  const line = text
+    .split(/\r?\n/)
+    .map((l) => l.trim())
+    .find((l) => l.length > 0 && !l.startsWith("<!--"));
+  return line || "";
+}
```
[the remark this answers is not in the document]
mfittko: Fixed in 3f7d8ab9: firstMeaningfulLine now skips the entire HTML comment block (single- and multi-line), so the template's instructional comment no longer hides the real status line. Added a test covering the multi-line comment form.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,281 @@
...
+import { isDirectCliRun } from "../_core-helpers.mjs";
+
+const FILENAME_RE = /^\d{4}-[a-z0-9-]+\.md$/;
+const TEMPLATE = "0000-template.md"; // permanently reserved template, not a record
+const DECISIONS_DIR = "docs/decisions";
+
+/**
+ * Split a decision record into its Status section (lines under `## Status` up to
+ * the next `## ` heading) and everything else. The Status heading stays in `rest`
+ * (it is identical in both versions, so equality is unaffected).
```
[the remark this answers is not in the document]
mfittko: Fixed in 3f7d8ab9: the splitStatus docstring now states that the heading's own text is identical in both versions (so it never affects the rest-equality comparison) and that a benign deviating heading is still recognized as the section boundary.

test/docs/validate-decision-records.test.mjs
```diff
@@ -0,0 +1,392 @@
...
+  assert.equal(isAcceptedOrSuperseded(splitStatus(SUPERSEDED_4047).status), true);
+  assert.equal(isAcceptedOrSuperseded(splitStatus("## Status\n\nProposed\n").status), false);
+});
+
+test("firstMeaningfulLine ignores the template's HTML comment", () => {
+  assert.equal(
+    firstMeaningfulLine("<!-- exactly one of: ... -->\nProposed"),
+    "Proposed",
+  );
+});
```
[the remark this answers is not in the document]
mfittko: Fixed in 3f7d8ab9: added a test asserting firstMeaningfulLine skips a multi-line HTML comment block (and handles a comment-only input / empty input), matching the real 0000-template.md form.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,281 @@
...
+  return { ok: errors.length === 0, errors, filesScanned: names.length, rule3 };
+}
+
+function resolveDefaultRepoRoot() {
+  return fileURLToPath(new URL("../../", import.meta.url));
+}
+
+function renderError(error) {
+  return `${error.rule}: ${error.file} - ${error.message}`;
+}
```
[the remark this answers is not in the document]
mfittko: Fixed in 3f7d8ab9: renderError now returns the error.message (which already embeds the rule and, where relevant, the file) instead of re-prefixing rule:file. The duplicate-prefix message now includes the current file name so no filename info is lost.

.github/workflows/ci.yml
```diff
@@ -90,6 +90,8 @@ jobs:
     steps:
       - name: Check out repository
         uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
```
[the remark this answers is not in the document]
mfittko: Addressed in 3f7d8ab9: scoped fetch-depth via the verify-suite matrix — only the test:docs leg uses fetch-depth: 0 (needed for the decision-record merge-base); every other leg uses the default depth 1, avoiding the extra checkout/bandwidth cost for legs that don't need full history. Also moved the actionlint comment above the Install actionlint step it describes.

.github/workflows/ci.yml
```diff
@@ -86,10 +87,17 @@ jobs:
           - test:pack
           - test:dev-loop
           - test:workflows
+        include:
+          # Only the test:docs leg needs full history for the decision-record
+          # merge-base comparison; every other leg checks out the default (depth 1).
+          - suite: test:docs
+            fetch-depth: 0
```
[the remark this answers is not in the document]
mfittko: Renamed the matrix key from `fetch-depth` to `fetch_depth` (both the `include` entry and the `${{ matrix.fetch_depth }}` reference in the checkout step) to avoid hyphenated-name expression parsing issues. Resolving commit: f447852e.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
...
...it skips only rule 3 rather than failing every local run. `verify-suite`'s checkout now uses `fetch-depth: 0` so the base-ref comparison has the history it needs in CI, and the `test:docs` leg additionally fetches the base branch (actions/checkout does not fetch `origin/<default>` remote-tracking refs on a `pull_request`, which previously made rule 3 silently degrade instead of enforcing in PR CI). Rule 3 now fails closed on the real fail-open cases it previously swallowed: only base-resolution failure degrades it, while deleting an Accepted/Superseded record (and, via `--no-renames`, renaming one) is refused as an `ADR-SUPERSEDE-NOT-REWRITE` violation. The existing `0047` violation — a "Partially amended by [0048]" line appended to an Accepted record's Status after acceptance, which is neither the sanctioned Superseded-by flip nor a legal `ADR-STATUS-VALUES` value — is reverted to the clean `Accepted` status, consistent with how `0021`'s partial supersession by `0047` is recorded (the older record stays `Accepted`; the newer record documents the partial amendment). Each check has a mutation test in `test/docs/validate-decision-records.test.mjs`; refusals name the rule they uphold.
```
mfittko, line 8: **low** (`acceptance-criteria`): DoD item 'npm run verify green' is not locally reproducible at this head: the gate-validation record stamps npm run verify exit 1 (48 failures in test/loop run-watch-cycle, resolve-dev-loop-startup-cli-contract, reply-resolve-review-threads — none touch this diff). The DoD is satisfied only via the PR's CI verify-suite assertion, not a recorded local green at head.

mfittko, line 8: **low** (`contradiction-lens`): Run() fails closed in CI on a degraded rule 3 (validate-decision-records.mjs:275-279: env.CI && rule3.degraded -> exit 1), but the CHANGELOG states rule 3 'degrades gracefully when the base ref is unavailable (shallow checkout, no origin)' and 'only base-resolution failure degrades it' as if unconditional. Both claims hold only for local (non-CI) runs; in CI an unavailable base is a hard failure, not a graceful degrade. The two CHANGELOG sentences are only reconciled by an unstated env.CI distinction, and 'only base-resolution failure degrades it' is itself false in CI.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,296 @@
...
+      // empty output is a cleanly-positive "not in base" (a newly added record —
+      // the only legitimate skip case) while a real git failure still rejects
+      // and fails closed. A base resolved via mergeBase is always a valid rev.
+      const { stdout } = await run(["ls-tree", "-z", rev, "--", rel]);
+      return stdout.length > 0;
+    },
+  };
+}
+
+async function resolveBaseRef(git, env = process.env) {
```
mfittko, line 137: **low** (`dry`): resolveBaseRef re-implements the default-branch candidate + merge-base loop already present in scripts/loop/resolve-dev-loop-startup.mjs (ISSUELESS_BASE_REF_CANDIDATES / resolveIssuelessMergeBase): try origin/HEAD, then iterate origin/main|origin/master candidates, with try/catch skip-per-candidate. Same pattern, second copy.

scripts/docs/validate-decision-records.mjs
```diff
@@ -0,0 +1,296 @@
...
+ */
+export function detectIndexErrors(names) {
+  const errors = [];
+  const seen = new Map();
+  for (const name of names) {
+    if (name === TEMPLATE) continue; // template is exempt, not a record
+    if (!FILENAME_RE.test(name)) {
+      errors.push({
+        kind: "adr_filename",
+        rule: "ADR-PATH-NUMBERING",
```
mfittko, line 79: **nit** (`dry`): Rule-name string literals ADR-PATH-NUMBERING / ADR-SUPERSEDE-NOT-REWRITE are hardcoded ~10 times across the error objects (both in the structured rule field lines 79/89/232/244 and embedded inside the corresponding message strings lines 81/92/234/246) with no shared constant.

mfittko (commented):


**Reviewed head SHA:** `feaf2e0ef46f29cdf00677dc622c519d29ae9741`
**Verdict:** clean

**Findings summary:** 7 angles reviewed; 5 findings (see per-angle breakdown below).

- `contradiction-lens` → `findings_present` (2 findings)
- `dry` → `findings_present` (2 findings)
- `yagni` → `clean`
- `docs` → `clean`

> **question** (`contradiction-lens`): 0048's Status carries 'Partially amends &#91;0047]' — the exact mirror of the 'Partially amended by &#91;0048]' line in 0047 that issue #1624 judged 'neither the sanctioned Superseded-by flip nor a legal ADR-STATUS-VALUES value' and which this PR reverts. The PR treats the newer-record 'Partially amends' phrasing as the sanctioned documentation convention while rejecting the older-record mirror as illegal, a double-standard in the same ADR-STATUS-VALUES discipline.
> Location: `docs/decisions/0048-gate-full-dispatches-grouped-two-knob-dispatch-bounds.md:7`

**Next action:** Proceed to merge: current-head pre_approval_gate is clean (Copilot converged round 2/2, CI green, 0 unresolved).

CHANGELOG.md
mfittko: Deferred to follow-up: DoD 'npm run verify green' is satisfied via the CI verify-suite (green on this head); the local 48 are pre-existing, out-of-diff test/loop failures. Non-blocking (low) for the clean pre_approval_gate verdict (blockCleanOnFindingSeverities: high).

mfittko: Deferred to follow-up: documentation-wording refinement of the CHANGELOG rule-3 graceful-degrade phrasing to scope it to local (non-CI) runs. Non-blocking (low) for the clean pre_approval_gate verdict.

scripts/docs/validate-decision-records.mjs
mfittko: Deferred to follow-up: merge-base resolution shares a pattern with resolve-dev-loop-startup.mjs; deliberate for standalone-script isolation but worth a shared helper later. Non-blocking (low) for the clean pre_approval_gate verdict.

scripts/docs/validate-decision-records.mjs
mfittko: Deferred to follow-up: hoist the ADR rule-name string literals into shared constants. Cosmetic (nit), non-blocking for the clean pre_approval_gate verdict.

PR #1742, opened 2026-08-17
gate: refuse a --verdict contradicting the wrapper overallVerdict (#1641)

### LINKED ISSUE #1641: write-gate-findings-log: refuse a --verdict contradicting the wrapper's overallVerdict
## Summary

Follow-up from #1616's Copilot review. `write-gate-findings-log.mjs` now persists `overallVerdict` (threaded from `consolidate-fanin.mjs --ledger-out`'s `{ overallVerdict, findings }` wrapper, unwrapped in `scripts/github/_findings-input.mjs`) alongside the caller-passed `--verdict`. The two can disagree: an orchestrator can pass a `--verdict` that contradicts the wrapper's `overallVerdict`, and the durable ledger records both, leaving an ambiguous `verdict` vs `overallVerdict` pair. On current main the producer normalizes the wrapper value (`scripts/github/write-gate-findings-log.mjs`, `writeGateFindingsLog`, lines 451-460) and writes both fields into the log object (lines 495-510) without any equality check.

`upsert-checkpoint-verdict.mjs` enforces consistency against `overallVerdict` (#1616; refusal at `scripts/github/upsert-checkpoint-verdict.mjs:1793-1808`), so the contradiction in the legacy `verdict` field does not affect enforcement. But the producer accepting the contradiction is the same defect shape #1616 closes at the consumer: a producer accepts input the contract declares invalid.

## Proposed approach

When the wrapper carries `overallVerdict` AND the two values disagree, `write-gate-findings-log.mjs` fails closed, mirroring `upsert-checkpoint-verdict.mjs`'s refusal. Details, grounded in current main:

- Placement: extend the existing `overallVerdict` normalization block in `writeGateFindingsLog` (`scripts/github/write-gate-findings-log.mjs:451-460`). After `normalizeVerdict(overallVerdict)` succeeds, compare the normalized value to `options.verdict` with strict string equality. `--verdict` is already normalized at parse time (line 353-357), so both sides are canonical before comparison.
- Error shape: throw via `parseError(...)`, matching the malformed-wrapper rejection at line 455. `main()`'s catch (lines 552-558) emits `{ ok: false, error }` and exit code 1, same as every other validation failure in this script.
- Error content: name both values (the passed `--verdict` and the wrapper's `overallVerdict`) and cite `GATE-COMMENT-VERDICT-VALUES` (`skills/docs/gate-review-comment-contract.md`), mirroring the consumer's refusal text at `upsert-checkpoint-verdict.mjs:1807`.
- Domain: `normalizeVerdict` accepts `clean|findings_present|blocked` for both flags. `consolidate-fanin.mjs` only emits `clean|findings_present` into the wrapper, but the check is domain-agnostic: any strict inequality fails closed, including `--verdict blocked` against `overallVerdict findings_present`.
- Matching `--verdict`: accepted; the ledger writes both `verdict` and `overallVerdict` exactly as today.
- `overallVerdict` absent (bare-array `--findings` / `--findings-file`, legacy producers): current behavior preserved byte-identically. The existing test "writeGateFindingsLog omits overallVerdict for a bare-array input" (`test/github/write-gate-findings-log.test.mjs:1390`) pins this.
- `--verdict` stays required at the parse layer (line 394-398). This script does not gain a derive-from-wrapper default; that derivation belongs to the consumer (`upsert-checkpoint-verdict.mjs`).
- Caller impact: the documented flow (`skills/docs/gate-review-sub-loop-contract.md`, `GATE-EXEC-DISPOSITION-LEDGER`) feeds `--ledger-out`'s wrapper straight to `--findings-file` with the consolidator's own verdict, so a compliant round is unaffected. Only a contradicting caller breaks, which is the point.

## Origin

Copilot review comment on PR #1640 (issue #1616). Out of #1616's upsert-scoped AC; filed separately to avoid scope growth.

## Acceptance criteria

- [x] `writeGateFindingsLog` rejects a wrapper `overallVerdict` that differs from `--verdict`: the call throws before any ledger file is written, and no `<gate>-<headSha>.json` appears under `--tmp-root`.
- [x] The rejection error names both values (the passed `--verdict` and the wrapper's `overallVerdict`) and cites `GATE-COMMENT-VERDICT-VALUES` with the contract doc path `skills/docs/gate-review-comment-contract.md`.
- [x] A matching `--verdict` and wrapper `overallVerdict` writes the ledger with both fields, identical to current behavior (existing test at `test/github/write-gate-findings-log.test.mjs:1348` stays green).
- [x] A bare-array input (no wrapper) is unaffected: ledger writes byte-identically, no `overallVerdict` key (existing test at `test/github/write-gate-findings-log.test.mjs:1390` stays green).
- [x] The malformed-wrapper rejection (`overallVerdict: "bogus"`) still fires before the contradiction check (existing test at `test/github/write-gate-findings-log.test.mjs:1370` stays green).
- [x] CLI path: a direct `write-gate-findings-log.mjs` invocation with a contradicting pair exits 1 with a `{ ok: false, error }` JSON line on stderr, matching the script's existing failure shape.
- [x] New regression test in `test/github/write-gate-findings-log.test.mjs` covers the contradiction rejection (both directions: `--verdict clean` vs `overallVerdict findings_present`, and the reverse) and asserts the error message names both values.
- [x] `test/contracts/gate-verdict-consistency-contract.test.mjs` gains a producer-side pin: `write-gate-findings-log.mjs` cites `GATE-COMMENT-VERDICT-VALUES` and the contract doc path, mirroring the existing upsert-side citation test.

## Definition of done

- [x] `npm run verify` green.
- [x] `npm run assets:check` green.
- [x] `npm run schema:check` green.
- [x] `node --test test/github/write-gate-findings-log.test.mjs` green.
- [x] `node --test test/contracts/gate-verdict-consistency-contract.test.mjs` green.
- [x] Cross-harness non-regression (#1086).
- [x] Scope-bounded; no unrelated changes.

## Non-goals

- Changing how `overallVerdict` is computed (`consolidate-fanin.mjs` / `@dev-loops/core/loop/gate-fanin` stay untouched).
- Removing the `verdict` field from the ledger (backward compat for inline/fallback paths).
- Making `--verdict` optional or deriving it from the wrapper; that derive-or-refuse behavior lives in `upsert-checkpoint-verdict.mjs` (#1616) and stays there.
- Changing `post-gate-findings.mjs`, which unwraps and ignores `overallVerdict` by design.
- Any change to provenance, angle-coverage, or judge-verdict handling in this script.

## Refined (2026-08-17)

Auto-grill against main 81e86581: defect confirmed present (producer normalizes `overallVerdict` at write-gate-findings-log.mjs:451-460 and records both fields at 495-510 with no equality check; consumer refusal exists only at upsert-checkpoint-verdict.mjs:1793-1808); line pins, error-shape, domain, and test anchors verified from source.

mfittko:
## Summary

`write-gate-findings-log.mjs` now refuses a `--verdict` that contradicts the wrapper's `overallVerdict`, so an ambiguous verdict pair can never be persisted as the round's durable findings ledger. The refusal fires at write time, before any ledger file is created, and cites `GATE-COMMENT-VERDICT-VALUES` plus the owning contract doc.

## Scope and context

Producer-side guard only, inside `writeGateFindingsLog`. The check validates the caller verdict's domain first (a non-string or out-of-domain value gets the parser's own domain error, mirroring the CLI parse path), then compares normalized-vs-normalized, and persists the canonical verdict value. The malformed-wrapper rejection still fires first. Contract docs (`gate-review-sub-loop-contract.md`, `gate-review-comment-contract.md`) state the new write-time enforcement point; their generated mirrors are regenerated in the same change.

## Acceptance criteria

- [x] `writeGateFindingsLog` rejects a wrapper `overallVerdict` that differs from `--verdict`: the call throws before any ledger file is written, and no `<gate>-<headSha>.json` appears under `--tmp-root` (asserted via ENOENT on the exact `buildLogPath` target).
- [x] The rejection error names both values and cites `GATE-COMMENT-VERDICT-VALUES` with the contract doc path `skills/docs/gate-review-comment-contract.md`.
- [x] A matching `--verdict` and wrapper `overallVerdict` writes the ledger with both fields, identical to current behavior (existing test at `test/github/write-gate-findings-log.test.mjs` stays green).
- [x] A bare-array input (no wrapper) is unaffected: ledger writes byte-identically, no `overallVerdict` key.
- [x] The malformed-wrapper rejection (`overallVerdict: "bogus"`) still fires before the contradiction check.
- [x] CLI path: a contradicting pair exits 1 with a `{ ok: false, error }` JSON line on stderr.
- [x] New regression tests cover the contradiction rejection in both directions plus the `blocked` vs `findings_present` pair, and assert the error names both values.
- [x] `test/contracts/gate-verdict-consistency-contract.test.mjs` gains a producer-side pin anchored on the distinctive refusal literal (verified to fail on pre-change main).

## Definition of done

- [x] `npm run verify` green (gate validation artifact recorded at the current head; one known-flaky stdin-idle test in `reply-resolve-review-threads` passed in isolation, tracked by the verify-suite stabilization issue).
- [x] `npm run assets:check` green (95 checked).
- [x] `npm run schema:check` green.
- [x] `node --test test/github/write-gate-findings-log.test.mjs` green (86 tests).
- [x] `node --test test/contracts/gate-verdict-consistency-contract.test.mjs` green.
- [x] Cross-harness non-regression: the change is a harness-neutral producer script plus contract docs; the generated `.claude` mirrors were regenerated byte-identical (`assets:check` 95 checked) and no Pi-only or Claude-only path is touched.
- [x] Scope-bounded; no unrelated changes.

|---|---|---|
| Contradicting verdict pair throws before any ledger write (ENOENT on the buildLogPath target) | write-gate-findings-log test suite green (86 tests) | overallVerdict computation unchanged (consolidate-fanin untouched) |
| Refusal names both values and cites GATE-COMMENT-VERDICT-VALUES plus the contract doc path | producer-side contract pin green (gate-verdict-consistency suite) | no derive-from-wrapper default (stays in upsert-checkpoint-verdict) |
| Matching pair and bare-array inputs write byte-identically to before | existing matching/bare-array tests stay green | verdict field kept in the ledger (backward compat) |
| Malformed-wrapper rejection still fires first | malformed-wrapper ordering test green | post-gate-findings unchanged (unwraps and ignores overallVerdict) |
| CLI contradicting pair exits 1 with the ok-false JSON error shape | CLI-path test asserts exit code equal 1 | judge-verdict interaction contract is a tracked follow-up |
| Both contradiction directions plus the blocked pair covered; error names both values | regression tests green in the producer suite | — |
| Producer-side citation pin fails on pre-change main | contract-test tightening verified against main | — |
| Full verification | npm run verify, assets:check (95), schema:check all green | — |

## Non-goals

- Changing how `overallVerdict` is computed (`consolidate-fanin.mjs` / `@dev-loops/core/loop/gate-fanin` untouched).
- Removing the `verdict` field from the ledger (backward compat for inline/fallback paths).
- Making `--verdict` optional or deriving it from the wrapper; that derive-or-refuse behavior lives in `upsert-checkpoint-verdict.mjs` and stays there.
- Changing `post-gate-findings.mjs`, which unwraps and ignores `overallVerdict` by design.
- Any change to provenance, angle-coverage, or judge-verdict handling in this script (the judge-verdict interaction contract is a tracked follow-up).

## Validation

```sh
node --test test/github/write-gate-findings-log.test.mjs
node --test test/contracts/gate-verdict-consistency-contract.test.mjs
npm run assets:check && npm run schema:check
npm run verify
```

Closes #1641

### DIFF
--- a/scripts/github/write-gate-findings-log.mjs
+++ b/scripts/github/write-gate-findings-log.mjs
@@ -459,10 +459,14 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
     normalizedOverallVerdict = verdict;
     // Fail closed on a caller-passed --verdict that contradicts the
     // consolidator's computed verdict, mirroring the consumer-side refusal in
-    // upsert-checkpoint-verdict.mjs (#1616, GATE-COMMENT-VERDICT-VALUES). Both
-    // values are already canonical here (--verdict normalized at parse time,
-    // overallVerdict normalized above), so strict string equality is correct.
-    if (options.verdict !== normalizedOverallVerdict) {
+    // upsert-checkpoint-verdict.mjs (#1616, GATE-COMMENT-VERDICT-VALUES).
+    // Normalize options.verdict here so the comparison is symmetric for every
+    // caller — a direct programmatic caller passes a non-canonical verdict by
+    // the same normalizeVerdict path the CLI parse normally canonicalizes it
+    // through, so a canonical-value wrapper never false-rejects a
+    // differently-cased/whitespaced but semantically equal caller verdict.
+    const callerVerdict = normalizeVerdict(options.verdict);
+    if (callerVerdict !== normalizedOverallVerdict) {
       throw parseError(
         `--verdict ${JSON.stringify(options.verdict)} contradicts the wrapper's "overallVerdict" ${JSON.stringify(normalizedOverallVerdict)} (GATE-COMMENT-VERDICT-VALUES; skills/docs/gate-review-comment-contract.md)`,
       );
--- a/scripts/github/write-gate-findings-log.mjs
+++ b/scripts/github/write-gate-findings-log.mjs
@@ -465,12 +465,25 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
     // the same normalizeVerdict path the CLI parse normally canonicalizes it
     // through, so a canonical-value wrapper never false-rejects a
     // differently-cased/whitespaced but semantically equal caller verdict.
-    const callerVerdict = normalizeVerdict(options.verdict);
+    // Validate options.verdict's own domain first (same message the CLI parse
+    // path throws for --verdict) so a null/undefined/non-string/out-of-domain
+    // value is reported as an invalid verdict, not misattributed as a
+    // contradiction with the wrapper. typeof is checked before normalizeVerdict
+    // runs because normalizeVerdict coerces via String(), which would otherwise
+    // let a non-string (e.g. an array) silently pass as its stringified form.
+    const callerVerdict = typeof options.verdict === "string" ? normalizeVerdict(options.verdict) : null;
+    if (!callerVerdict) {
+      throw parseError("--verdict must be clean, findings_present, or blocked");
+    }
     if (callerVerdict !== normalizedOverallVerdict) {
       throw parseError(
-        `--verdict ${JSON.stringify(options.verdict)} contradicts the wrapper's "overallVerdict" ${JSON.stringify(normalizedOverallVerdict)} (GATE-COMMENT-VERDICT-VALUES; skills/docs/gate-review-comment-contract.md)`,
+        `--verdict ${JSON.stringify(callerVerdict)} contradicts the wrapper's "overallVerdict" ${JSON.stringify(normalizedOverallVerdict)} (GATE-COMMENT-VERDICT-VALUES; skills/docs/gate-review-comment-contract.md)`,
       );
     }
+    // Persist the canonical (normalized) verdict, not the raw caller value, so
+    // the durable ledger never records a differently-cased/whitespaced verdict
+    // beside the canonical overallVerdict it was just checked against.
+    options.verdict = callerVerdict;
   }
   let provenance;
   if (options.provenance === undefined) {
--- a/scripts/github/write-gate-findings-log.mjs
+++ b/scripts/github/write-gate-findings-log.mjs
@@ -457,6 +457,16 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
       );
     }
     normalizedOverallVerdict = verdict;
+    // Fail closed on a caller-passed --verdict that contradicts the
+    // consolidator's computed verdict, mirroring the consumer-side refusal in
+    // upsert-checkpoint-verdict.mjs (#1616, GATE-COMMENT-VERDICT-VALUES). Both
+    // values are already canonical here (--verdict normalized at parse time,
+    // overallVerdict normalized above), so strict string equality is correct.
+    if (options.verdict !== normalizedOverallVerdict) {
+      throw parseError(
+        `--verdict ${JSON.stringify(options.verdict)} contradicts the wrapper's "overallVerdict" ${JSON.stringify(normalizedOverallVerdict)} (GATE-COMMENT-VERDICT-VALUES; skills/docs/gate-review-comment-contract.md)`,
+      );
+    }
   }
   let provenance;
   if (options.provenance === undefined) {
--- a/scripts/github/write-gate-findings-log.mjs
+++ b/scripts/github/write-gate-findings-log.mjs
@@ -449,6 +449,7 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
   // malformed wrapper cannot silently record an invalid verdict as the
   // consolidator's truth (#1616).
   let normalizedOverallVerdict;
+  let persistedVerdict = options.verdict;
   if (overallVerdict !== undefined) {
     const verdict = normalizeVerdict(overallVerdict);
     if (!verdict) {
@@ -482,8 +483,10 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
     }
     // Persist the canonical (normalized) verdict, not the raw caller value, so
     // the durable ledger never records a differently-cased/whitespaced verdict
-    // beside the canonical overallVerdict it was just checked against.
-    options.verdict = callerVerdict;
+    // beside the canonical overallVerdict it was just checked against. Kept in
+    // a local instead of mutating the caller-provided options object, which
+    // programmatic callers may reuse across calls.
+    persistedVerdict = callerVerdict;
   }
   let provenance;
   if (options.provenance === undefined) {
@@ -524,7 +527,7 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
     pr: options.pr,
     gate: options.gate,
     headSha: options.headSha,
-    verdict: options.verdict,
+    verdict: persistedVerdict,
     loggedAt: new Date().toISOString(),
     findings,
   };
--- a/test/contracts/gate-verdict-consistency-contract.test.mjs
+++ b/test/contracts/gate-verdict-consistency-contract.test.mjs
@@ -58,14 +58,15 @@ test("consolidate-fanin emits overallVerdict (the value the enforcement reads) i
 
 test("write-gate-findings-log rejects a --verdict contradicting the wrapper overallVerdict and cites GATE-COMMENT-VERDICT-VALUES", async () => {
   const src = await readFile(WRITE_GATE_FINDINGS_SCRIPT, "utf8");
-  // Producer-side refusal mirrors the consumer's: it cites the rule ID and the
-  // contract doc path so both sides of the seam point at the same rule.
-  assert.match(src, /GATE-COMMENT-VERDICT-VALUES/);
-  assert.match(src, /skills\/docs\/gate-review-comment-contract\.md/);
-  // The refusal names both the passed --verdict and the wrapper's overallVerdict,
-  // so a contradicting caller can identify the mismatch from the error alone.
-  assert.match(src, /--verdict/);
-  assert.match(src, /"overallVerdict"/);
+  // Distinctive refusal literal (not just the standalone tokens "--verdict" or
+  // "overallVerdict", which also match the CLI usage text and the unrelated
+  // #1616 malformed-wrapper message) — mirrors the upsert-side pin above: the
+  // producer-side refusal phrase must be followed, within a bounded window, by
+  // both the rule ID and the contract doc path, so the citation actually
+  // belongs to this refusal and not to an unrelated mention elsewhere in the
+  // file.
+  assert.match(src, /contradicts the wrapper's "overallVerdict"[\s\S]{0,160}GATE-COMMENT-VERDICT-VALUES/);
+  assert.match(src, /contradicts the wrapper's "overallVerdict"[\s\S]{0,160}skills\/docs\/gate-review-comment-contract\.md/);
 });
 
 test("upsert-checkpoint-verdict derives the verdict from the ledger's overallVerdict by default (#1616 AC: no --verdict is valid)", async () => {
--- a/test/contracts/gate-verdict-consistency-contract.test.mjs
+++ b/test/contracts/gate-verdict-consistency-contract.test.mjs
@@ -24,6 +24,7 @@ const REPO_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "..
 const CONTRACT_DOC = path.join(REPO_ROOT, "skills/docs/gate-review-comment-contract.md");
 const UPSERT_SCRIPT = path.join(REPO_ROOT, "scripts/github/upsert-checkpoint-verdict.mjs");
 const CONSOLIDATE_SCRIPT = path.join(REPO_ROOT, "scripts/loop/consolidate-fanin.mjs");
+const WRITE_GATE_FINDINGS_SCRIPT = path.join(REPO_ROOT, "scripts/github/write-gate-findings-log.mjs");
 
 test("GATE-COMMENT-VERDICT-VALUES states the clean/findings_present meanings the consolidator computes", async () => {
   const doc = await readFile(CONTRACT_DOC, "utf8");
@@ -53,6 +54,18 @@ test("consolidate-fanin emits overallVerdict (the value the enforcement reads) i
   // ...and embeds it in the --ledger-out wrapper that write-gate-findings-log
   // threads into the durable ledger upsert-checkpoint-verdict enforces against.
   assert.match(src, /JSON\.stringify\(\{ overallVerdict: consolidated\.verdict, findings \}/);
+});
+
+test("write-gate-findings-log rejects a --verdict contradicting the wrapper overallVerdict and cites GATE-COMMENT-VERDICT-VALUES", async () => {
+  const src = await readFile(WRITE_GATE_FINDINGS_SCRIPT, "utf8");
+  // Producer-side refusal mirrors the consumer's: it cites the rule ID and the
+  // contract doc path so both sides of the seam point at the same rule.
+  assert.match(src, /GATE-COMMENT-VERDICT-VALUES/);
+  assert.match(src, /skills\/docs\/gate-review-comment-contract\.md/);
+  // The refusal names both the passed --verdict and the wrapper's overallVerdict,
+  // so a contradicting caller can identify the mismatch from the error alone.
+  assert.match(src, /--verdict/);
+  assert.match(src, /"overallVerdict"/);
 });
 
 test("upsert-checkpoint-verdict derives the verdict from the ledger's overallVerdict by default (#1616 AC: no --verdict is valid)", async () => {
--- a/test/github/write-gate-findings-log.test.mjs
+++ b/test/github/write-gate-findings-log.test.mjs
@@ -1,10 +1,11 @@
 import assert from "node:assert/strict";
-import { mkdtemp, readdir, readFile, rm, writeFile } from "node:fs/promises";
+import { mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 import test from "node:test";
 
 import {
+  buildLogPath,
   checkProvenanceAngleCoverage,
   parseProvenanceJson,
   parseWriteGateFindingsLogCliArgs,
@@ -1392,9 +1393,15 @@ test("#1641: writeGateFindingsLog rejects a --verdict contradicting the wrapper 
         return true;
       },
     );
-    // No ledger must be written on the rejection.
-    const leaked = (await readdir(tmpDir)).filter((f) => f.endsWith(".json"));
-    assert.equal(leaked.length, 0, "a contradicting pair must not write a ledger before failing");
+    // No ledger must be written on the rejection. A non-recursive readdir of
+    // tmpDir would only ever see the "gate-findings" directory (buildLogPath
+    // nests the ledger three levels below tmpRoot), so it can never observe a
+    // leak; assert the actual write target is absent instead.
+    await assert.rejects(
+      () => readFile(buildLogPath({ repo: "o/n", pr: 7, gate: "draft_gate", headSha: "a1".repeat(20), tmpRoot: tmpDir }), "utf8"),
+      /ENOENT/,
+      "a contradicting pair must not write a ledger before failing",
+    );
 
     // Direction 2: --verdict findings_present vs wrapper clean
     await assert.rejects(
@@ -1413,6 +1420,51 @@ test("#1641: writeGateFindingsLog rejects a --verdict contradicting the wrapper 
         return true;
       },
     );
+
+    // Direction 3: --verdict blocked vs wrapper findings_present (issue's
+    // Domain bullet names this pair explicitly).
+    await assert.rejects(
+      () => writeGateFindingsLog({
+        repo: "o/n",
+        pr: 9,
+        gate: "draft_gate",
+        headSha: "c3".repeat(20),
+        verdict: "blocked",
+        findings: JSON.stringify({ overallVerdict: "findings_present", findings: [] }),
+        tmpRoot: tmpDir,
+      }),
+      (err) => {
+        assert.match(err.message, /--verdict "blocked"/);
+        assert.match(err.message, /"overallVerdict" "findings_present"/);
+        return true;
+      },
+    );
+  } finally {
+    await rm(tmpDir, { recursive: true, force: true });
+  }
+});
+
+test("#1641: writeGateFindingsLog rejects an invalid or non-string caller --verdict as a domain error, not a contradiction", async () => {
+  const tmpDir = await mkdtemp(path.join(os.tmpdir(), "gate-findings-ov-invalid-"));
+  try {
+    for (const badVerdict of ["bogus", undefined, ["clean"]]) {
+      await assert.rejects(
+        () => writeGateFindingsLog({
+          repo: "o/n",
+          pr: 7,
+          gate: "draft_gate",
+          headSha: "a1".repeat(20),
+          verdict: badVerdict,
+          findings: JSON.stringify({ overallVerdict: "clean", findings: [] }),
+          tmpRoot: tmpDir,
+        }),
+        (err) => {
+          assert.match(err.message, /--verdict must be clean, findings_present, or blocked/);
+          assert.doesNotMatch(err.message, /contradicts/);
+          return true;
+        },
+      );
+    }
   } finally {
     await rm(tmpDir, { recursive: true, force: true });
   }
@@ -1438,6 +1490,26 @@ test("#1641: writeGateFindingsLog accepts a matching --verdict + wrapper overall
   }
 });
 
+test("#1641: writeGateFindingsLog accepts a non-canonical caller --verdict and persists the canonical value", async () => {
+  const tmpDir = await mkdtemp(path.join(os.tmpdir(), "gate-findings-ov-noncanonical-"));
+  try {
+    const result = await writeGateFindingsLog({
+      repo: "o/n",
+      pr: 7,
+      gate: "draft_gate",
+      headSha: "a1".repeat(20),
+      verdict: " Clean ",
+      findings: JSON.stringify({ overallVerdict: "clean", findings: [] }),
+      tmpRoot: tmpDir,
+    });
+    const parsed = JSON.parse(await readFile(result.path, "utf8"));
+    assert.equal(parsed.verdict, "clean", "the ledger must persist the canonical verdict, not the raw caller casing/whitespace");
+    assert.equal(parsed.overallVerdict, "clean");
+  } finally {
+    await rm(tmpDir, { recursive: true, force: true });
+  }
+});
+
 test("#1641: CLI path — a direct write-gate-findings-log.mjs with a contradicting pair exits 1 with {ok:false,error} and writes no ledger", async () => {
   const tmpDir = await mkdtemp(path.join(os.tmpdir(), "gate-findings-ov-cli-conflict-"));
   try {
@@ -1452,14 +1524,19 @@ test("#1641: CLI path — a direct write-gate-findings-log.mjs with a contradict
       "--tmp-root", tmpDir,
     ], { cwd: tmpDir });
 
-    assert.notEqual(result.code, 0, "a contradicting pair must exit non-zero");
+    assert.equal(result.code, 1, "a contradicting pair must exit 1");
     assert.match(result.stderr, /\{"ok":false,"error"/);
     assert.match(result.stderr, /contradicts the wrapper/);
     assert.match(result.stderr, /GATE-COMMENT-VERDICT-VALUES/);
 
-    // No ledger must be written on the rejection.
-    const leaked = (await readdir(tmpDir)).filter((f) => f.endsWith(".json"));
-    assert.equal(leaked.length, 0, "a contradicting CLI pair must not write a ledger before failing");
+    // No ledger must be written on the rejection. Same non-recursive-readdir
+    // caveat as the programmatic-path test above: assert the actual write
+    // target is absent instead of scanning tmpDir's top level.
+    await assert.rejects(
+      () => readFile(buildLogPath({ repo: "owner/repo", pr: 42, gate: "draft_gate", headSha: "945391c0abcdef1234567890abcdef1234567890", tmpRoot: tmpDir }), "utf8"),
+      /ENOENT/,
+      "a contradicting CLI pair must not write a ledger before failing",
+    );
   } finally {
     await rm(tmpDir, { recursive: true, force: true });
   }
--- a/test/github/write-gate-findings-log.test.mjs
+++ b/test/github/write-gate-findings-log.test.mjs
@@ -10,6 +10,9 @@ import {
   parseWriteGateFindingsLogCliArgs,
   writeGateFindingsLog,
 } from "../../scripts/github/write-gate-findings-log.mjs";
+import { runNode as runNodeHelper } from "../_helpers.mjs";
+
+const writeGateFindingsLogScript = path.resolve("scripts/github/write-gate-findings-log.mjs");
 
 // #1592: several fixtures below deliberately keep pre-rename severity
 // spellings ("must-fix"/"worth-fixing-now"/"nice-to-have") as INPUT — this is
@@ -1430,6 +1433,33 @@ test("#1641: writeGateFindingsLog accepts a matching --verdict + wrapper overall
     const parsed = JSON.parse(await readFile(result.path, "utf8"));
     assert.equal(parsed.verdict, "findings_present");
     assert.equal(parsed.overallVerdict, "findings_present");
+  } finally {
+    await rm(tmpDir, { recursive: true, force: true });
+  }
+});
+
+test("#1641: CLI path — a direct write-gate-findings-log.mjs with a contradicting pair exits 1 with {ok:false,error} and writes no ledger", async () => {
+  const tmpDir = await mkdtemp(path.join(os.tmpdir(), "gate-findings-ov-cli-conflict-"));
+  try {
+    // Wrapper overallVerdict findings_present contradicts --verdict clean.
+    const result = await runNodeHelper(writeGateFindingsLogScript, [
+      "--repo", "owner/repo",
+      "--pr", "42",
+      "--gate", "draft_gate",
+      "--head-sha", "945391c0abcdef1234567890abcdef1234567890",
+      "--verdict", "clean",
+      "--findings", JSON.stringify({ overallVerdict: "findings_present", findings: [] }),
+      "--tmp-root", tmpDir,
+    ], { cwd: tmpDir });
+
+    assert.notEqual(result.code, 0, "a contradicting pair must exit non-zero");
+    assert.match(result.stderr, /\{"ok":false,"error"/);
+    assert.match(result.stderr, /contradicts the wrapper/);
+    assert.match(result.stderr, /GATE-COMMENT-VERDICT-VALUES/);
+
+    // No ledger must be written on the rejection.
+    const leaked = (await readdir(tmpDir)).filter((f) => f.endsWith(".json"));
+    assert.equal(leaked.length, 0, "a contradicting CLI pair must not write a ledger before failing");
   } finally {
     await rm(tmpDir, { recursive: true, force: true });
   }
--- a/test/github/write-gate-findings-log.test.mjs
+++ b/test/github/write-gate-findings-log.test.mjs
@@ -1,5 +1,5 @@
 import assert from "node:assert/strict";
-import { mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
+import { mkdtemp, readdir, readFile, rm, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 import test from "node:test";
@@ -1362,6 +1362,74 @@ test("#1616: writeGateFindingsLog persists overallVerdict from the {overallVerdi
     assert.equal(parsed.overallVerdict, "findings_present");
     assert.equal(parsed.findings.length, 1);
     assert.equal(parsed.findings[0].severity, "high"); // "must-fix" normalizes to canonical "high"
+  } finally {
+    await rm(tmpDir, { recursive: true, force: true });
+  }
+});
+
+test("#1641: writeGateFindingsLog rejects a --verdict contradicting the wrapper overallVerdict", async () => {
+  const tmpDir = await mkdtemp(path.join(os.tmpdir(), "gate-findings-ov-conflict-"));
+  try {
+    // Direction 1: --verdict clean vs wrapper findings_present
+    await assert.rejects(
+      () => writeGateFindingsLog({
+        repo: "o/n",
+        pr: 7,
+        gate: "draft_gate",
+        headSha: "a1".repeat(20),
+        verdict: "clean",
+        findings: JSON.stringify({ overallVerdict: "findings_present", findings: [] }),
+        tmpRoot: tmpDir,
+      }),
+      (err) => {
+        assert.match(err.message, /--verdict "clean"/);
+        assert.match(err.message, /"overallVerdict" "findings_present"/);
+        assert.match(err.message, /GATE-COMMENT-VERDICT-VALUES/);
+        assert.match(err.message, /skills\/docs\/gate-review-comment-contract\.md/);
+        return true;
+      },
+    );
+    // No ledger must be written on the rejection.
+    const leaked = (await readdir(tmpDir)).filter((f) => f.endsWith(".json"));
+    assert.equal(leaked.length, 0, "a contradicting pair must not write a ledger before failing");
+
+    // Direction 2: --verdict findings_present vs wrapper clean
+    await assert.rejects(
+      () => writeGateFindingsLog({
+        repo: "o/n",
+        pr: 8,
+        gate: "draft_gate",
+        headSha: "b2".repeat(20),
+        verdict: "findings_present",
+        findings: JSON.stringify({ overallVerdict: "clean", findings: [] }),
+        tmpRoot: tmpDir,
+      }),
+      (err) => {
+        assert.match(err.message, /--verdict "findings_present"/);
+        assert.match(err.message, /"overallVerdict" "clean"/);
+        return true;
+      },
+    );
+  } finally {
+    await rm(tmpDir, { recursive: true, force: true });
+  }
+});
+
+test("#1641: writeGateFindingsLog accepts a matching --verdict + wrapper overallVerdict", async () => {
+  const tmpDir = await mkdtemp(path.join(os.tmpdir(), "gate-findings-ov-match-"));
+  try {
+    const result = await writeGateFindingsLog({
+      repo: "o/n",
+      pr: 7,
+      gate: "draft_gate",
+      headSha: "a1".repeat(20),
+      verdict: "findings_present",
+      findings: JSON.stringify({ overallVerdict: "findings_present", findings: [] }),
+      tmpRoot: tmpDir,
+    });
+    const parsed = JSON.parse(await readFile(result.path, "utf8"));
+    assert.equal(parsed.verdict, "findings_present");
+    assert.equal(parsed.overallVerdict, "findings_present");
   } finally {
     await rm(tmpDir, { recursive: true, force: true });
   }
--- a/.claude/skills/docs/gate-review-comment-contract.md
+++ b/.claude/skills/docs/gate-review-comment-contract.md
@@ -127,16 +127,19 @@ with the fixed meaning below:
 | `findings_present` | The gate found issues at blocking severities; fixes are required before the gate boundary can be crossed |
 | `blocked` | The gate could not complete or a hard blocker prevented a verdict |
 
-This rule is enforced at post time, not just documented: `upsert-checkpoint-verdict.mjs`
-refuses a `--verdict` that contradicts the consolidated ledger's `overallVerdict`
-for the same head and gate (#1616). The consolidator (`consolidate-fanin.mjs`)
-already computes `overallVerdict` from this rule's definitions; it threads
-through `--ledger-out`'s `{ overallVerdict, findings }` wrapper into the durable
-ledger (`write-gate-findings-log.mjs`), and `upsert-checkpoint-verdict.mjs`
-reads it and derives the verdict by default (passing no `--verdict` is valid),
-accepts a matching explicit value, and refuses a contradiction citing this
-rule. No override flag — a round whose verdict genuinely differs from the
-computed one is a consolidator bug to fix, not an operator decision to override.
+This rule is enforced at write time and at post time, not just documented:
+`write-gate-findings-log.mjs` refuses a `--verdict` that contradicts the
+`--findings`/`--findings-file` wrapper's `overallVerdict` before any ledger is
+written, and `upsert-checkpoint-verdict.mjs` refuses a `--verdict` that
+contradicts the consolidated ledger's `overallVerdict` for the same head and
+gate (#1616). The consolidator (`consolidate-fanin.mjs`) already computes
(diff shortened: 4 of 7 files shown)


mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `a1e4e257428e328dd3e5a76847ff90908a22920a`
**Verdict:** findings_present

**Findings summary:** 8 angles reviewed; 23 findings (see per-angle breakdown below).

- `contract-surface` → `findings_present` (6 findings)
- `coverage` → `findings_present` (6 findings)

> **medium** (`contract-surface`): The documented write-seam surface still says write-gate-findings-log.mjs only threads the wrapper's overallVerdict into the ledger "so upsert-checkpoint-verdict.mjs enforces verdict consistency against it", and describes the refusal as living at the post seam. At the reviewed head the write seam itself fails closed (scripts/github/write-gate-findings-log.mjs:468-473): a caller following this documented invocation with a --verdict that disagrees with the --ledger-out wrapper now exits 1 with no ledger written, a fail-closed behavior no doc on this surface states.
> Location: `skills/docs/gate-review-sub-loop-contract.md:1314`


> **low** (`contract-surface`): GATE-COMMENT-VERDICT-VALUES states "This rule is enforced at post time, not just documented: upsert-checkpoint-verdict.mjs refuses a --verdict that contradicts ... overallVerdict". The new producer-side refusal cites this exact rule and doc path (scripts/github/write-gate-findings-log.mjs:471), so the rule's own enforcement-point list is now incomplete: enforcement is at write time as well as post time.
> Location: `skills/docs/gate-review-comment-contract.md:130`


> **low** (`contract-surface`): CLI --help drift: --findings &lt;json> is documented as "JSON array of finding objects", but the accepted input also includes the consolidator's { overallVerdict, findings } wrapper (resolveFindingsInput), and that wrapper now carries a new fail-closed rule (its overallVerdict must equal --verdict). Neither the accepted wrapper shape nor the equality requirement appears in USAGE, so an operator reading --help cannot predict the exit-1 path.
> Location: `scripts/github/write-gate-findings-log.mjs:24`


> **medium** (`correctness`): The new check compares a NORMALIZED caller verdict (line 468 normalizeVerdict(options.verdict)) but the ledger still persists the RAW options.verdict here. The diff's own comment says this widening exists so a programmatic caller may pass a "differently-cased/whitespaced but semantically equal" verdict; such a value now passes the equality check and is written verbatim, so the durable ledger can still record {"verdict": " Clean ", "overallVerdict": "clean"}. That is exactly the ambiguous verdict/overallVerdict pair the PR summary claims to close, and it also departs from the issue's proposed approach ("--verdict is already normalized at parse time, so both sides are canonical before comparison").
> Location: `scripts/github/write-gate-findings-log.mjs:514`


> **low** (`determinism`): Normalization is asymmetric between the accept check and the persisted value. Line 468 compares normalizeVerdict(options.verdict) (trim + lowercase, line 52-56), but line 514 persists the RAW options.verdict. For the programmatic caller the new comment at 463-467 explicitly says it supports, a verdict of "CLEAN" or " clean " passes the equality check and the ledger then records verdict: "CLEAN" next to overallVerdict: "clean" — a non-canonical, caller-casing-dependent durable value sitting beside the canonical one, which is the ambiguous pair this change exists to remove. Not reachable via the CLI (parse canonicalizes at line 353-357) and the consumer reads only overallVerdict (upsert-checkpoint-verdict.mjs:1793-1808), hence low. Related, same line 468: normalizeVerdict returns null for an omitted or invalid options.verdict, so a programmatic call without a verdict now throws --verdict undefined contradicts the wrapper's "overallVerdict" ..., which misattributes a missing-input error to a contradiction.
> Location: `scripts/github/write-gate-findings-log.mjs:514`


> **low** (`gate-evidence`): The recorded validation artifact for this head runs exactly one suite (npm run verify, exitCode 0), but the linked issue's definition of done also requires npm run assets:check and npm run schema:check, and the PR body asserts both green. At the reviewed head, package.json defines verify as run-p ... test:assets test:extension test:scripts test:core test:docs test:pack test:dev-loop test:workflows, which does NOT include assets:check (generate-claude-assets.mjs --check) or schema:check (generate-config-schema.mjs --check). Two DoD claims therefore have no gate-recorded evidence at head a1e4e25; they rest on an operator assertion in the PR body only. Impact is small because the diff touches no generated-asset or config-schema source, but the evidence record does not support the claim.


> **medium** (`pr-description`): The PR body is missing four sections that .github/pull_request_template.md (at the reviewed head) and the pr-description contract both require: "Scope / context", "Acceptance criteria", "Definition of done", and "Non-goals". The AC/DoD/Non-goals matrix is also absent. The body has Summary, Change, Tests, and Validation only, so the merge-readiness spine (what is in/out of bounds, what must be true to accept, what is deliberately not done) exists only in linked issue issue 1641 and is not stated in the PR contract itself. The scope angle has no in-PR scope statement to review against as a result.

scripts/github/write-gate-findings-log.mjs
```diff
@@ -457,6 +457,20 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
...
     normalizedOverallVerdict = verdict;
+    // Fail closed on a caller-passed --verdict that contradicts the
+    // consolidator's computed verdict, mirroring the consumer-side refusal in
+    // upsert-checkpoint-verdict.mjs (#1616, GATE-COMMENT-VERDICT-VALUES).
+    // Normalize options.verdict here so the comparison is symmetric for every
+    // caller — a direct programmatic caller passes a non-canonical verdict by
+    // the same normalizeVerdict path the CLI parse normally canonicalizes it
+    // through, so a canonical-value wrapper never false-rejects a
+    // differently-cased/whitespaced but semantically equal caller verdict.
+    const callerVerdict = normalizeVerdict(options.verdict);
```
mfittko: **low** (`contract-surface`): The comparison normalizes the caller verdict (callerVerdict = normalizeVerdict(options.verdict)) but the ledger still records the RAW value (verdict: options.verdict, line 514). A programmatic caller passing " Clean " with wrapper overallVerdict: "clean" now passes the equality check and persists verdict: " Clean " next to overallVerdict: "clean" — a ledger field outside the documented clean|findings_present|blocked domain (GATE-COMMENT-VERDICT-VALUES) whose two verdict fields disagree byte-wise, which is the exact ambiguity this PR removes. The CLI path is safe only because parse normalizes at line 353-357.

mfittko: **low** (`contract-surface`): Error-surface mismatch on the programmatic path: normalizeVerdict returns null for a missing or out-of-domain options.verdict, so null !== normalizedOverallVerdict and the caller gets --verdict undefined contradicts the wrapper's "overallVerdict" ... — a contradiction message for what is actually an invalid/absent verdict. The stderr {ok:false,error} text then names a cause that is not the failure.

test/github/write-gate-findings-log.test.mjs
```diff
@@ -1367,6 +1370,101 @@ test("#1616: writeGateFindingsLog persists overallVerdict from the {overallVerdi
...
+      (err) => {
+        assert.match(err.message, /--verdict "clean"/);
+        assert.match(err.message, /"overallVerdict" "findings_present"/);
+        assert.match(err.message, /GATE-COMMENT-VERDICT-VALUES/);
+        assert.match(err.message, /skills\/docs\/gate-review-comment-contract\.md/);
+        return true;
+      },
+    );
+    // No ledger must be written on the rejection.
+    const leaked = (await readdir(tmpDir)).filter((f) => f.endsWith(".json"));
```
mfittko: **low** (`contract-surface`): The "no ledger written" assertion cannot fail. readdir(tmpDir) is non-recursive and the ledger is written to &lt;tmpRoot>/gate-findings/&lt;repo-slug>/pr-&lt;N>/&lt;gate>-&lt;headSha>.json (buildLogPath), so the top level would only ever contain the gate-findings directory, never a .json entry. The same vacuous check is repeated in the CLI test at line 1461. The AC "no &lt;gate>-&lt;headSha>.json appears under --tmp-root" is therefore unpinned: the test stays green even if a ledger were written.

mfittko: **medium** (`correctness`): The "no ledger must be written" assertion is vacuous. buildLogPath writes to &lt;tmpRoot>/gate-findings/&lt;repo-slug>/pr-&lt;n>/&lt;gate>-&lt;headSha>.json, but this check is a non-recursive readdir(tmpDir) filtered on .json; the top level would only ever contain the directory entry gate-findings, which never ends in .json. The filter is empty whether or not a ledger was written, so AC 1 ("no &lt;gate>-&lt;headSha>.json appears under --tmp-root") is unpinned. Line 1461 in the CLI test has the identical defect, and direction 2 (line 1400) has no leak check at all.

test/github/write-gate-findings-log.test.mjs
```diff
@@ -1367,6 +1370,101 @@ test("#1616: writeGateFindingsLog persists overallVerdict from the {overallVerdi
...
+      "--repo", "owner/repo",
+      "--pr", "42",
+      "--gate", "draft_gate",
+      "--head-sha", "945391c0abcdef1234567890abcdef1234567890",
+      "--verdict", "clean",
+      "--findings", JSON.stringify({ overallVerdict: "findings_present", findings: [] }),
+      "--tmp-root", tmpDir,
+    ], { cwd: tmpDir });
+
+    assert.notEqual(result.code, 0, "a contradicting pair must exit non-zero");
```
mfittko: **low** (`correctness`): The CLI AC states the invocation "exits 1", but the test asserts assert.notEqual(result.code, 0), which also passes on exit code 2, a signal-kill (code === null), or any unrelated crash before the refusal is reached.

test/github/write-gate-findings-log.test.mjs
mfittko: **high** (`coverage`): The "no ledger written" assertion is vacuous and cannot fail. buildLogPath writes to &lt;tmpRoot>/gate-findings/&lt;repoSlug>/pr-&lt;n>/&lt;gate>-&lt;headSha>.json, so a non-recursive readdir(tmpDir) returns only a single "gate-findings" entry and the .endsWith(".json") filter is always empty. Verified empirically: a SUCCESSFUL writeGateFindingsLog into a fresh tmpRoot still yields leaked.length === 0. AC 1 ("no &lt;gate>-&lt;headSha>.json appears under --tmp-root") is therefore asserted but not verified; a regression that wrote the ledger before throwing would ship green.

test/github/write-gate-findings-log.test.mjs
```diff
@@ -1367,6 +1370,101 @@ test("#1616: writeGateFindingsLog persists overallVerdict from the {overallVerdi
...
+    assert.match(result.stderr, /\{"ok":false,"error"/);
+    assert.match(result.stderr, /contradicts the wrapper/);
+    assert.match(result.stderr, /GATE-COMMENT-VERDICT-VALUES/);
+
+    // No ledger must be written on the rejection.
+    const leaked = (await readdir(tmpDir)).filter((f) => f.endsWith(".json"));
```
mfittko: **medium** (`coverage`): Same vacuous no-ledger check on the CLI path: readdir(tmpDir) is non-recursive while the ledger lands two directories deeper, so this assertion also cannot fail. The CLI half of AC 1/AC 6 ("exits 1 ... and writes no ledger", per the test's own name) is unverified; only the exit code and stderr shape are actually pinned.

test/contracts/gate-verdict-consistency-contract.test.mjs
```diff
@@ -55,6 +56,18 @@ test("consolidate-fanin emits overallVerdict (the value the enforcement reads) i
...
 
+test("write-gate-findings-log rejects a --verdict contradicting the wrapper overallVerdict and cites GATE-COMMENT-VERDICT-VALUES", async () => {
+  const src = await readFile(WRITE_GATE_FINDINGS_SCRIPT, "utf8");
+  // Producer-side refusal mirrors the consumer's: it cites the rule ID and the
+  // contract doc path so both sides of the seam point at the same rule.
+  assert.match(src, /GATE-COMMENT-VERDICT-VALUES/);
+  assert.match(src, /skills\/docs\/gate-review-comment-contract\.md/);
+  // The refusal names both the passed --verdict and the wrapper's overallVerdict,
+  // so a contradicting caller can identify the mismatch from the error alone.
+  assert.match(src, /--verdict/);
```
mfittko: **medium** (`coverage`): The two "names both values" assertions are vacuous: /--verdict/ matches the USAGE text at write-gate-findings-log.mjs:23 and the parse error at :355 (4 matches already on origin/main), and /"overallVerdict"/ matches the pre-existing issue 1616 malformed-wrapper message at :456. Both pass unchanged on pre-change main, so they pin nothing about the new refusal and do not mirror the upsert-side test at line 47, which pins a distinctive literal ("contradicts the consolidated ledger's overallVerdict"). Relatedly, the /GATE-COMMENT-VERDICT-VALUES/ assertion at line 63 is satisfied by the new code COMMENT at write-gate-findings-log.mjs:462 alone, so deleting the citation from the thrown message keeps this test green despite the test name claiming "cites GATE-COMMENT-VERDICT-VALUES".

scripts/github/write-gate-findings-log.mjs
mfittko: **medium** (`coverage`): The caller-verdict normalization added here is entirely untested. The 8-line comment above it justifies normalizing options.verdict so a "differently-cased/whitespaced but semantically equal caller verdict" is not false-rejected, but no test in test/github/write-gate-findings-log.test.mjs ever passes a non-canonical programmatic verdict (grep finds only canonical "clean"/"findings_present"/"blocked" inputs). Replacing this line with a plain options.verdict !== normalizedOverallVerdict leaves every test green, so the documented behavior has no regression guard.

scripts/github/write-gate-findings-log.mjs
```diff
@@ -457,6 +457,20 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
...
+    if (callerVerdict !== normalizedOverallVerdict) {
+      throw parseError(
+        `--verdict ${JSON.stringify(options.verdict)} contradicts the wrapper's "overallVerdict" ${JSON.stringify(normalizedOverallVerdict)} (GATE-COMMENT-VERDICT-VALUES; skills/docs/gate-review-comment-contract.md)`,
```
mfittko: **low** (`coverage`): Untested error path: normalizeVerdict returns null for an invalid or missing programmatic verdict, so a direct caller passing e.g. verdict: "bogus" (or omitting it) alongside any wrapper falls into this branch and gets "--verdict \"bogus\" contradicts the wrapper's \"overallVerdict\" \"clean\"" — a contradiction message for what is really an invalid-domain value. Fail-closed, but the diagnostic misdescribes the fault and nothing pins the behavior, so a later refactor could silently turn it into a write.

test/github/write-gate-findings-log.test.mjs
```diff
@@ -1367,6 +1370,101 @@ test("#1616: writeGateFindingsLog persists overallVerdict from the {overallVerdi
   }
 });
 
+test("#1641: writeGateFindingsLog rejects a --verdict contradicting the wrapper overallVerdict", async () => {
```
mfittko, line 1374: **nit** (`coverage`): The issue's Domain bullet explicitly calls out "--verdict blocked against overallVerdict findings_present" as in-scope for the domain-agnostic check, but only the clean vs findings_present pair (both directions) is covered. The blocked value never appears in a contradiction case.

test/github/write-gate-findings-log.test.mjs
mfittko: **medium** (`determinism`): The "no ledger written" assertion cannot fail: the ledger is never written at the top level of --tmp-root. buildLogPath (scripts/github/write-gate-findings-log.mjs:418) returns path.join(tmpRoot, "gate-findings", repoSlug, pr-${pr}, ${gate}-${headSha}.json), so a written ledger lands three directories below tmpDir (the same nesting the existing tests pin at lines 191 and 224). A non-recursive readdir(tmpDir) returns at most a single "gate-findings" entry, and the .endsWith(".json") filter drops that, so leaked.length is 0 whether or not the refusal fired. The AC "no &lt;gate>-&lt;headSha>.json appears under --tmp-root" is therefore asserted by a check that is green by construction, and a regression that writes the ledger before throwing would pass this test.

test/github/write-gate-findings-log.test.mjs
mfittko: **medium** (`determinism`): Same vacuous leak check on the CLI path: readdir(tmpDir) is non-recursive while the CLI writes to &lt;tmpRoot>/gate-findings/owner-repo/pr-42/draft_gate-&lt;sha>.json (buildLogPath, scripts/github/write-gate-findings-log.mjs:418). leaked is always empty, so the "a contradicting CLI pair must not write a ledger before failing" assertion holds regardless of behavior. Fixing only line 1396 leaves this second copy vacuous.

scripts/github/write-gate-findings-log.mjs
mfittko: **medium** (`input-validation`): normalizeVerdict (line 52) returns null for any out-of-domain value, and the new code feeds that null straight into the inequality check, so an INVALID or ABSENT caller verdict is reported as a contradiction rather than as a domain error. A programmatic caller passing verdict: "bogus" gets --verdict "bogus" contradicts the wrapper's "overallVerdict" "clean", and one omitting verdict entirely gets --verdict undefined contradicts ... (JSON.stringify(undefined) interpolates as the bare word undefined). Both misname the actual fault. The same helper also coerces with String(value), so non-string inputs such as an array-wrapped "clean" or new String("CLEAN ") pass the new symmetry check and are then persisted raw at line 514 — a widening this diff introduces at a seam that previously only saw the CLI-canonicalized value.

test/contracts/gate-verdict-consistency-contract.test.mjs
mfittko: **low** (`input-validation`): Two of the four assertions in the new producer-side pin are vacuous against the whole-file source read. origin/main's write-gate-findings-log.mjs already contains 45 occurrences of --verdict (the usage block) and 1 of "overallVerdict" (the pre-existing issue 1616 malformed-wrapper error at line 456), so lines 67-68 pass without the refusal existing at all — yet the comment above them claims they pin that "the refusal names both the passed --verdict and the wrapper's overallVerdict". Only the two citation assertions (lines 65-66) actually fail on main and pin new behavior.

scripts/github/write-gate-findings-log.mjs
mfittko: **low** (`scope`): The second commit adds a local normalizeVerdict(options.verdict) that is outside the approach the linked issue states ("--verdict is already normalized at parse time, so both sides are canonical before comparison"). It silently widens accepted input: a programmatic caller passing a non-canonical verdict (e.g. " Clean ") now passes the equality check, but line 514 still persists the RAW options.verdict into the ledger, so the ledger can record a non-canonical verdict next to a canonical overallVerdict — the ambiguous pair this PR exists to prevent. It also changes behavior for a programmatic caller that omits verdict: normalizeVerdict(undefined) yields undefined, which trips the contradiction branch and reports a misleading "contradicts the wrapper" error instead of a missing-argument error.

mfittko: Fixed in 1b9618b8: the local normalization no longer widens accepted input — a typeof-string domain guard now rejects non-string and out-of-domain caller verdicts with the parser's own domain error before any comparison, and only the canonical value is persisted. Semantically-equal spellings (case/whitespace) are accepted per the normalization contract; everything else is refused.

mfittko: Fixed in 1b9618b8: after the contradiction check passes, the canonical (normalized) verdict is persisted instead of the raw options.verdict, closing the normalized-compare/raw-persist asymmetry. A regression test pins " Clean " being persisted as clean.

mfittko: Fixed in 1b9618b8: options.verdict is domain-validated before the contradiction check — null/undefined/non-string/out-of-domain values now throw the parser's own domain-error message (mirroring the CLI parse path) instead of a misattributed contradiction error.

test/contracts/gate-verdict-consistency-contract.test.mjs
mfittko: Fixed in 1b9618b8: the vacuous standalone regexes were replaced with windowed pins requiring the distinctive refusal phrase (contradicts the wrapper's overallVerdict) within 160 chars of GATE-COMMENT-VERDICT-VALUES and of the contract doc path — both fail on pre-change main.

scripts/github/write-gate-findings-log.mjs
mfittko: Fixed in 1b9618b8: a missing or out-of-domain programmatic verdict now surfaces the domain error (--verdict must be clean, findings_present, or blocked) rather than a contradiction message interpolating undefined. Covered by a new test asserting the message does not contain the word contradicts.

test/github/write-gate-findings-log.test.mjs
mfittko: Fixed in 1b9618b8: the CLI-path no-ledger assertion now checks the actual buildLogPath target is absent (readFile rejects with ENOENT) instead of a non-recursive readdir of the tmp root, so a regression that writes the ledger before refusing fails the test.

test/contracts/gate-verdict-consistency-contract.test.mjs
mfittko: Fixed in 1b9618b8: both pins were tightened to windowed regexes anchored on the distinctive refusal literal (contradicts the wrapper's overallVerdict) plus GATE-COMMENT-VERDICT-VALUES / the doc path, mirroring the upsert-side pin; they no longer match USAGE text or the pre-existing malformed-wrapper message.

test/github/write-gate-findings-log.test.mjs
mfittko: Fixed in 1b9618b8: the programmatic-path assertion now proves the buildLogPath target file is absent (ENOENT) after the refusal, replacing the non-recursive readdir check that could never fail.

scripts/github/write-gate-findings-log.mjs
mfittko: Fixed in 1b9618b8: added a test asserting a differently-cased/whitespaced caller verdict (" Clean ") is accepted against a clean wrapper and the canonical value is what lands in the persisted ledger.

scripts/github/write-gate-findings-log.mjs
mfittko: Fixed in 1b9618b8: added a test pinning the invalid/missing programmatic verdict path for a bogus string, undefined, and an array-wrapped value — each asserts the domain-error message and asserts the contradiction text is NOT used.

test/github/write-gate-findings-log.test.mjs
mfittko: Fixed in 1b9618b8: the no-ledger assertion now targets the buildLogPath location directly (readFile rejects with ENOENT), so it can actually fail if a regression writes the ledger before throwing.

test/github/write-gate-findings-log.test.mjs
mfittko: Fixed in 1b9618b8: the contradiction test now also covers the blocked vs findings_present direction named by the issue's Domain bullet.

test/github/write-gate-findings-log.test.mjs
mfittko: Fixed in 1b9618b8: the programmatic no-ledger assertion was replaced with an ENOENT check against the exact buildLogPath target, removing the vacuous top-level readdir.

test/github/write-gate-findings-log.test.mjs
mfittko: Fixed in 1b9618b8: the CLI refusal assertion was tightened from notEqual(code, 0) to equal(code, 1), matching the stated exit-1 contract.

test/github/write-gate-findings-log.test.mjs
mfittko: Fixed in 1b9618b8: the vacuous no-ledger assertion (non-recursive readdir that only ever sees the gate-findings directory) was replaced with an ENOENT assertion against the exact buildLogPath target on both the programmatic and CLI paths — verified to fail when the ledger write precedes the refusal.

test/github/write-gate-findings-log.test.mjs
mfittko: Fixed in 1b9618b8: the CLI-path leak check now asserts the buildLogPath target is absent (ENOENT) instead of scanning the tmp root non-recursively.

scripts/github/write-gate-findings-log.mjs
```diff
@@ -457,6 +457,33 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
...
+    // the same normalizeVerdict path the CLI parse normally canonicalizes it
+    // through, so a canonical-value wrapper never false-rejects a
+    // differently-cased/whitespaced but semantically equal caller verdict.
+    // Validate options.verdict's own domain first (same message the CLI parse
+    // path throws for --verdict) so a null/undefined/non-string/out-of-domain
+    // value is reported as an invalid verdict, not misattributed as a
+    // contradiction with the wrapper. typeof is checked before normalizeVerdict
+    // runs because normalizeVerdict coerces via String(), which would otherwise
+    // let a non-string (e.g. an array) silently pass as its stringified form.
+    const callerVerdict = typeof options.verdict === "string" ? normalizeVerdict(options.verdict) : null;
```
mfittko, line 475: **low** (`contract-surface`): The caller-verdict domain guard and the canonicalization that persists it live INSIDE the wrapper branch that starts at line 452, so they only run when the findings input carries overallVerdict. On the bare-array path the exported writeGateFindingsLog still records options.verdict raw: a programmatic caller passing a non-canonical value such as a padded capitalized clean persists that raw string as the ledger verdict, and a caller omitting verdict entirely writes a ledger with no verdict key at all, because JSON.stringify drops undefined. The durable ledger verdict field is consumed with a strict equality comparison in scripts/github/resolve-angle-carry-forward.mjs line 183, which then refuses carry-forward reporting the raw or null value, so the two write paths disagree about the documented verdict domain in GATE-COMMENT-VERDICT-VALUES.

scripts/github/write-gate-findings-log.mjs
```diff
@@ -457,6 +457,33 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
...
+    }
+    if (callerVerdict !== normalizedOverallVerdict) {
+      throw parseError(
+        `--verdict ${JSON.stringify(callerVerdict)} contradicts the wrapper's "overallVerdict" ${JSON.stringify(normalizedOverallVerdict)} (GATE-COMMENT-VERDICT-VALUES; skills/docs/gate-review-comment-contract.md)`,
+      );
+    }
+    // Persist the canonical (normalized) verdict, not the raw caller value, so
+    // the durable ledger never records a differently-cased/whitespaced verdict
+    // beside the canonical overallVerdict it was just checked against.
+    options.verdict = callerVerdict;
```
mfittko: **nit** (`correctness`): The canonicalization writes back into the caller-owned options object rather than a local, so writeGateFindingsLog now mutates its input argument as a side effect. Harmless today because the only non-test caller is main, which owns a freshly parsed local object, but the exported function is the programmatic API and a caller that reuses one options object across two writes gets its verdict silently rewritten by the first call.

test/github/write-gate-findings-log.test.mjs
```diff
@@ -1367,6 +1371,177 @@ test("#1616: writeGateFindingsLog persists overallVerdict from the {overallVerdi
...
+        assert.match(err.message, /"overallVerdict" "findings_present"/);
+        return true;
+      },
+    );
+  } finally {
+    await rm(tmpDir, { recursive: true, force: true });
+  }
+});
+
+test("#1641: writeGateFindingsLog rejects an invalid or non-string caller --verdict as a domain error, not a contradiction", async () => {
```
mfittko, line 1447: **medium** (`coverage`): The new caller-verdict domain check and canonicalization live inside the wrapper branch only, and every negative test exercises that branch. No test covers the no-wrapper path, where a caller verdict is still written raw and unvalidated. A bounded probe against this head confirms writeGateFindingsLog with a bare-array findings input and verdict " BOGUS " writes verdict " BOGUS " into the durable ledger, while the same value with a wrapper is rejected as a domain error. The suite therefore pins neither the guarded nor the unguarded behavior, so the asymmetry the change introduces is invisible to tests. The bare-array pin at line 1565 only asserts the absence of the overallVerdict key, so the acceptance criterion claiming the bare-array ledger writes byte-identically is under-evidenced on the verdict field.

test/github/write-gate-findings-log.test.mjs
```diff
@@ -1367,6 +1371,177 @@ test("#1616: writeGateFindingsLog persists overallVerdict from the {overallVerdi
...
+          return true;
+        },
+      );
+    }
+  } finally {
+    await rm(tmpDir, { recursive: true, force: true });
+  }
+});
+
+test("#1641: writeGateFindingsLog accepts a matching --verdict + wrapper overallVerdict", async () => {
```
mfittko, line 1473: **medium** (`coverage`): The combination of a wrapper carrying overallVerdict and --judge-verdict is untested, yet it is the documented Phase 3 flow: skills/dev-loop/SKILL.md line 185 writes the durable ledger with write-gate-findings-log --judge-verdict, and the sub-loop contract feeds the consolidator's ledger-out wrapper into --findings-file. judge-pass passes overallVerdict through unchanged at scripts/loop/judge-pass.mjs line 312 even when the judge rejects every finding, so an orchestrator whose verdict reflects post-judge dispositions now hits the new refusal. A bounded probe at this head confirms it: wrapper overallVerdict findings_present plus judge dispositions rejecting the only finding plus verdict clean throws the contradiction error. The existing judge test at line 1587 uses a bare array, so nothing in the suite covers this interaction.

test/github/write-gate-findings-log.test.mjs
```diff
@@ -1367,6 +1371,177 @@ test("#1616: writeGateFindingsLog persists overallVerdict from the {overallVerdi
   }
 });
 
+test("#1641: writeGateFindingsLog rejects a --verdict contradicting the wrapper overallVerdict", async () => {
```
mfittko, line 1374: **low** (`coverage`): Every wrapper test, both the new ones for issue 1641 and the pre-existing one, supplies the wrapper as an inline findings string. No test supplies a wrapper through findingsFile, which is the form the documented production flow uses (consolidate-fanin --ledger-out written to disk, then passed as --findings-file). The unwrap seam in scripts/github/_findings-input.mjs is shared between both forms, so divergence is unlikely, but the shipped path itself is never exercised end to end against the new refusal.

scripts/github/write-gate-findings-log.mjs
mfittko: **low** (`determinism`): The new canonicalization of the caller verdict happens only inside the wrapper branch and mutates the caller's options object, so the durable ledger's verdict field is now branch-dependent: with a wrapper carrying overallVerdict a programmatic verdict of ' Clean ' persists as 'clean', while the bare-array path persists the raw ' Clean ' unchanged at line 527. Same caller input, different persisted value, decided by an unrelated field. The strict reader at resolve-angle-carry-forward.mjs:183 compares log.verdict !== 'clean' with no normalization, so a raw value there fails closed on a semantically clean round. The in-place write to options.verdict also leaks across calls when a caller reuses one options object for several gates. Not reachable from production today because main() is the only non-test caller and the CLI parse already canonicalizes at line 354, but the divergence is introduced by this diff.

scripts/github/write-gate-findings-log.mjs
mfittko, line 475: **low** (`input-validation`): The new caller-verdict domain guard and canonicalization sit INSIDE the wrapper-present branch, so writeGateFindingsLog validates and normalizes options.verdict only when the findings payload carries overallVerdict. A bare-array payload, the legacy shape the PR preserves, still reaches the ledger write with options.verdict entirely unchecked: a programmatic caller passing a padded or mixed-case value persists it raw, and passing undefined persists a ledger with no verdict key at all. The guard is also dead for the CLI, whose parse layer already normalizes, so as written it covers neither entrypoint consistently. The read seam in scripts/github/_gate-finding-surface.mjs validates the ledger verdict with strict Set membership and no trimming, so such a ledger fails closed only later, at its consumer, after being written.

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `32163d258072e07ce21317e5935fe4363804d62b`
**Verdict:** clean

**Findings summary:** 9 angles reviewed; 9 findings (see per-angle breakdown below).

- `gate-evidence` → `clean`
- `link-check` → `clean`
- `scope` → `clean`

> **low** (`input-validation`): The wrapper value itself is still normalized with a bare normalizeVerdict call, which coerces through String, so a malformed overallVerdict that is a one-element JSON array holding the string clean stringifies to clean and is accepted as canonical. That is the exact coercion hazard the round-1 fix closed on the caller side with a typeof check two lines below, left open on the wrapper side, and it contradicts the comment just above which claims a malformed wrapper cannot record an invalid verdict. The consumer at scripts/github/_gate-finding-surface.mjs uses strict Set membership with no coercion, so producer and consumer disagree about what a valid overallVerdict is.
> Location: `scripts/github/write-gate-findings-log.mjs:453`


> **medium** (`pr-description`): The PR body's Change and Tests sections no longer describe the delivered implementation, so the body contract misstates the behavior it governs. Change still asserts that both sides are canonical at the comparison because --verdict is normalized at parse time; at the reviewed head the function re-normalizes options.verdict locally behind a typeof guard, refuses a non-string or out-of-domain caller verdict with its own domain error, and rewrites options.verdict so the ledger persists the canonical value. The refusal message also names the canonical caller verdict rather than the raw passed one, which the body does not say. Tests lists two additions while the head adds six, including the caller-verdict domain error, the canonical-persistence case, the blocked-versus-findings_present direction, and the direct CLI exit-1 path. The drift is new since the round-1 reviewed head a1e4e257: commit 1b9618b8 added the domain validation and the canonical persistence and the body was not updated. This is distinct from the deferred missing-sections finding, which concerns absent sections rather than inaccurate present ones.

**Next action:** Mark the PR ready for review (draft gate clean; judge deferred all 9 residual findings to follow-ups)

scripts/github/write-gate-findings-log.mjs
mfittko: Deferred at gate close (round 3, fingerprint 59c9a69943882350, severity low, angle contract-surface): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/write-gate-findings-log.mjs
mfittko: Deferred at gate close (round 3, fingerprint 7b43ea46153748d4, severity nit, angle correctness): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

test/github/write-gate-findings-log.test.mjs
mfittko: Deferred at gate close (round 3, fingerprint 928954d9fb38e9d3, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/write-gate-findings-log.mjs
mfittko: Deferred at gate close (round 3, fingerprint b9696a0404f505ea, severity low, angle determinism): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/write-gate-findings-log.mjs
mfittko: Deferred at gate close (round 3, fingerprint ef0bab27e0fc00c9, severity low, angle input-validation): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

test/github/write-gate-findings-log.test.mjs
mfittko: Judge disposition: deferred. Unifying the domain guard across the wrapper and bare-array paths (and dropping the options mutation) is tracked as a follow-up in issue 1744, milestoned into the current rc; fixing it inside this PR would break the byte-identical-preservation acceptance criterion.

test/github/write-gate-findings-log.test.mjs
mfittko: Judge disposition: deferred. The wrapper-plus-judge-verdict interaction needs a contract decision that is out of this producer guard's scope (declared non-goal); tracked as issue 1745, milestoned into the current rc.

scripts/github/write-gate-findings-log.mjs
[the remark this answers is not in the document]
mfittko: Fixed in f6358ba3: the canonical verdict now lives in a local (persistedVerdict) used at the write site; the caller-provided options object is no longer mutated.

scripts/github/write-gate-findings-log.mjs
```diff
@@ -457,6 +458,35 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
...
+    // contradiction with the wrapper. typeof is checked before normalizeVerdict
+    // runs because normalizeVerdict coerces via String(), which would otherwise
+    // let a non-string (e.g. an array) silently pass as its stringified form.
+    const callerVerdict = typeof options.verdict === "string" ? normalizeVerdict(options.verdict) : null;
+    if (!callerVerdict) {
+      throw parseError("--verdict must be clean, findings_present, or blocked");
+    }
+    if (callerVerdict !== normalizedOverallVerdict) {
+      throw parseError(
+        `--verdict ${JSON.stringify(callerVerdict)} contradicts the wrapper's "overallVerdict" ${JSON.stringify(normalizedOverallVerdict)} (GATE-COMMENT-VERDICT-VALUES; skills/docs/gate-review-comment-contract.md)`,
```
mfittko, line 481: **nit** (`contradiction-lens`): The comment directly above says the refusal mirrors the consumer-side refusal in upsert-checkpoint-verdict.mjs, but the two messages do not share a shape. The consumer at upsert-checkpoint-verdict.mjs line 1807 renders the gate and the canonical head sha inside the sentence and leaves the key name unquoted, reading as contradicts the consolidated ledger overallVerdict followed by the quoted value. The producer omits gate and head sha entirely, even though options.gate and options.headSha are in hand, and double-quotes the key name so the rendered text places a quoted key immediately beside a quoted value. Both name both values and both cite the rule and the doc path, so enforcement semantics agree; only the surface text diverges from the mirror the comment claims.

scripts/github/write-gate-findings-log.mjs
```diff
@@ -457,6 +458,35 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
...
+    // differently-cased/whitespaced but semantically equal caller verdict.
+    // Validate options.verdict's own domain first (same message the CLI parse
+    // path throws for --verdict) so a null/undefined/non-string/out-of-domain
+    // value is reported as an invalid verdict, not misattributed as a
+    // contradiction with the wrapper. typeof is checked before normalizeVerdict
+    // runs because normalizeVerdict coerces via String(), which would otherwise
+    // let a non-string (e.g. an array) silently pass as its stringified form.
+    const callerVerdict = typeof options.verdict === "string" ? normalizeVerdict(options.verdict) : null;
+    if (!callerVerdict) {
+      throw parseError("--verdict must be clean, findings_present, or blocked");
```
mfittko, line 477: **nit** (`dry`): The verdict domain-error string is now duplicated verbatim in two places in this file: the CLI parser throws it at line 355 and the new caller-verdict domain check throws the same literal at line 477. The two must stay byte-identical because both the producer suite and the CLI-path test assert on that wording, so editing one message silently diverges the other. The same file already imports FULL_HEAD_SHA_ERROR from lib/head-sha.mjs for exactly this reason, so the hoisted-constant pattern for a repeated CLI domain error already exists here and the new site did not use it.

scripts/github/write-gate-findings-log.mjs
```diff
@@ -457,6 +458,35 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
...
+    // the same normalizeVerdict path the CLI parse normally canonicalizes it
+    // through, so a canonical-value wrapper never false-rejects a
+    // differently-cased/whitespaced but semantically equal caller verdict.
+    // Validate options.verdict's own domain first (same message the CLI parse
+    // path throws for --verdict) so a null/undefined/non-string/out-of-domain
+    // value is reported as an invalid verdict, not misattributed as a
+    // contradiction with the wrapper. typeof is checked before normalizeVerdict
+    // runs because normalizeVerdict coerces via String(), which would otherwise
+    // let a non-string (e.g. an array) silently pass as its stringified form.
+    const callerVerdict = typeof options.verdict === "string" ? normalizeVerdict(options.verdict) : null;
```
mfittko, line 475: **nit** (`yagni`): The caller-verdict typeof guard, the re-normalization, and the persistedVerdict canonicalization only ever fire for a programmatic caller of writeGateFindingsLog. Every in-repo importer of this module pulls only buildLogPath or VALID_DISPOSITIONS (upsert-checkpoint-verdict.mjs, detect-checkpoint-evidence.mjs, resolve-angle-carry-forward.mjs, _gate-finding-surface.mjs), so the sole live entry point is the CLI, whose parse path at line 355 already validated and canonicalized options.verdict. None of the acceptance criteria for issue 1641 ask for caller-domain validation or for persisting a re-canonicalized verdict; the criteria only require the contradiction refusal, the malformed-wrapper ordering, and byte-identical matching and bare-array writes. The canonicalization also drifts slightly from the criterion that a matching pair writes identically to current behavior, since a non-canonical programmatic verdict such as a padded, capitalized value now lands in the ledger as the canonical form rather than as passed.

mfittko (commented):


**Reviewed head SHA:** `48af14d5f721c3f2f3ebdafa72a92463a17c27e4`
**Verdict:** clean

**Findings summary:** 7 angles reviewed; 3 findings (see per-angle breakdown below).

- `docs` → `clean`

**Next action:** Stop at the human approval checkpoint; merge only on explicit operator authorization

scripts/github/write-gate-findings-log.mjs
mfittko: Deferred at gate close (round 2, fingerprint a4ec0c6240f48384, severity nit, angle contradiction-lens): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/write-gate-findings-log.mjs
mfittko: Deferred at gate close (round 2, fingerprint d9a9ed0a73cdf1b4, severity nit, angle dry): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/write-gate-findings-log.mjs
mfittko: Deferred at gate close (round 2, fingerprint 575511a065f14ad9, severity nit, angle yagni): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

PR #1749, opened 2026-08-18
fix(hook): SubagentStop guard survives >1MB porcelain — 10MB maxBuffer + capped reason (#1686)

### LINKED ISSUE #1686: Raise execFileSync maxBuffer in SubagentStop guard (git status --porcelain >1MB defeats the guard) (#1619 follow-up)
## Summary

Follow-up from the Copilot review on PR #1682 (cf1a20d6): the SubagentStop uncommitted-work guard's `git status --porcelain` call can exceed Node's default `execFileSync` `maxBuffer` (1MB). In that case `execFileSync` throws, the catch sets `porcelain=""` (fail-safe allow), and the guard is defeated — a dirty worktree with >1MB of status output (~7000+ uncommitted files) would be allowed to stop, contradicting the guard's data-loss-prevention purpose.

## Acceptance criteria

- [x] Raise the `execFileSync` `maxBuffer` for the `git status --porcelain` call in `.claude/hooks/subagent-stop-uncommitted-guard.mjs` (e.g. 10MB) so a large-but-realistic dirty worktree doesn't trip the limit and fall back to allow.
- [x] A test exercises a porcelain output exceeding the default 1MB maxBuffer and asserts the guard still blocks (or documents the ceiling).

## Non-goals

- Changing the fail-safe-allow behavior on genuine git errors / timeouts (that is correct).

## Context

- PR #1682 (#1619) introduced the guard with a 5s timeout + fail-safe allow. The maxBuffer ceiling was not raised.
- Deferred at the pre_approval_gate (Copilot thread resolved as non-blocking; the common-case guard behavior is correct).

mfittko:
## Summary

The SubagentStop uncommitted-work guard's `git status --porcelain` call could exceed Node's default 1MB `execFileSync` maxBuffer, throw, and fall into the fail-safe-allow catch — defeating the guard on exactly the worktrees with the most uncommitted work. The call now runs with a 10MB maxBuffer, and the block reason caps its dirty-path enumeration at 50 paths plus a full-count summary so the output stays consumable by the harness and the stopping subagent.

## Scope and context

Six files: the hook (`.claude/hooks/subagent-stop-uncommitted-guard.mjs`, hand-written), the shared decision function in core (`packages/core/src/claude/hook-decisions.mjs`) with its regenerated vendored copy (`.claude/hooks/_hook-decisions.mjs`), one test file per layer, and a CHANGELOG entry. The reason-cap is a companion fix discovered by the new e2e test: an unbounded path list on a 4600-file worktree produced a >1MB stderr reason that consumers truncate or choke on.

## Acceptance criteria

- [x] The `git status --porcelain` call in the SubagentStop guard runs with a 10MB maxBuffer; a large-but-realistic dirty worktree no longer trips the default limit into fail-safe allow.
- [x] A test exercises porcelain output exceeding the 1MB default and asserts the guard still blocks; the 10MB ceiling is documented in the hook comment.
- [x] The block reason lists at most 50 dirty paths plus a full-count summary line, pinned by unit tests at the 50/51 boundary (order-preserving) and by the e2e truncation-summary assertion.

## Definition of done

- [x] `node --test test/contracts/claude-hooks-settings.test.mjs` green (24 tests, incl. the new >1MB e2e case).
- [x] `node --test packages/core/test/claude-hook-decisions.test.mjs` green (73 tests, incl. the path-cap unit case and the 50/51 boundary cases).
- [x] `npm run assets:check` green (vendored hook copy regenerated).
- [x] `npm run verify` green.

|---|---|---|
| 10MB maxBuffer on the porcelain call | claude-hooks-settings e2e suite green | fail-safe allow on genuine git errors unchanged |
| >1MB porcelain still blocks, ceiling documented | new e2e test (4600-file fixture, >1MB verified) plus the hook comment | no behavior change beyond the buffer and the reason cap |
| Reason stays consumable (path list capped at 50 + full count) | new unit test in claude-hook-decisions suite | — |

## Non-goals

- Changing the fail-safe-allow behavior on genuine git errors / timeouts (correct as-is; pinned by the existing non-git e2e test).

## Open questions

None. The one open design point surfaced in review — whether the reason should also carry a byte-level cap in addition to the 50-path cap — was judged a follow-up, not a blocker (the 50-path cap bounds the realistic case; a byte cap would only matter for pathological single-path lengths).

## Validation

```sh
node --test test/contracts/claude-hooks-settings.test.mjs
node --test packages/core/test/claude-hook-decisions.test.mjs
npm run assets:check && npm run verify
```

Closes #1686

### DIFF
--- a/.claude/hooks/subagent-stop-uncommitted-guard.mjs
+++ b/.claude/hooks/subagent-stop-uncommitted-guard.mjs
@@ -6,9 +6,10 @@
  * uncommitted changes in a worktree are destroyed with no warning — the only data-loss gap in
  * the enforcement audit. `LOCAL-COMMIT-BEFORE-EXIT` existed only as prose; this hook makes it
  * mechanical: when a subagent stops with a dirty worktree under `tmp/worktrees/`, refuse the
- * stop (exit code 2 + stderr JSON naming `LOCAL-COMMIT-BEFORE-EXIT` and the dirty paths) so the
- * subagent commits before exiting. Post-merge cleanup that force-removes worktrees can no
- * longer destroy uncommitted work silently.
+ * stop (exit code 2 + stderr JSON naming `LOCAL-COMMIT-BEFORE-EXIT` and up to 50 dirty paths,
+ * plus a summary line with the full count when there are more) so the subagent commits before
+ * exiting. Post-merge cleanup that force-removes worktrees can no longer destroy uncommitted
+ * work silently.
  *
  * Interactive sessions awaiting commit authorization set `DEVLOOPS_COMMIT_AUTH_PENDING=1` (an
  * opt-in operator/coordination-path signal, like `DEVLOOPS_MAIN_AGENT_READONLY`) and are
@@ -35,10 +36,13 @@ try {
   // Bounded: a stalled `git status` (slow/networked FS, held index lock) must not hang the
   // subagent stop. On timeout git is killed and execFileSync throws → caught → porcelain=""
   // → fail-safe allow (better than blocking the exit with no message).
-  // maxBuffer is raised to 10MB: the Node default (1MB) makes execFileSync throw on a
-  // dirty worktree with ~4000+ status lines, which the catch would turn into a fail-safe
-  // allow — defeating the guard exactly when the most work is at risk. Beyond 10MB
-  // (~40k+ dirty paths) the fail-safe allow is the documented ceiling.
+  // maxBuffer is raised to 10MB: the Node default (1MB) makes execFileSync throw once
+  // `git status --porcelain` output crosses that size, which the catch would turn into a
+  // fail-safe allow — defeating the guard exactly when the most work is at risk. The e2e
+  // fixture (test/contracts/claude-hooks-settings.test.mjs) trips the 1MB default at
+  // ~4300+ dirty paths using its long (~245-byte) porcelain lines; real paths are much
+  // shorter, so a real worktree needs far more entries to hit either bound. Beyond 10MB
+  // (~43k+ such long-line paths) the fail-safe allow is the documented ceiling.
   porcelain = execFileSync("git", ["status", "--porcelain"], { cwd, encoding: "utf8", timeout: 5000, maxBuffer: 10 * 1024 * 1024 });
 } catch {
   // Not a git repo / git unavailable / status timed out — nothing to guard, allow the stop.
--- a/.claude/hooks/subagent-stop-uncommitted-guard.mjs
+++ b/.claude/hooks/subagent-stop-uncommitted-guard.mjs
@@ -34,8 +34,12 @@ let porcelain = "";
 try {
   // Bounded: a stalled `git status` (slow/networked FS, held index lock) must not hang the
   // subagent stop. On timeout git is killed and execFileSync throws → caught → porcelain=""
-  // → fail-safe allow (better than blocking the exit with no message). #1619 review finding.
-  porcelain = execFileSync("git", ["status", "--porcelain"], { cwd, encoding: "utf8", timeout: 5000 });
+  // → fail-safe allow (better than blocking the exit with no message).
+  // maxBuffer is raised to 10MB: the Node default (1MB) makes execFileSync throw on a
+  // dirty worktree with ~4000+ status lines, which the catch would turn into a fail-safe
+  // allow — defeating the guard exactly when the most work is at risk. Beyond 10MB
+  // (~40k+ dirty paths) the fail-safe allow is the documented ceiling.
+  porcelain = execFileSync("git", ["status", "--porcelain"], { cwd, encoding: "utf8", timeout: 5000, maxBuffer: 10 * 1024 * 1024 });
 } catch {
   // Not a git repo / git unavailable / status timed out — nothing to guard, allow the stop.
   porcelain = "";
--- a/packages/core/src/claude/hook-decisions.mjs
+++ b/packages/core/src/claude/hook-decisions.mjs
@@ -422,6 +422,15 @@ export function decideSubagentStopGuard({ cwd, porcelain, pendingCommitAuthoriza
     .split("\n")
     .map((l) => l.trim())
     .filter(Boolean);
+  // The reason is fed back to the stopping subagent as context, so the path
+  // enumeration is capped: an unbounded list on a very dirty worktree produces a
+  // multi-megabyte reason that consumers truncate or choke on, burying the one
+  // actionable line. The count always reports the full total.
+  const MAX_LISTED_DIRTY_PATHS = 50;
+  const listed = dirty.slice(0, MAX_LISTED_DIRTY_PATHS).map((p) => "  " + p);
+  if (dirty.length > MAX_LISTED_DIRTY_PATHS) {
+    listed.push(`  … and ${dirty.length - MAX_LISTED_DIRTY_PATHS} more (run \`git status --porcelain\` for the full list)`);
+  }
   return {
     decision: "block",
     reason:
@@ -429,6 +438,6 @@ export function decideSubagentStopGuard({ cwd, porcelain, pendingCommitAuthoriza
       "to prevent silent data loss from post-merge worktree cleanup (cleanup-worktree.mjs runs " +
       "`git worktree remove --force`). Commit your work before stopping. " +
       `Dirty paths (${dirty.length}):\n` +
-      dirty.map((p) => "  " + p).join("\n"),
+      listed.join("\n"),
   };
 }
--- a/packages/core/test/claude-hook-decisions.test.mjs
+++ b/packages/core/test/claude-hook-decisions.test.mjs
@@ -541,6 +541,16 @@ test("decideSubagentStopGuard treats a non-string cwd as out-of-scope (allow)", 
   assert.equal(decideSubagentStopGuard({ cwd: undefined, porcelain: " M x" }).decision, "allow");
 });
 
+test("decideSubagentStopGuard caps the enumerated dirty paths at 50 and reports the full count", () => {
+  const lines = Array.from({ length: 120 }, (_, i) => `?? file-${String(i).padStart(3, "0")}.txt`);
+  const d = decideSubagentStopGuard({ cwd: WT, porcelain: lines.join("\n") });
+  assert.equal(d.decision, "block");
+  assert.match(d.reason, /Dirty paths \(120\):/);
+  assert.match(d.reason, /file-049\.txt/);
+  assert.doesNotMatch(d.reason, /file-050\.txt/);
+  assert.match(d.reason, /… and 70 more/);
+});
+
 // ---------------------------------------------------------------------------
 // decideBashGate — the six guard rules enforced at one seam (#1622)
 // ---------------------------------------------------------------------------
--- a/packages/core/test/claude-hook-decisions.test.mjs
+++ b/packages/core/test/claude-hook-decisions.test.mjs
@@ -541,14 +541,40 @@ test("decideSubagentStopGuard treats a non-string cwd as out-of-scope (allow)", 
   assert.equal(decideSubagentStopGuard({ cwd: undefined, porcelain: " M x" }).decision, "allow");
 });
 
-test("decideSubagentStopGuard caps the enumerated dirty paths at 50 and reports the full count", () => {
-  const lines = Array.from({ length: 120 }, (_, i) => `?? file-${String(i).padStart(3, "0")}.txt`);
+test("decideSubagentStopGuard caps the enumerated dirty paths at 50, reports the full count, and preserves porcelain order", () => {
+  // Descending index order (not ascending/sorted) so the assertions can tell "preserves
+  // porcelain order" apart from "sorts the paths" — git porcelain lists tracked
+  // staged/modified paths before untracked ones, and a sort creeping into the decider would
+  // displace the higher-data-loss-risk entries out of the first 50 shown.
+  const lines = Array.from({ length: 120 }, (_, i) => `?? file-${String(119 - i).padStart(3, "0")}.txt`);
   const d = decideSubagentStopGuard({ cwd: WT, porcelain: lines.join("\n") });
   assert.equal(d.decision, "block");
   assert.match(d.reason, /Dirty paths \(120\):/);
+  assert.match(d.reason, /file-119\.txt/);
+  assert.match(d.reason, /file-070\.txt/);
+  assert.doesNotMatch(d.reason, /file-069\.txt/);
+  assert.match(d.reason, /… and 70 more/);
+  const firstListedPath = d.reason.split("\n")[1].trim();
+  assert.equal(firstListedPath, lines[0], "first listed path must be the first porcelain line verbatim (order preserved, not sorted)");
+});
+
+test("decideSubagentStopGuard lists all paths with no truncation summary at exactly the 50-path cap", () => {
+  const lines = Array.from({ length: 50 }, (_, i) => `?? file-${String(i).padStart(3, "0")}.txt`);
+  const d = decideSubagentStopGuard({ cwd: WT, porcelain: lines.join("\n") });
+  assert.equal(d.decision, "block");
+  assert.match(d.reason, /Dirty paths \(50\):/);
+  assert.match(d.reason, /file-049\.txt/);
+  assert.doesNotMatch(d.reason, /… and/);
+});
+
+test("decideSubagentStopGuard adds a one-more truncation summary at 51 dirty paths (smallest over-cap input)", () => {
+  const lines = Array.from({ length: 51 }, (_, i) => `?? file-${String(i).padStart(3, "0")}.txt`);
+  const d = decideSubagentStopGuard({ cwd: WT, porcelain: lines.join("\n") });
+  assert.equal(d.decision, "block");
+  assert.match(d.reason, /Dirty paths \(51\):/);
   assert.match(d.reason, /file-049\.txt/);
   assert.doesNotMatch(d.reason, /file-050\.txt/);
-  assert.match(d.reason, /… and 70 more/);
+  assert.match(d.reason, /… and 1 more/);
 });
 
 // ---------------------------------------------------------------------------
--- a/test/contracts/claude-hooks-settings.test.mjs
+++ b/test/contracts/claude-hooks-settings.test.mjs
@@ -363,3 +363,29 @@ test("SubagentStop hook fail-safe-allows a mkdir-only (non-git) dir under tmp/wo
     fs.rmSync(dir, { recursive: true, force: true });
   }
 });
+
+test("SubagentStop hook still blocks when porcelain output exceeds the 1MB Node default maxBuffer (#1686)", () => {
+  // With Node's default execFileSync maxBuffer (1MB), ~4000+ dirty paths make
+  // `git status --porcelain` throw ERR_CHILD_PROCESS_STDIO_MAXBUFFER, and the catch
+  // turns that into a fail-safe allow — defeating the guard on exactly the worktrees
+  // with the most uncommitted work. The hook raises maxBuffer to 10MB; this test
+  // creates enough long-named untracked files to push porcelain past 1MB and asserts
+  // the guard still blocks. Beyond 10MB the fail-safe allow is the documented ceiling.
+  const dir = makeWorktree("maxbuffer", false);
+  try {
+    const name = (i) => `f${String(i).padStart(5, "0")}-${"x".repeat(230)}.txt`;
+    // ~4600 files x ~240-byte porcelain lines ≈ 1.1MB of status output.
+    for (let i = 0; i < 4600; i++) {
+      fs.writeFileSync(path.join(dir, name(i)), "", "utf8");
+    }
+    const porcelainBytes = spawnSync("git", ["status", "--porcelain"], { cwd: dir, encoding: "utf8", maxBuffer: 16 * 1024 * 1024 }).stdout.length;
+    assert.ok(porcelainBytes > 1024 * 1024, `fixture must exceed the 1MB default maxBuffer (got ${porcelainBytes} bytes)`);
+    const { code, stderrJson } = runHook("subagent-stop-uncommitted-guard.mjs", { cwd: dir });
+    assert.equal(code, 2, "a >1MB-porcelain dirty worktree must still be refused (exit 2)");
+    assert.ok(stderrJson, "block reason JSON on stderr");
+    assert.equal(stderrJson.decision, "block");
+    assert.match(stderrJson.reason, /LOCAL-COMMIT-BEFORE-EXIT/);
+  } finally {
+    fs.rmSync(dir, { recursive: true, force: true });
+  }
+});
--- a/test/contracts/claude-hooks-settings.test.mjs
+++ b/test/contracts/claude-hooks-settings.test.mjs
@@ -380,7 +380,7 @@ test("SubagentStop hook still blocks when porcelain output exceeds the 1MB Node 
     for (let i = 0; i < 4600; i++) {
       fs.writeFileSync(path.join(dir, name(i)), "", "utf8");
     }
-    const porcelainBytes = spawnSync("git", ["status", "--porcelain"], { cwd: dir, encoding: "utf8", maxBuffer: 16 * 1024 * 1024 }).stdout.length;
+    const porcelainBytes = Buffer.byteLength(spawnSync("git", ["status", "--porcelain"], { cwd: dir, encoding: "utf8", maxBuffer: 16 * 1024 * 1024 }).stdout, "utf8");
     assert.ok(porcelainBytes > 1024 * 1024, `fixture must exceed the 1MB default maxBuffer (got ${porcelainBytes} bytes)`);
     const { code, stderrJson } = runHook("subagent-stop-uncommitted-guard.mjs", { cwd: dir });
     assert.equal(code, 2, "a >1MB-porcelain dirty worktree must still be refused (exit 2)");
--- a/test/contracts/claude-hooks-settings.test.mjs
+++ b/test/contracts/claude-hooks-settings.test.mjs
@@ -365,12 +365,14 @@ test("SubagentStop hook fail-safe-allows a mkdir-only (non-git) dir under tmp/wo
 });
 
 test("SubagentStop hook still blocks when porcelain output exceeds the 1MB Node default maxBuffer (#1686)", () => {
-  // With Node's default execFileSync maxBuffer (1MB), ~4000+ dirty paths make
-  // `git status --porcelain` throw ERR_CHILD_PROCESS_STDIO_MAXBUFFER, and the catch
+  // With Node's default execFileSync maxBuffer (1MB), `git status --porcelain` throws
+  // ERR_CHILD_PROCESS_STDIO_MAXBUFFER once the output crosses that size, and the catch
   // turns that into a fail-safe allow — defeating the guard on exactly the worktrees
   // with the most uncommitted work. The hook raises maxBuffer to 10MB; this test
   // creates enough long-named untracked files to push porcelain past 1MB and asserts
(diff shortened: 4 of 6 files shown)


.claude/hooks/subagent-stop-uncommitted-guard.mjs
```diff
@@ -34,8 +34,12 @@ let porcelain = "";
 try {
   // Bounded: a stalled `git status` (slow/networked FS, held index lock) must not hang the
   // subagent stop. On timeout git is killed and execFileSync throws → caught → porcelain=""
-  // → fail-safe allow (better than blocking the exit with no message). #1619 review finding.
-  porcelain = execFileSync("git", ["status", "--porcelain"], { cwd, encoding: "utf8", timeout: 5000 });
+  // → fail-safe allow (better than blocking the exit with no message).
+  // maxBuffer is raised to 10MB: the Node default (1MB) makes execFileSync throw on a
+  // dirty worktree with ~4000+ status lines, which the catch would turn into a fail-safe
```
mfittko: **low** (`contract-surface`): The documented thresholds in the new maxBuffer comment are fixture-derived, not realistic. The comment claims the 1MB Node default throws at "~4000+ status lines" and that 10MB is reached at "~40k+ dirty paths", which implies ~262 bytes per porcelain line. That is only true for the e2e fixture, which pads every filename with 230 x characters. Real porcelain lines in this repo average 45.3 bytes across 930 tracked paths, so 1MB is ~23,000 lines and 10MB is ~231,000 lines. The linked issue 1686 states a third figure, ~7000+ files for 1MB. Acceptance criterion 2 pins this comment as the place the ceiling is documented, so the stated numbers are the contract surface and they disagree with runtime by roughly 5x.

test/contracts/claude-hooks-settings.test.mjs
```diff
@@ -363,3 +363,29 @@ test("SubagentStop hook fail-safe-allows a mkdir-only (non-git) dir under tmp/wo
     fs.rmSync(dir, { recursive: true, force: true });
   }
 });
+
+test("SubagentStop hook still blocks when porcelain output exceeds the 1MB Node default maxBuffer (#1686)", () => {
+  // With Node's default execFileSync maxBuffer (1MB), ~4000+ dirty paths make
```
mfittko: **nit** (`contract-surface`): The new e2e test comment repeats the same fixture-derived threshold, "~4000+ dirty paths make git status --porcelain throw", immediately above the fixture that reaches it only by using 230-character filenames. Left as is, the inaccurate number has two sources and a later corrective edit to the hook comment can miss this copy.

.claude/hooks/subagent-stop-uncommitted-guard.mjs
```diff
@@ -34,8 +34,12 @@ let porcelain = "";
 try {
   // Bounded: a stalled `git status` (slow/networked FS, held index lock) must not hang the
   // subagent stop. On timeout git is killed and execFileSync throws → caught → porcelain=""
-  // → fail-safe allow (better than blocking the exit with no message). #1619 review finding.
-  porcelain = execFileSync("git", ["status", "--porcelain"], { cwd, encoding: "utf8", timeout: 5000 });
+  // → fail-safe allow (better than blocking the exit with no message).
+  // maxBuffer is raised to 10MB: the Node default (1MB) makes execFileSync throw on a
```
mfittko: **nit** (`correctness`): The documented buffer arithmetic is derived from the artificial test fixture, not from realistic path lengths. The comment states the 1MB default throws at roughly 4000 status lines and that 10MB tops out around 40k dirty paths; both figures assume the e2e fixture's ~245-byte lines. Typical porcelain lines are 40 to 60 bytes, so the real thresholds are roughly 17k and 200k paths, and linked issue 1686 itself states 7000+ for the 1MB case.

packages/core/test/claude-hook-decisions.test.mjs
```diff
@@ -541,6 +541,16 @@ test("decideSubagentStopGuard treats a non-string cwd as out-of-scope (allow)",
   assert.equal(decideSubagentStopGuard({ cwd: undefined, porcelain: " M x" }).decision, "allow");
 });
 
+test("decideSubagentStopGuard caps the enumerated dirty paths at 50 and reports the full count", () => {
```
mfittko: **medium** (`coverage`): The only test for the new 50-path cap uses 120 dirty paths, so the cap's own boundary is unpinned. The slice bound is killed by the 120-path case, but the suffix comparison is not: at exactly 50 dirty paths the current code correctly emits no truncation summary, and mutating that comparison from greater-than to greater-or-equal produces a spurious 'and 0 more' line with the whole suite still green. The 51-path case, the smallest input where the summary is legitimate, is likewise untested.

test/contracts/claude-hooks-settings.test.mjs
```diff
@@ -363,3 +363,29 @@ test("SubagentStop hook fail-safe-allows a mkdir-only (non-git) dir under tmp/wo
...
+    for (let i = 0; i < 4600; i++) {
+      fs.writeFileSync(path.join(dir, name(i)), "", "utf8");
+    }
+    const porcelainBytes = spawnSync("git", ["status", "--porcelain"], { cwd: dir, encoding: "utf8", maxBuffer: 16 * 1024 * 1024 }).stdout.length;
+    assert.ok(porcelainBytes > 1024 * 1024, `fixture must exceed the 1MB default maxBuffer (got ${porcelainBytes} bytes)`);
+    const { code, stderrJson } = runHook("subagent-stop-uncommitted-guard.mjs", { cwd: dir });
+    assert.equal(code, 2, "a >1MB-porcelain dirty worktree must still be refused (exit 2)");
+    assert.ok(stderrJson, "block reason JSON on stderr");
+    assert.equal(stderrJson.decision, "block");
+    assert.match(stderrJson.reason, /LOCAL-COMMIT-BEFORE-EXIT/);
```
mfittko, line 389: **low** (`coverage`): The new e2e test exercises the 4600-file fixture that motivated the companion reason-cap fix, but asserts nothing about the cap: it checks exit code 2, decision block, and the LOCAL-COMMIT-BEFORE-EXIT marker only. The bounded-stderr outcome is pinned only incidentally, because runHook's spawnSync at line 27 uses Node's default 1MB buffer, so an uncapped multi-megabyte reason happens to break the stderr parse. That coverage is fragile and reports a misleading cause: any unrelated bump of runHook's maxBuffer silently drops it, and a regression surfaces as 'block reason JSON on stderr' rather than as an unbounded reason.

.claude/hooks/subagent-stop-uncommitted-guard.mjs
```diff
@@ -34,8 +34,12 @@ let porcelain = "";
...
+  // dirty worktree with ~4000+ status lines, which the catch would turn into a fail-safe
+  // allow — defeating the guard exactly when the most work is at risk. Beyond 10MB
+  // (~40k+ dirty paths) the fail-safe allow is the documented ceiling.
+  porcelain = execFileSync("git", ["status", "--porcelain"], { cwd, encoding: "utf8", timeout: 5000, maxBuffer: 10 * 1024 * 1024 });
```
mfittko, line 46: **low** (`coverage`): The 10MB ceiling is asserted only as prose. Tests exercise roughly 1.1MB of porcelain, which pins 'above the 1MB default still blocks' but not the chosen value: lowering maxBuffer to 2MB keeps every suite green, and the comment's documented beyond-10MB fail-safe-allow behavior has no test at all. The generic catch-to-allow branch it would land on is already pinned by the non-git fail-safe test at line 343 of the contracts suite, so the fallback shape is covered even though the maxBuffer trigger is not. The linked issue's second acceptance criterion explicitly permits documenting the ceiling instead of testing it, so this is a noted residual rather than an unmet criterion.

packages/core/test/claude-hook-decisions.test.mjs
```diff
@@ -541,6 +541,16 @@ test("decideSubagentStopGuard treats a non-string cwd as out-of-scope (allow)",
...
+  const lines = Array.from({ length: 120 }, (_, i) => `?? file-${String(i).padStart(3, "0")}.txt`);
```
mfittko: **low** (`determinism`): The cap test's fixture lines are generated in already-sorted order (file-000..file-119), so the assertions cannot distinguish 'preserves porcelain order' from 'sorts the paths'. Order preservation is behaviourally load-bearing: git porcelain lists tracked staged/modified entries before untracked ones, so a sort creeping into the decider would displace the higher-data-loss-risk modified files out of the 50 shown by alphabetically earlier untracked ones.

packages/core/src/claude/hook-decisions.mjs
```diff
@@ -422,13 +422,22 @@ export function decideSubagentStopGuard({ cwd, porcelain, pendingCommitAuthoriza
...
     .map((l) => l.trim())
     .filter(Boolean);
+  // The reason is fed back to the stopping subagent as context, so the path
+  // enumeration is capped: an unbounded list on a very dirty worktree produces a
+  // multi-megabyte reason that consumers truncate or choke on, burying the one
+  // actionable line. The count always reports the full total.
+  const MAX_LISTED_DIRTY_PATHS = 50;
+  const listed = dirty.slice(0, MAX_LISTED_DIRTY_PATHS).map((p) => "  " + p);
+  if (dirty.length > MAX_LISTED_DIRTY_PATHS) {
+    listed.push(`  … and ${dirty.length - MAX_LISTED_DIRTY_PATHS} more (run \`git status --porcelain\` for the full list)`);
```
mfittko, line 432: **nit** (`determinism`): The truncation line reports the remaining count but not that the enumerated sample is the leading 50 in git's path order. Because git status sorts within its sections, the visible sample is systematically biased toward alphabetically early paths, which can read as an arbitrary or representative sample to the stopping subagent.

test/contracts/claude-hooks-settings.test.mjs
```diff
@@ -363,3 +363,29 @@ test("SubagentStop hook fail-safe-allows a mkdir-only (non-git) dir under tmp/wo
...
+  try {
+    const name = (i) => `f${String(i).padStart(5, "0")}-${"x".repeat(230)}.txt`;
+    // ~4600 files x ~240-byte porcelain lines ≈ 1.1MB of status output.
+    for (let i = 0; i < 4600; i++) {
+      fs.writeFileSync(path.join(dir, name(i)), "", "utf8");
+    }
+    const porcelainBytes = spawnSync("git", ["status", "--porcelain"], { cwd: dir, encoding: "utf8", maxBuffer: 16 * 1024 * 1024 }).stdout.length;
+    assert.ok(porcelainBytes > 1024 * 1024, `fixture must exceed the 1MB default maxBuffer (got ${porcelainBytes} bytes)`);
+    const { code, stderrJson } = runHook("subagent-stop-uncommitted-guard.mjs", { cwd: dir });
+    assert.equal(code, 2, "a >1MB-porcelain dirty worktree must still be refused (exit 2)");
```
mfittko, line 386: **nit** (`determinism`): The e2e asserts exit code 2, but the hook also fail-safe-allows (exit 0) when its 5s git status timeout trips. On a 4600-file fixture under I/O contention a timeout would fail this assertion with a message that reads as a maxBuffer regression, making the two distinct causes indistinguishable from the failure output.

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `c377b8ebe95453026c5b25bc4f482489adc72dec`
**Verdict:** clean

**Findings summary:** 8 angles reviewed; 15 findings (see per-angle breakdown below).

- `determinism` → `findings_present` (3 findings)
- `gate-evidence` → `clean`
- `scope` → `clean`

> **low** (`contract-surface`): The hook docblock still describes the pre-cap reason format. It says the block emits "stderr JSON naming LOCAL-COMMIT-BEFORE-EXIT and the dirty paths", but after this change the reason enumerates at most 50 paths and then a truncation line with the remaining count. This is a live, hand-written doc in the changed file that pins the old unbounded-enumeration behavior.
> Location: `.claude/hooks/subagent-stop-uncommitted-guard.mjs:9`


> **low** (`contract-surface`): The Unreleased section is empty, so the only changelog description of the guard remains the released 1.0.0-rc.6 entry at line 34, which states the hook refuses the stop "naming LOCAL-COMMIT-BEFORE-EXIT and the dirty paths". That released line must not be rewritten, so with no Unreleased entry the shipped changelog record of the reason format stays the unbounded one and the 10MB maxBuffer change is undocumented outside the hook comment.
> Location: `CHANGELOG.md:5`


> **low** (`pr-description`): The repo's own PR-body checker fails on this body. Running scripts/loop/validate-pr-body-spec.mjs --repo mfittko/dev-loops --pr 1749 --expected-issue 1686 returns ok false with the single error missing_open_questions, because the body has no Open questions or Risks section. All other spec invariants pass, including the closing reference to issue 1686. Severity is low rather than medium because the coordination detector resolves specSource as linked_issue for this PR, so the ARTIFACT-LIGHTWEIGHT-BODY-INVARIANTS rule is not unconditionally binding here, but the sanctioned checker still reports a failure against this body.


> **low** (`pr-description`): The reason path cap is shipped behavior with no acceptance criterion. The Acceptance criteria section has two checkbox items, both about the 10MB maxBuffer. The user-visible change that truncates the block reason's dirty-path enumeration at 50 entries appears only in the Scope and context prose and as the third row of the AC / DoD / Non-goals matrix, which is a row with no matching entry in the AC section. Behavior that is not an acceptance criterion is not checked off at merge and is not traced back to issue 1686.

**Next action:** Fixer applies the 9 code-side act findings; conductor applies 2 body-only fixes; re-gate at the fixed head, then mark ready

packages/core/test/claude-hook-decisions.test.mjs
mfittko: **low** (`input-validation`): The cap boundary is pinned only from above, at 120 dirty paths. The exactly-50 case is untested, so a regression from the strict greater-than to greater-or-equal would append a nonsensical trailing line reading and 0 more with no test failing. The 51-path case, the smallest input that must produce the summary line, is likewise unpinned.

packages/core/src/claude/hook-decisions.mjs
```diff
@@ -422,13 +422,22 @@ export function decideSubagentStopGuard({ cwd, porcelain, pendingCommitAuthoriza
     .split("\n")
     .map((l) => l.trim())
     .filter(Boolean);
+  // The reason is fed back to the stopping subagent as context, so the path
+  // enumeration is capped: an unbounded list on a very dirty worktree produces a
+  // multi-megabyte reason that consumers truncate or choke on, burying the one
+  // actionable line. The count always reports the full total.
+  const MAX_LISTED_DIRTY_PATHS = 50;
+  const listed = dirty.slice(0, MAX_LISTED_DIRTY_PATHS).map((p) => "  " + p);
```
mfittko, line 430: **low** (`input-validation`): The cap bounds the number of listed paths but not the byte length of each one, so the reason size is still driven by untrusted input. Porcelain paths can approach the filesystem path limit and git C-quotes unusual characters, which roughly doubles the encoded length, so 50 pathological paths still yield a reason in the hundreds of kilobytes. That is far better than the multi-megabyte case being fixed, but it leaves the stated consumability goal dependent on path length rather than enforced.

.claude/hooks/subagent-stop-uncommitted-guard.mjs
```diff
@@ -6,9 +6,10 @@
  * uncommitted changes in a worktree are destroyed with no warning — the only data-loss gap in
  * the enforcement audit. `LOCAL-COMMIT-BEFORE-EXIT` existed only as prose; this hook makes it
  * mechanical: when a subagent stops with a dirty worktree under `tmp/worktrees/`, refuse the
- * stop (exit code 2 + stderr JSON naming `LOCAL-COMMIT-BEFORE-EXIT` and the dirty paths) so the
- * subagent commits before exiting. Post-merge cleanup that force-removes worktrees can no
- * longer destroy uncommitted work silently.
+ * stop (exit code 2 + stderr JSON naming `LOCAL-COMMIT-BEFORE-EXIT` and up to 50 dirty paths,
+ * plus a summary line with the full count when there are more) so the subagent commits before
```
mfittko, line 10: **low** (`contract-surface`): The hook docblock says the block reason emits "a summary line with the full count when there are more", but the runtime summary line emits the REMAINING count, not the total. decideSubagentStopGuard pushes "… and ${dirty.length - 50} more" and reports the full total only in the preceding "Dirty paths (N):" header. The CHANGELOG entry for the same change states this correctly -- "a summary line naming the remaining count (the reported total is always the full count)" -- so the two documentation surfaces describing one behavior disagree, and the docblock is the inaccurate one.

test/contracts/claude-hooks-settings.test.mjs
```diff
@@ -363,3 +363,35 @@ test("SubagentStop hook fail-safe-allows a mkdir-only (non-git) dir under tmp/wo
...
+    assert.ok(stderrJson, "block reason JSON on stderr");
+    assert.equal(stderrJson.decision, "block");
+    assert.match(stderrJson.reason, /LOCAL-COMMIT-BEFORE-EXIT/);
+    // The reason itself stays bounded (the 50-path cap), not merely parseable off stderr —
+    // pins the cap end to end rather than relying on runHook's own maxBuffer as an incidental
+    // proxy for it.
+    assert.match(stderrJson.reason, /… and 4550 more/);
```
mfittko, line 393: **low** (`determinism`): The e2e truncation assertion hardcodes the remainder literal 'and 4550 more', which is derived from the assumption that all 4600 fixture files show up as untracked. That count is deterministic only for a git host with no interfering excludes: a global core.excludesFile pattern, a populated init.templateDir info/exclude, or a stray .DS_Store in the fixture dir shifts the untracked count by one or more and fails this assertion with a misleading message, while the sibling precondition assert on porcelainBytes greater than 1MB still passes. The spawnSync on the preceding line already captures the real porcelain stdout, so the expected remainder is available without a second git call.

mfittko, line 393: **low** (`input-validation`): The e2e truncation assertion hardcodes the remainder 4550, which assumes the fixture worktree contains exactly the 4600 files the test wrote. makeWorktree uses fs.mkdirSync with recursive true and never clears a pre-existing directory, and the directory name is keyed only on process.pid, so a stale directory left by a prior run killed before its finally-block rmSync would inflate the dirty count and fail this assertion with a misleading message. The other tests in this file assert only on path presence or absence, so this is the first assertion in the suite that depends on the fixture directory being empty on entry.

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `2f9b4b2c1f1250c4aa5252eb1847d6cb28342900`
**Verdict:** clean

**Findings summary:** 8 angles reviewed; 3 findings (see per-angle breakdown below).

- `correctness` → `clean`
- `coverage` → `clean`
- `gate-evidence` → `clean`
- `pr-description` → `clean`
- `scope` → `clean`

**Next action:** Mark the PR ready for review (draft gate clean; 2 judge-deferred lows tracked as a follow-up)

.claude/hooks/subagent-stop-uncommitted-guard.mjs
mfittko: Deferred at gate close (round 2, fingerprint add86b49ff2b9a55, severity low, angle contract-surface): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

test/contracts/claude-hooks-settings.test.mjs
mfittko: Deferred at gate close (round 2, fingerprint 40f519525d22f691, severity nit, angle contract-surface): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

.claude/hooks/subagent-stop-uncommitted-guard.mjs
mfittko: Deferred at gate close (round 2, fingerprint 15f3cb5d6983be98, severity nit, angle correctness): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

test/contracts/claude-hooks-settings.test.mjs
mfittko: Deferred at gate close (round 2, fingerprint 7b25b31cc1cc7c00, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

.claude/hooks/subagent-stop-uncommitted-guard.mjs
mfittko: Deferred at gate close (round 2, fingerprint 56f86892f37291c9, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/test/claude-hook-decisions.test.mjs
mfittko: Deferred at gate close (round 2, fingerprint d1415fe8f87a1256, severity low, angle determinism): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/claude/hook-decisions.mjs
mfittko: Deferred at gate close (round 2, fingerprint 51354eccb5542d7a, severity nit, angle determinism): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

test/contracts/claude-hooks-settings.test.mjs
mfittko: Deferred at gate close (round 2, fingerprint 02c41071ed1f5616, severity nit, angle determinism): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/test/claude-hook-decisions.test.mjs
mfittko: Deferred at gate close (round 2, fingerprint 09d9b51d4a271fd2, severity low, angle input-validation): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/claude/hook-decisions.mjs
mfittko: Deferred at gate close (round 2, fingerprint cab4d0319589c953, severity low, angle input-validation): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

.claude/hooks/subagent-stop-uncommitted-guard.mjs
mfittko: Deferred at gate close (round 2, fingerprint 5f657f58b30ceb49, severity low, angle contract-surface): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

test/contracts/claude-hooks-settings.test.mjs
mfittko: Deferred at gate close (round 2, fingerprint ddc39dbb8bdbd7fc, severity low, angle determinism): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

mfittko: Deferred at gate close (round 2, fingerprint a43035853ae38c22, severity low, angle input-validation): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/test/claude-hook-decisions.test.mjs
mfittko: Fixed in 2f9b4b2c: added boundary unit tests at exactly 50 paths (no truncation summary) and 51 paths (summary reports one more), and the 120-path cap test now uses a descending, non-sorted fixture asserting order preservation. The e2e test additionally pins the truncation summary end to end.

.claude/hooks/subagent-stop-uncommitted-guard.mjs
mfittko, line 10: **low** (`docs`): The added docblock says the block reason carries 'a summary line with the full count when there are more', but the summary line carries the REMAINING count ('… and N more'), and the full count is in the always-present 'Dirty paths (N):' header rather than only when the list overflows. The CHANGELOG entry for the same change states this correctly ('a summary line naming the remaining count (the reported total is always the full count)'), so the hook docblock is the one surface that misdescribes the emitted reason.

test/contracts/claude-hooks-settings.test.mjs
```diff
@@ -363,3 +363,35 @@ test("SubagentStop hook fail-safe-allows a mkdir-only (non-git) dir under tmp/wo
...
+  // turns that into a fail-safe allow — defeating the guard on exactly the worktrees
+  // with the most uncommitted work. The hook raises maxBuffer to 10MB; this test
+  // creates enough long-named untracked files to push porcelain past 1MB and asserts
+  // the guard still blocks. Using this fixture's long (~245-byte) porcelain lines, that
+  // takes ~4300+ files; real (much shorter) paths need far more to hit the same bound.
+  // Beyond 10MB (~43k+ such long-line paths) the fail-safe allow is the documented ceiling.
+  const dir = makeWorktree("maxbuffer", false);
+  try {
+    const name = (i) => `f${String(i).padStart(5, "0")}-${"x".repeat(230)}.txt`;
+    // ~4600 files x ~240-byte porcelain lines ≈ 1.1MB of status output.
```
mfittko, line 379: **nit** (`docs`): Within the same new test, the docblock at line 373 sizes the fixture's porcelain lines at ~245 bytes while the inline comment at line 379 sizes the same lines at ~240 bytes. The fixture's actual line is 245 bytes (3-byte '?? ' status prefix + a 241-char name + newline), so the ~240 figure is the drifting one; the 1.1MB conclusion it supports still holds either way.

mfittko (commented):


**Reviewed head SHA:** `3304782cda2601e17cb6a4933e0778be86728539`
**Verdict:** clean

**Findings summary:** 7 angles reviewed; 2 findings (see per-angle breakdown below).

- `docs` → `findings_present` (2 findings)
- `dry` → `clean`
- `yagni` → `clean`

**Next action:** Merge under the recorded standing authorization (ADR 0050): full gate pass complete

.claude/hooks/subagent-stop-uncommitted-guard.mjs
mfittko: Deferred at gate close (round 1, fingerprint 27db0c5644325c79, severity low, angle docs): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

test/contracts/claude-hooks-settings.test.mjs
mfittko: Deferred at gate close (round 1, fingerprint cc333fd3496720b5, severity nit, angle docs): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

PR #1759, opened 2026-08-19
fix(config): fail closed on schema-invalid blockCleanOnFindingSeverities in gate severity paths

### LINKED ISSUE #1753: Fail closed on schema-invalid config in gate severity paths instead of raw-config passthrough (deferred from the scan-dedup drive)
## Summary

Judge-deferred from the blocking-severity-scan dedup drive: `loadDevLoopConfig` returns the raw merged config when zod `safeParse` fails, and downstream severity-vocabulary consumers (the shared blocking-severity scan, gate-config resolution) then operate on unvalidated values. The refactor dropped a defensive `Object.hasOwn` vocabulary check that was reachable only through such a schema-invalid config. Decide and implement the fail-closed posture: an out-of-vocabulary `blockCleanOnFindingSeverities` (or a config that failed validation wholesale) should refuse gate operations rather than pass raw values through.

## Acceptance criteria

- [x] A schema-invalid config reaching gate verdict/severity paths produces an explicit refusal naming the validation failure, not silent raw-config passthrough.
- [x] In-vocabulary configs and the legacy severity aliases behave unchanged.
- [x] A test pins the refusal for an out-of-vocabulary blockCleanOnFindingSeverities value.

## Definition of done

- [x] Affected suites green (config, upsert-checkpoint-verdict).
- [x] npm run verify green.

## Non-goals

- Widening the severity vocabulary.
- Changing resolveGateConfig fallback semantics for merely-absent keys.

mfittko:
## Summary

The fail-closed posture the issue asked to decide is implemented on both arms. Arm 1, key-scoped: `resolveGateConfig` refuses every schema-invalid shape of a present `blockCleanOnFindingSeverities` key (non-array, empty array, entry outside the schema's exact spellings), so no severity consumer resolving the affected gate can operate on unvalidated vocabulary. Arm 2, wholesale: the verdict poster (`upsert-checkpoint-verdict.mjs`) now fails closed on a non-empty `loadDevLoopConfig` errors array before any GitHub read — previously it discarded the errors and posted a verdict computed from a raw merged config. Among the sibling severity consumers, consolidate-fanin already throws on a non-empty errors array; close-gate-findings and detect-checkpoint-evidence degrade to defaults instead, and aligning their posture (plus the adjacent comments) is follow-up issue 1763.

## Scope and context

Four files. `packages/core/src/config/config.mjs`: `resolveGateConfig` refuses every schema-invalid shape of a present `blockCleanOnFindingSeverities` key with an explicit error naming the offending value(s) and the `gates.<gate>.blockCleanOnFindingSeverities` key. The guard exact-matches raw entries against `BLOCKING_SEVERITY_SPELLINGS`, the same list the schema enum consumes, so its accept set is byte-identical to the schema's (a whitespace-varied or mixed-case spelling refuses rather than being trim-normalized into acceptance). Offending values render through `formatConfigValue`, safe against throwing `JSON.stringify` (BigInt, cycles), undefined-returning renders (symbols), and throwing `String()` (null-prototype cycles). The docblock states the refusal's per-gate blast radius and the boundary with the degrade-quietly convention (resolveMaxAnglesPerGroup substitutes its default, resolveFanoutGroups drops malformed entries). `scripts/github/upsert-checkpoint-verdict.mjs`: refuses on a non-empty config errors array at entry. `packages/core/test/config.test.mjs`: refusal pins for out-of-vocabulary tokens (single and multiple, named in the message), non-string, unrenderable, and symbol entries, mixed-case and whitespace-varied spellings, the empty array, present-but-non-array shapes, plus a schema-enum parity pin. `test/github/upsert-checkpoint-verdict.test.mjs`: pins the poster's refusal on a schema-invalid worktree config, and repairs a pre-existing test fixture whose `.devloops` was itself schema-invalid (a `lightMode` block missing the required `maxFiles`/`maxLines`) — concrete evidence of the wholesale refusal's blast radius: any worktree whose config fails validation for any reason now refuses at the poster.

## Acceptance criteria

- [ ] A schema-invalid config reaching gate verdict/severity paths produces an explicit refusal naming the validation failure, not silent raw-config passthrough: non-array, empty-array, and out-of-vocabulary `blockCleanOnFindingSeverities` shapes throw from `resolveGateConfig` (the boundary severity consumers resolve through), and the verdict poster refuses on any config that failed validation wholesale.
- [ ] In-vocabulary configs and the legacy severity aliases behave unchanged (existing normalization/dedup tests untouched and green; absent key still defaults to high).
- [ ] A test pins the refusal for an out-of-vocabulary `blockCleanOnFindingSeverities` value.

## Definition of done

- [ ] `node --test packages/core/test/config.test.mjs` green.
- [ ] `node --test test/github/upsert-checkpoint-verdict.test.mjs` green.
- [ ] `npm run schema:check` green (no schema change; vocabulary set mirrors the existing enum).
- [ ] `npm run verify` green.

|---|---|---|
| Explicit refusal on schema-invalid severity shapes | config suite green (resolver refusal pins for non-array, empty, out-of-vocabulary); upsert suite green (poster wholesale-refusal pin) | resolveGateConfig fallback for absent keys unchanged |
| In-vocabulary and legacy aliases unchanged | config suite green (existing normalization tests); npm run verify green | vocabulary not widened |
| Refusal test pinned | config suite green (out-of-vocabulary refusal pin present and passing) | — |

## Non-goals

- Widening the severity vocabulary.
- Changing `resolveGateConfig` fallback semantics for merely-absent keys (an absent key still resolves to the high default).
- Changing how other severity consumers (`consolidateFanin`, the verdict poster's own re-normalization) treat directly-passed severity lists; they receive resolved values and a follow-up covers their redundant re-normalization.
- Changing `loadDevLoopConfig`'s raw-config-plus-errors return shape.

## Open questions

None.

## Validation

```sh
node --test packages/core/test/config.test.mjs
node --test test/github/upsert-checkpoint-verdict.test.mjs
npm run schema:check
npm run verify
```

Closes #1753

### DIFF
--- a/packages/core/src/config/config.mjs
+++ b/packages/core/src/config/config.mjs
@@ -219,6 +219,13 @@ const GateDynamicConfig = z.strictObject({
 // accepted but INERT for spike (a findings-doc deliverable has no "clean
 // verdict" escalation path and no additive dynamic pool) rather than being
 // split into a second schema.
+// Canonical defect-severity vocabulary for blockCleanOnFindingSeverities:
+// the post-normalization form of the schema enum below (legacy spellings
+// normalize into these three). resolveGateConfig fails closed against this
+// set so a config that bypassed schema validation cannot reach severity
+// consumers with an unknown value.
+const BLOCKING_SEVERITY_VOCABULARY = new Set(["high", "medium", "low"]);
+
 const GateConfig = z.strictObject({
   angles: z.array(GateAngleEntry).optional().describe("Review lenses this gate fans out to. A bare string is sugar for { name }; an object may set mandatory/enabled/persona/prompt/model/tier."),
   dynamic: GateDynamicConfig.optional().describe("Diff-driven dynamic angle selection policy for this gate."),
@@ -1839,9 +1846,30 @@ export function resolveRefinement(config) {
  * @param {DevLoopConfig} config
  * @param {"draft"|"preApproval"|"spike"} gate
  * @returns {{ angles: string[]|null, excludeAngles: string[], mandatoryAngles: string[], required: boolean, requireCi: boolean, blockCleanOnFindingSeverities: string[], dynamicAngles: boolean, additiveAngles: boolean, mediumFixWindow: number, tiers: Array<{name: string, match: object, angles: string[]}> }}
+ * @throws {Error} when a configured `blockCleanOnFindingSeverities` entry is
+ *   outside the defect vocabulary after normalization — such a value can only
+ *   arrive through a config that failed schema validation (the schema's enum
+ *   rejects it), and passing it through would make the gate silently block on
+ *   nothing. Refusing here, at the one resolve boundary every severity
+ *   consumer shares, keeps unvalidated vocabulary out of every gate verdict
+ *   path. Absent keys still fall back to the defaults unchanged.
  */
 export function resolveGateConfig(config, gate) {
   const gateConfig = config?.gates?.[gate];
+  const rawBlocking = gateConfig?.blockCleanOnFindingSeverities;
+  let blockCleanOnFindingSeverities = ["high"];
+  if (rawBlocking && Array.isArray(rawBlocking)) {
+    blockCleanOnFindingSeverities = [...new Set(rawBlocking.map((s) => normalizeSeverity(s)))];
+    const invalid = blockCleanOnFindingSeverities.filter((s) => !BLOCKING_SEVERITY_VOCABULARY.has(s));
+    if (invalid.length > 0) {
+      throw new Error(
+        `Config validation failed: gates.${gate}.blockCleanOnFindingSeverities contains ` +
+        `out-of-vocabulary severity value(s) after normalization: ${invalid.map((s) => JSON.stringify(s)).join(", ")}. ` +
+        `Allowed: high, medium, low (legacy spellings must-fix, worth-fixing-now, nice-to-have, defer normalize to these). ` +
+        `Fix the config before gate operations can proceed.`
+      );
+    }
+  }
   const entries = normalizeAngleEntries(gateConfig?.angles);
   // An explicitly-empty (or all-garbage/malformed) array is a real configured
   // "no angles" — distinct from the key being absent entirely, which callers
@@ -1855,12 +1883,11 @@ export function resolveGateConfig(config, gate) {
     requireCi: gateConfig?.requireCi ?? true,
     dynamicAngles: gateConfig?.dynamic?.subtractive ?? true,
     additiveAngles: gateConfig?.dynamic?.additive ?? false,
-    // Normalized + deduped at the resolve boundary so every consumer (envelope,
-    // verdict poster, fan-in, viewer) sees canonical spellings only; a
-    // half-migrated ["must-fix","low","defer"] collapses to two entries.
-    blockCleanOnFindingSeverities: gateConfig?.blockCleanOnFindingSeverities && Array.isArray(gateConfig.blockCleanOnFindingSeverities)
-      ? [...new Set(gateConfig.blockCleanOnFindingSeverities.map((s) => normalizeSeverity(s)))]
-      : ["high"],
+    // Normalized + deduped at the resolve boundary (above) so every consumer
+    // (envelope, verdict poster, fan-in, viewer) sees canonical spellings
+    // only; a half-migrated ["must-fix","low","defer"] collapses to two
+    // entries, and anything outside the vocabulary has already thrown.
+    blockCleanOnFindingSeverities,
     // `mediumFixWindow` wins; `worthFixingNowFixWindow` is the deprecated
     // pre-rename key, still honored so an unmigrated config keeps its
     // configured window rather than silently reverting to the default.
--- a/packages/core/src/config/config.mjs
+++ b/packages/core/src/config/config.mjs
@@ -220,23 +220,29 @@ const GateDynamicConfig = z.strictObject({
 // verdict" escalation path and no additive dynamic pool) rather than being
 // split into a second schema.
 // Single source for the blockCleanOnFindingSeverities vocabulary: the schema
-// enum below consumes the spellings verbatim, and the fail-closed set in
-// resolveGateConfig derives from the same list through normalizeSeverity, so
-// widening the enum can never leave the runtime guard behind (drift would
-// have refused configs that pass schema validation).
+// enum below consumes these spellings verbatim, and resolveGateConfig's
+// fail-closed guard exact-matches raw entries against the same list, so the
+// guard's accept set is byte-identical to the schema's (no trim/normalize
+// superset) and widening the enum can never leave the runtime guard behind.
 const BLOCKING_SEVERITY_SPELLINGS = ["high", "medium", "low", "must-fix", "worth-fixing-now", "nice-to-have", "defer"];
-const BLOCKING_SEVERITY_VOCABULARY = new Set(BLOCKING_SEVERITY_SPELLINGS.map((s) => normalizeSeverity(s)));
+const BLOCKING_SEVERITY_SPELLING_SET = new Set(BLOCKING_SEVERITY_SPELLINGS);
 
 // Render an offending config value for a refusal message without letting the
 // renderer itself throw: JSON.stringify raises on BigInt and circular
-// structures and returns undefined for undefined/symbol/function, so those
-// fall back to String().
+// structures and returns undefined for undefined/symbol/function (those fall
+// back to String()), and String() itself can throw for exotic values (a
+// null-prototype cycle, a throwing Symbol.toPrimitive) — those get a literal
+// placeholder so the refusal always surfaces as the refusal.
 function formatConfigValue(value) {
   try {
     const rendered = JSON.stringify(value);
     return rendered === undefined ? String(value) : rendered;
   } catch {
-    return String(value);
+    try {
+      return String(value);
+    } catch {
+      return "<unrenderable value>";
+    }
   }
 }
 
@@ -1860,21 +1866,25 @@ export function resolveRefinement(config) {
  * @param {DevLoopConfig} config
  * @param {"draft"|"preApproval"|"spike"} gate
  * @returns {{ angles: string[]|null, excludeAngles: string[], mandatoryAngles: string[], required: boolean, requireCi: boolean, blockCleanOnFindingSeverities: string[], dynamicAngles: boolean, additiveAngles: boolean, mediumFixWindow: number, tiers: Array<{name: string, match: object, angles: string[]}> }}
- * @throws {Error} when a PRESENT `blockCleanOnFindingSeverities` key is any
- *   schema-invalid shape: a non-array value, an empty array, or an entry
- *   outside the defect vocabulary after normalization. Every such shape can
- *   only arrive through a config that failed schema validation (the schema
- *   requires a min-1 array of the enum spellings), and passing it through
- *   would make the gate block on the wrong severities or on nothing at all.
- *   The refusal deliberately reaches EVERY caller of this resolver — including
- *   ones reading unrelated fields (resolveRefinement's requireCi read, angle
- *   and tier resolution, state detection) — because no gate operation should
- *   proceed on a config whose verdict-semantics key failed validation. This
- *   is the stated boundary with the module's degrade-to-default convention
- *   (resolveMaxAnglesPerGroup, resolveFanoutGroups): keys that only shape
- *   dispatch ergonomics degrade quietly; the key that decides what blocks a
- *   clean verdict refuses. An ABSENT key still falls back to the default
- *   unchanged.
+ * @throws {Error} when THIS gate's PRESENT `blockCleanOnFindingSeverities`
+ *   key is any schema-invalid shape: a non-array value, an empty array, or an
+ *   entry that is not one of the schema enum's exact spellings. Every such
+ *   shape can only arrive through a config that failed schema validation (the
+ *   schema requires a min-1 array of the enum spellings; the guard
+ *   exact-matches raw entries against the same spelling list, so its accept
+ *   set is byte-identical to the schema's), and passing it through would make
+ *   the gate block on the wrong severities or on nothing at all. The refusal
+ *   reaches every caller that resolves the affected gate — including ones
+ *   reading unrelated fields from it (resolveRefinement's preApproval
+ *   requireCi read, angle and tier resolution, state detection) — because no
+ *   gate operation should proceed on a config whose verdict-semantics key
+ *   failed validation; a caller resolving a DIFFERENT gate of the same config
+ *   is untouched. This is the stated boundary with the module's
+ *   degrade-quietly convention for dispatch-ergonomics keys
+ *   (resolveMaxAnglesPerGroup substitutes its default, resolveFanoutGroups
+ *   drops malformed entries): those keys only shape dispatch, so they degrade;
+ *   the key that decides what blocks a clean verdict refuses. An ABSENT key
+ *   still falls back to the default unchanged.
  */
 export function resolveGateConfig(config, gate) {
   const gateConfig = config?.gates?.[gate];
@@ -1888,23 +1898,24 @@ export function resolveGateConfig(config, gate) {
         `Fix the config before gate operations can proceed.`
       );
     }
-    blockCleanOnFindingSeverities = [...new Set(rawBlocking.map((s) => normalizeSeverity(s)))];
-    if (blockCleanOnFindingSeverities.length === 0) {
+    if (rawBlocking.length === 0) {
       throw new Error(
         `Config validation failed: gates.${gate}.blockCleanOnFindingSeverities is empty; ` +
         `the schema requires at least one blocking severity, and an empty list would make the gate block on nothing. ` +
         `Fix the config before gate operations can proceed.`
       );
     }
-    const invalid = blockCleanOnFindingSeverities.filter((s) => !BLOCKING_SEVERITY_VOCABULARY.has(s));
+    const invalid = rawBlocking.filter((s) => typeof s !== "string" || !BLOCKING_SEVERITY_SPELLING_SET.has(s));
     if (invalid.length > 0) {
       throw new Error(
         `Config validation failed: gates.${gate}.blockCleanOnFindingSeverities contains ` +
-        `out-of-vocabulary severity value(s) after normalization: ${invalid.map((s) => formatConfigValue(s)).join(", ")}. ` +
-        `Allowed: high, medium, low (legacy spellings must-fix, worth-fixing-now, nice-to-have, defer normalize to these). ` +
+        `value(s) outside the schema's severity vocabulary: ${invalid.map((s) => formatConfigValue(s)).join(", ")}. ` +
+        `Allowed (exact spellings): ${BLOCKING_SEVERITY_SPELLINGS.join(", ")} ` +
+        `(the legacy spellings normalize to high/medium/low). ` +
         `Fix the config before gate operations can proceed.`
       );
     }
+    blockCleanOnFindingSeverities = [...new Set(rawBlocking.map((s) => normalizeSeverity(s)))];
   }
   const entries = normalizeAngleEntries(gateConfig?.angles);
   // An explicitly-empty (or all-garbage/malformed) array is a real configured
--- a/packages/core/src/config/config.mjs
+++ b/packages/core/src/config/config.mjs
@@ -219,12 +219,26 @@ const GateDynamicConfig = z.strictObject({
 // accepted but INERT for spike (a findings-doc deliverable has no "clean
 // verdict" escalation path and no additive dynamic pool) rather than being
 // split into a second schema.
-// Canonical defect-severity vocabulary for blockCleanOnFindingSeverities:
-// the post-normalization form of the schema enum below (legacy spellings
-// normalize into these three). resolveGateConfig fails closed against this
-// set so a config that bypassed schema validation cannot reach severity
-// consumers with an unknown value.
-const BLOCKING_SEVERITY_VOCABULARY = new Set(["high", "medium", "low"]);
+// Single source for the blockCleanOnFindingSeverities vocabulary: the schema
+// enum below consumes the spellings verbatim, and the fail-closed set in
+// resolveGateConfig derives from the same list through normalizeSeverity, so
+// widening the enum can never leave the runtime guard behind (drift would
+// have refused configs that pass schema validation).
+const BLOCKING_SEVERITY_SPELLINGS = ["high", "medium", "low", "must-fix", "worth-fixing-now", "nice-to-have", "defer"];
+const BLOCKING_SEVERITY_VOCABULARY = new Set(BLOCKING_SEVERITY_SPELLINGS.map((s) => normalizeSeverity(s)));
+
+// Render an offending config value for a refusal message without letting the
+// renderer itself throw: JSON.stringify raises on BigInt and circular
+// structures and returns undefined for undefined/symbol/function, so those
+// fall back to String().
+function formatConfigValue(value) {
+  try {
+    const rendered = JSON.stringify(value);
+    return rendered === undefined ? String(value) : rendered;
+  } catch {
+    return String(value);
+  }
+}
 
 const GateConfig = z.strictObject({
   angles: z.array(GateAngleEntry).optional().describe("Review lenses this gate fans out to. A bare string is sugar for { name }; an object may set mandatory/enabled/persona/prompt/model/tier."),
@@ -238,7 +252,7 @@ const GateConfig = z.strictObject({
   // admitting either here would let a config block on a severity the
   // disposition pass simultaneously auto-resolves.
   blockCleanOnFindingSeverities: z
-    .array(z.enum(["high", "medium", "low", "must-fix", "worth-fixing-now", "nice-to-have", "defer"]))
+    .array(z.enum(/** @type {[string, ...string[]]} */ (BLOCKING_SEVERITY_SPELLINGS)))
     .min(1)
     .default(["high"])
     .describe("Defect finding severities that block a clean gate verdict (high/medium/low only — \"question\"/\"nit\" are non-defect categories and never block by severity). \"must-fix\" is the deprecated legacy spelling of \"high\", \"worth-fixing-now\" of \"medium\", and \"nice-to-have\"/\"defer\" of \"low\"; consumers normalize them."),
@@ -1846,25 +1860,47 @@ export function resolveRefinement(config) {
  * @param {DevLoopConfig} config
  * @param {"draft"|"preApproval"|"spike"} gate
  * @returns {{ angles: string[]|null, excludeAngles: string[], mandatoryAngles: string[], required: boolean, requireCi: boolean, blockCleanOnFindingSeverities: string[], dynamicAngles: boolean, additiveAngles: boolean, mediumFixWindow: number, tiers: Array<{name: string, match: object, angles: string[]}> }}
- * @throws {Error} when a configured `blockCleanOnFindingSeverities` entry is
- *   outside the defect vocabulary after normalization — such a value can only
- *   arrive through a config that failed schema validation (the schema's enum
- *   rejects it), and passing it through would make the gate silently block on
- *   nothing. Refusing here, at the one resolve boundary every severity
- *   consumer shares, keeps unvalidated vocabulary out of every gate verdict
- *   path. Absent keys still fall back to the defaults unchanged.
+ * @throws {Error} when a PRESENT `blockCleanOnFindingSeverities` key is any
+ *   schema-invalid shape: a non-array value, an empty array, or an entry
+ *   outside the defect vocabulary after normalization. Every such shape can
+ *   only arrive through a config that failed schema validation (the schema
+ *   requires a min-1 array of the enum spellings), and passing it through
+ *   would make the gate block on the wrong severities or on nothing at all.
+ *   The refusal deliberately reaches EVERY caller of this resolver — including
+ *   ones reading unrelated fields (resolveRefinement's requireCi read, angle
+ *   and tier resolution, state detection) — because no gate operation should
+ *   proceed on a config whose verdict-semantics key failed validation. This
+ *   is the stated boundary with the module's degrade-to-default convention
+ *   (resolveMaxAnglesPerGroup, resolveFanoutGroups): keys that only shape
+ *   dispatch ergonomics degrade quietly; the key that decides what blocks a
+ *   clean verdict refuses. An ABSENT key still falls back to the default
+ *   unchanged.
  */
 export function resolveGateConfig(config, gate) {
   const gateConfig = config?.gates?.[gate];
   const rawBlocking = gateConfig?.blockCleanOnFindingSeverities;
   let blockCleanOnFindingSeverities = ["high"];
-  if (rawBlocking && Array.isArray(rawBlocking)) {
+  if (rawBlocking !== undefined) {
+    if (!Array.isArray(rawBlocking)) {
+      throw new Error(
+        `Config validation failed: gates.${gate}.blockCleanOnFindingSeverities must be an array ` +
+        `of defect severities, got ${formatConfigValue(rawBlocking)}. ` +
+        `Fix the config before gate operations can proceed.`
+      );
+    }
     blockCleanOnFindingSeverities = [...new Set(rawBlocking.map((s) => normalizeSeverity(s)))];
+    if (blockCleanOnFindingSeverities.length === 0) {
+      throw new Error(
+        `Config validation failed: gates.${gate}.blockCleanOnFindingSeverities is empty; ` +
+        `the schema requires at least one blocking severity, and an empty list would make the gate block on nothing. ` +
+        `Fix the config before gate operations can proceed.`
+      );
+    }
     const invalid = blockCleanOnFindingSeverities.filter((s) => !BLOCKING_SEVERITY_VOCABULARY.has(s));
     if (invalid.length > 0) {
       throw new Error(
         `Config validation failed: gates.${gate}.blockCleanOnFindingSeverities contains ` +
-        `out-of-vocabulary severity value(s) after normalization: ${invalid.map((s) => JSON.stringify(s)).join(", ")}. ` +
+        `out-of-vocabulary severity value(s) after normalization: ${invalid.map((s) => formatConfigValue(s)).join(", ")}. ` +
         `Allowed: high, medium, low (legacy spellings must-fix, worth-fixing-now, nice-to-have, defer normalize to these). ` +
         `Fix the config before gate operations can proceed.`
       );
--- a/packages/core/test/config.test.mjs
+++ b/packages/core/test/config.test.mjs
@@ -2613,6 +2613,62 @@ describe("role resolution", () => {
       );
     });
 
+    test("resolveGateConfig names every out-of-vocabulary value when several are invalid", () => {
+      const config = {
+        version: 1,
+        gates: { draft: { blockCleanOnFindingSeverities: ["critical", "blocker", "high"] } },
+      };
+      assert.throws(
+        () => resolveGateConfig(config, "draft"),
+        (err) => err.message.includes('"critical"') && err.message.includes('"blocker"') && !err.message.includes('"high"'),
+      );
+    });
+
+    test("resolveGateConfig refuses an entry JSON.stringify cannot render instead of crashing the renderer", () => {
+      const circular = {};
+      circular.self = circular;
+      for (const entry of [10n, circular]) {
+        assert.throws(
+          () => resolveGateConfig({ version: 1, gates: { draft: { blockCleanOnFindingSeverities: [entry] } } }, "draft"),
+          /out-of-vocabulary/,
+        );
+      }
+    });
+
+    // The schema requires min 1; an empty list reaches this resolver only
+    // through a config that failed validation, and passing it through would
+    // make the gate block on nothing (and severity consumers diverge on it).
+    test("resolveGateConfig throws on an explicitly-empty blockCleanOnFindingSeverities array", () => {
+      const config = {
+        version: 1,
+        gates: { draft: { blockCleanOnFindingSeverities: [] } },
+      };
+      assert.throws(() => resolveGateConfig(config, "draft"), /is empty/);
+    });
+
+    // The schema requires an array; a scalar or mapping here is the same
+    // failed-validation channel and previously fell back to the high default
+    // silently, under-blocking relative to the author's intent.
+    test("resolveGateConfig throws on a present-but-non-array blockCleanOnFindingSeverities value", () => {
+      for (const value of ["medium", { medium: true }, 3, null]) {
+        assert.throws(
+          () => resolveGateConfig({ version: 1, gates: { draft: { blockCleanOnFindingSeverities: value } } }, "draft"),
+          /must be an array/,
+          String(value),
+        );
+      }
+    });
+
+    // Parity pin for the derived vocabulary: every spelling the schema enum
+    // accepts must resolve without refusal, so the guard can never drift
+    // ahead of the schema.
+    test("resolveGateConfig accepts every schema-enum blockCleanOnFindingSeverities spelling", () => {
+      for (const spelling of ["high", "medium", "low", "must-fix", "worth-fixing-now", "nice-to-have", "defer"]) {
+        const result = resolveGateConfig({ version: 1, gates: { draft: { blockCleanOnFindingSeverities: [spelling] } } }, "draft");
+        assert.equal(result.blockCleanOnFindingSeverities.length, 1, spelling);
+      }
+    });
+
     // Case-sensitivity is deliberate (normalizeSeverity never lowercases):
     // a mixed-case spelling is out of vocabulary and must refuse, not coerce.
     test("resolveGateConfig throws on a mixed-case blockCleanOnFindingSeverities spelling", () => {
--- a/packages/core/test/config.test.mjs
+++ b/packages/core/test/config.test.mjs
@@ -2577,6 +2577,52 @@ describe("role resolution", () => {
       assert.deepEqual(result.blockCleanOnFindingSeverities, ["high", "medium"]);
     });
 
+    // Fail-closed posture: an out-of-vocabulary blocking severity can only
+    // arrive through a config that failed schema validation (the schema enum
+    // rejects it; loadDevLoopConfig returns the raw merged config alongside
+    // its errors). Passing it through would make the gate silently block on
+    // nothing, so the resolve boundary refuses instead.
+    test("resolveGateConfig throws on an out-of-vocabulary blockCleanOnFindingSeverities value, naming the value and the gate key", () => {
+      const config = {
+        version: 1,
+        gates: {
+          draft: {
+            angles: ["scope"],
+            blockCleanOnFindingSeverities: ["high", "critical"],
+          },
+        },
+      };
+      assert.throws(
+        () => resolveGateConfig(config, "draft"),
+        (err) =>
+          err instanceof Error &&
+          err.message.includes("Config validation failed") &&
+          err.message.includes("gates.draft.blockCleanOnFindingSeverities") &&
+          err.message.includes('"critical"'),
+      );
+    });
+
+    test("resolveGateConfig throws on a non-string blockCleanOnFindingSeverities entry", () => {
+      const config = {
+        version: 1,
+        gates: { preApproval: { blockCleanOnFindingSeverities: ["high", 3] } },
+      };
+      assert.throws(
+        () => resolveGateConfig(config, "preApproval"),
+        (err) => err instanceof Error && err.message.includes("gates.preApproval.blockCleanOnFindingSeverities"),
+      );
+    });
+
+    // Case-sensitivity is deliberate (normalizeSeverity never lowercases):
+    // a mixed-case spelling is out of vocabulary and must refuse, not coerce.
+    test("resolveGateConfig throws on a mixed-case blockCleanOnFindingSeverities spelling", () => {
+      const config = {
+        version: 1,
+        gates: { draft: { blockCleanOnFindingSeverities: ["HIGH"] } },
+      };
+      assert.throws(() => resolveGateConfig(config, "draft"), /out-of-vocabulary/);
+    });
+
     test("resolveGateConfig honors the deprecated worthFixingNowFixWindow alias when mediumFixWindow is absent", () => {
       const config = {
         version: 1,
--- a/packages/core/test/config.test.mjs
+++ b/packages/core/test/config.test.mjs
@@ -2627,12 +2627,34 @@ describe("role resolution", () => {
     test("resolveGateConfig refuses an entry JSON.stringify cannot render instead of crashing the renderer", () => {
       const circular = {};
       circular.self = circular;
-      for (const entry of [10n, circular]) {
+      const nullProtoCycle = Object.create(null);
+      nullProtoCycle.self = nullProtoCycle;
+      for (const entry of [10n, circular, nullProtoCycle]) {
         assert.throws(
           () => resolveGateConfig({ version: 1, gates: { draft: { blockCleanOnFindingSeverities: [entry] } } }, "draft"),
-          /out-of-vocabulary/,
+          /outside the schema's severity vocabulary/,
         );
       }
+    });
+
+    // JSON.stringify returns undefined (without throwing) for symbol values;
+    // the formatter's String() fallback must render a real token so the
+    // refusal message never carries an empty slot.
+    test("resolveGateConfig renders a non-empty token for a symbol entry in the refusal message", () => {
+      assert.throws(
+        () => resolveGateConfig({ version: 1, gates: { draft: { blockCleanOnFindingSeverities: [Symbol("oops")] } } }, "draft"),
+        (err) => err.message.includes("Symbol(oops)"),
+      );
+    });
+
+    // The guard's accept set is exact-spelling parity with the schema enum:
+    // a whitespace-varied entry fails schema validation, so it must refuse
+    // here too rather than being trim-normalized into acceptance.
+    test("resolveGateConfig refuses a whitespace-varied spelling the schema rejects", () => {
+      assert.throws(
+        () => resolveGateConfig({ version: 1, gates: { draft: { blockCleanOnFindingSeverities: ["high "] } } }, "draft"),
+        /outside the schema's severity vocabulary/,
+      );
     });
 
     // The schema requires min 1; an empty list reaches this resolver only
@@ -2676,7 +2698,7 @@ describe("role resolution", () => {
         version: 1,
         gates: { draft: { blockCleanOnFindingSeverities: ["HIGH"] } },
       };
-      assert.throws(() => resolveGateConfig(config, "draft"), /out-of-vocabulary/);
+      assert.throws(() => resolveGateConfig(config, "draft"), /outside the schema's severity vocabulary/);
     });
 
     test("resolveGateConfig honors the deprecated worthFixingNowFixWindow alias when mediumFixWindow is absent", () => {
--- a/scripts/github/upsert-checkpoint-verdict.mjs
+++ b/scripts/github/upsert-checkpoint-verdict.mjs
@@ -1626,6 +1626,17 @@ async function applyGateFullLabel({ repo, pr }, { env, ghCommand, runChild = def
 
 export async function upsertCheckpointVerdict(options, { env = process.env, ghCommand = "gh", repoRoot = process.cwd(), runChild = defaultRunChild } = {}) {
   const gh = { env, ghCommand, repoRoot, runChild };
+  // loadDevLoopConfig never throws: on a validation failure it returns the raw
+  // merged config alongside its errors. Every other severity consumer
+  // (consolidate-fanin, close-gate-findings, detect-checkpoint-evidence) fails
+  // closed on a non-empty errors array; the poster that WRITES the verdict
+  // must not be the one place a schema-invalid config still resolves gate
+  // config and posts. Checked before any GitHub read so the refusal is
+  // immediate and side-effect free.
+  const { config, errors: configErrors } = await loadDevLoopConfig({ repoRoot });
+  if (Array.isArray(configErrors) && configErrors.length > 0) {
+    throw new Error(`This worktree's config (--repo-root ${JSON.stringify(repoRoot)}) could not be fully loaded/validated; refusing to post a gate verdict: ${JSON.stringify(configErrors)}`);
+  }
   // Root cause 1: allow resurrected sessions to claim ownership when the previous
   // run's coordination record is stale. Without this, a new run ID is rejected even
   // though the old run is dead, forcing manual file deletion.
@@ -1647,7 +1658,6 @@ export async function upsertCheckpointVerdict(options, { env = process.env, ghCo
   const coordinationContext = await loadPrGateCoordinationContext({ repo: options.repo, pr: options.pr, lightweight: options.lightweight === true }, gh);
   const evidence = coordinationContext.gateEvidence;
   const canonicalHeadSha = resolveRequestedHeadSha(options.headSha, evidence.currentHeadSha);
-  const { config } = await loadDevLoopConfig({ repoRoot });
   const draftGateConfig = resolveGateConfig(config, "draft");
   const preApprovalGateConfig = resolveGateConfig(config, "preApproval");
   const maxCopilotRounds = options.lightweight === true
--- a/scripts/github/upsert-checkpoint-verdict.mjs
+++ b/scripts/github/upsert-checkpoint-verdict.mjs
@@ -1635,7 +1635,7 @@ export async function upsertCheckpointVerdict(options, { env = process.env, ghCo
   // immediate and side-effect free.
   const { config, errors: configErrors } = await loadDevLoopConfig({ repoRoot });
   if (Array.isArray(configErrors) && configErrors.length > 0) {
-    throw new Error(`This worktree's config (--repo-root ${JSON.stringify(repoRoot)}) could not be fully loaded/validated; refusing to post a gate verdict: ${JSON.stringify(configErrors)}`);
+    throw new Error(`This worktree's config (repoRoot ${JSON.stringify(repoRoot)}) could not be fully loaded/validated; refusing to post a gate verdict: ${JSON.stringify(configErrors)}`);
   }
   // Root cause 1: allow resurrected sessions to claim ownership when the previous
   // run's coordination record is stale. Without this, a new run ID is rejected even
--- a/test/github/upsert-checkpoint-verdict.test.mjs
+++ b/test/github/upsert-checkpoint-verdict.test.mjs
@@ -338,6 +338,36 @@ test("buildCoordinationEvaluatorInput threads postConvergenceReviewSuppressed fr
   assert.equal(missingFieldInput.postConvergenceReviewSuppressed, false);
 });
 
+// The poster is the artifact that WRITES the verdict; like its sibling
+// severity consumers (consolidate-fanin, close-gate-findings,
+// detect-checkpoint-evidence) it must fail closed on a config that failed
+// schema validation instead of resolving gate config from the raw merged
+// object. The refusal happens before any gh read, so no stub is needed.
+test("upsert refuses to post when the worktree config failed schema validation", async () => {
+  const tempDir = await mkdtemp(path.join(os.tmpdir(), "dev-loops-upsert-invalid-config-"));
+  try {
+    await writeFile(
+      path.join(tempDir, ".devloops"),
+      "version: 1\ngates:\n  draft:\n    blockCleanOnFindingSeverity:\n      - high\n",
+      "utf8",
+    );
+    const result = await runNodeHelper(scriptPath, [
+      "--repo", "owner/repo",
+      "--pr", "17",
+      "--gate", "draft_gate",
+      "--head-sha", "abc1234000000000000000000000000000000000",
+      "--verdict", "clean",
+      "--findings-summary", "no issues found",
+      "--next-action", "n/a",
+      "--inline-reason", "config refusal test",
+    ], { cwd: tempDir, env: runIdFreeEnv({ DEVLOOPS_RUN_ID: "" }) });
+    assert.notEqual(result.code, 0);
+    assert.match(result.stderr, /could not be fully loaded\/validated; refusing to post a gate verdict/);
+  } finally {
+    await rm(tempDir, { recursive: true, force: true });
+  }
+});
+
 test("parseUpsertCheckpointVerdictCliArgs rejects malformed arguments deterministically", () => {
   assert.throws(
     () => parseUpsertCheckpointVerdictCliArgs([]),
@@ -5261,6 +5291,8 @@ test("upsert-checkpoint-verdict refuses an inline pre_approval_gate verdict at p
     // lightMode disabled so the lazy light-facts fetch does not fire (no fetch
     // needed to prove the marker is evaluated); requireFanoutEvidence on so the
     // pre_approval_gate candidate marker IS evaluated by enforcePostTimeFanoutMode.
+    // maxFiles/maxLines are schema-required on a present lightMode object, and
+    // the poster now refuses outright on a schema-invalid config.
     await writeFile(path.join(tempDir, ".devloops"), [
       "version: 1",
       "gates:",
@@ -5268,6 +5300,8 @@ test("upsert-checkpoint-verdict refuses an inline pre_approval_gate verdict at p
       "localImplementation:",
       "  lightMode:",
       "    enabled: false",
+      "    maxFiles: 5",
+      "    maxLines: 200",
       "",
     ].join("\n"), "utf8");
     const env = await writeGhStub(tempDir, [


packages/core/src/config/config.mjs
```diff
@@ -219,6 +219,13 @@ const GateDynamicConfig = z.strictObject({
 // accepted but INERT for spike (a findings-doc deliverable has no "clean
 // verdict" escalation path and no additive dynamic pool) rather than being
 // split into a second schema.
+// Canonical defect-severity vocabulary for blockCleanOnFindingSeverities:
+// the post-normalization form of the schema enum below (legacy spellings
+// normalize into these three). resolveGateConfig fails closed against this
+// set so a config that bypassed schema validation cannot reach severity
+// consumers with an unknown value.
+const BLOCKING_SEVERITY_VOCABULARY = new Set(["high", "medium", "low"]);
```
mfittko: **medium** (`contract-surface`): BLOCKING_SEVERITY_VOCABULARY duplicates the blockCleanOnFindingSeverities schema enum by hand and its comment asserts it is the post-normalization form of that enum, but nothing derives or pins the relation. The enum at line 241 and the set at line 227 are independent literals. A future widening of the schema enum, for example adding a new defect spelling, leaves the set behind, and resolveGateConfig then throws on a config that PASSES schema validation. That converts a legitimate config into a hard refusal on every gate resolution path, so the drift fails closed against the operator rather than against bad input. — judge: act

packages/core/src/config/config.mjs
```diff
@@ -1839,9 +1846,30 @@ export function resolveRefinement(config) {
  * @param {DevLoopConfig} config
  * @param {"draft"|"preApproval"|"spike"} gate
  * @returns {{ angles: string[]|null, excludeAngles: string[], mandatoryAngles: string[], required: boolean, requireCi: boolean, blockCleanOnFindingSeverities: string[], dynamicAngles: boolean, additiveAngles: boolean, mediumFixWindow: number, tiers: Array<{name: string, match: object, angles: string[]}> }}
+ * @throws {Error} when a configured `blockCleanOnFindingSeverities` entry is
+ *   outside the defect vocabulary after normalization — such a value can only
+ *   arrive through a config that failed schema validation (the schema's enum
+ *   rejects it), and passing it through would make the gate silently block on
```
mfittko: **low** (`contract-surface`): The @throws rationale states that passing the value through would make the gate silently block on nothing. That sentence equally describes an explicitly-empty blockCleanOnFindingSeverities array, which is also schema-invalid via the min-one constraint at line 242 and also reachable only through a config that failed validation, yet the guard tolerates it: an empty array is truthy and an Array, normalizes to empty, yields no invalid entries, and resolves as a configured block-on-nothing set. PR 1759 declares that passthrough a non-goal, so the behavior is intended, but the docblock rationale as written promises coverage the guard does not give. — judge: act

packages/core/src/config/config.mjs
```diff
@@ -1839,9 +1846,30 @@ export function resolveRefinement(config) {
...
+ *   nothing. Refusing here, at the one resolve boundary every severity
```
mfittko: **low** (`contract-surface`): The @throws describes the refusal as sited at the one resolve boundary every severity consumer shares and as keeping unvalidated vocabulary out of every gate verdict path. The actual blast radius is wider than severity consumers and verdict paths: resolveRefinement at line 1825 calls resolveGateConfig purely for requireCi and now throws on a bad preApproval severity list, and resolveGateAngles, resolveGateTier, and scripts/loop/print-gates.mjs inherit the same throw while reading unrelated fields. A reader of this docblock would not expect state detection and angle resolution to refuse on this key. — judge: act

packages/core/src/config/config.mjs
```diff
@@ -1839,9 +1846,30 @@ export function resolveRefinement(config) {
  * @param {DevLoopConfig} config
  * @param {"draft"|"preApproval"|"spike"} gate
  * @returns {{ angles: string[]|null, excludeAngles: string[], mandatoryAngles: string[], required: boolean, requireCi: boolean, blockCleanOnFindingSeverities: string[], dynamicAngles: boolean, additiveAngles: boolean, mediumFixWindow: number, tiers: Array<{name: string, match: object, angles: string[]}> }}
+ * @throws {Error} when a configured `blockCleanOnFindingSeverities` entry is
```
mfittko: **low** (`contract-surface`): This module documents an explicit opposite convention for the same input channel, a raw merged config that zod rejected while still returning it: resolveMaxAnglesPerGroup at lines 2137 to 2139 and resolveFanoutGroups at lines 2200 to 2207 both state that such input degrades to a default and never crashes, and the second one calls that the sibling convention it mirrors. The new fail-closed throw is the reverse posture for a sibling gate key, and neither docblock reconciles with the other, so the module now teaches two contradictory rules for malformed gate config with no stated boundary between them. — judge: act

packages/core/src/config/config.mjs
mfittko: **low** (`correctness`): BLOCKING_SEVERITY_VOCABULARY hardcodes high, medium, low and is asserted to mirror the schema enum only by a code comment. Nothing derives it from the enum and no test pins parity, so widening the schema enum - a new canonical spelling, or a new legacy alias whose target is not one of the three - makes a schema-VALID config throw at every resolveGateConfig call site, which is a fail-closed break on legitimate input rather than on invalid input. The existing coverage does not catch that drift: the enum-completeness test at packages/core/test/config.test.mjs line 2686 only calls FileConfigSchema.safeParse, and the resolve-path tests exercise hand-picked spellings, so a widened enum fails no test. — judge: act

packages/core/src/config/config.mjs
```diff
@@ -1839,9 +1846,30 @@ export function resolveRefinement(config) {
...
+ *   rejects it), and passing it through would make the gate silently block on
+ *   nothing. Refusing here, at the one resolve boundary every severity
+ *   consumer shares, keeps unvalidated vocabulary out of every gate verdict
+ *   path. Absent keys still fall back to the defaults unchanged.
  */
 export function resolveGateConfig(config, gate) {
   const gateConfig = config?.gates?.[gate];
+  const rawBlocking = gateConfig?.blockCleanOnFindingSeverities;
+  let blockCleanOnFindingSeverities = ["high"];
+  if (rawBlocking && Array.isArray(rawBlocking)) {
```
mfittko: **medium** (`coverage`): The refusal is gated on Array.isArray, so a present-but-non-array blockCleanOnFindingSeverities silently falls back to the high default with no refusal and no test. That shape is exactly as schema-invalid as the array shapes the three new pins cover: the schema requires an array, loadDevLoopConfig still returns the raw merged config, and a scalar or mapping value such as blockCleanOnFindingSeverities: critical, or a YAML mapping written by mistake, resolves to high only. A config author who intended low and medium to block therefore keeps a gate that blocks on high alone, which is the same under-blocking class issue 1753 asks to refuse, and no test pins the branch either before or after this diff. — judge: act

packages/core/test/config.test.mjs
```diff
@@ -2577,6 +2577,52 @@ describe("role resolution", () => {
...
+        () => resolveGateConfig(config, "draft"),
+        (err) =>
+          err instanceof Error &&
+          err.message.includes("Config validation failed") &&
+          err.message.includes("gates.draft.blockCleanOnFindingSeverities") &&
+          err.message.includes('"critical"'),
+      );
+    });
+
+    test("resolveGateConfig throws on a non-string blockCleanOnFindingSeverities entry", () => {
```
mfittko, line 2605: **low** (`coverage`): The message-contract pins only exercise the single-invalid-value path. No test drives two or more out-of-vocabulary entries in one call, so the invalid.map plus join rendering that produces the plural list is unasserted, and no test asserts the Allowed hint text that tells an operator which spellings are legal. The non-string pin also asserts the key name only, not that the offending value is named, so a regression that dropped value naming for non-string entries would still pass. Gate-name parametrization itself is adequately covered by the draft and preApproval pins, so a separate spike-gate refusal test is not warranted. — judge: act

packages/core/src/config/config.mjs
```diff
@@ -1839,9 +1846,30 @@ export function resolveRefinement(config) {
...
+    blockCleanOnFindingSeverities = [...new Set(rawBlocking.map((s) => normalizeSeverity(s)))];
+    const invalid = blockCleanOnFindingSeverities.filter((s) => !BLOCKING_SEVERITY_VOCABULARY.has(s));
+    if (invalid.length > 0) {
+      throw new Error(
```
mfittko, line 1910: **low** (`determinism`): The refusal fires at the top of resolveGateConfig, before the unrelated angles / requireCi / mediumFixWindow / tiers resolution, so callers that resolve a gate only for those fields abort too, and callers disagree on WHICH gates they resolve eagerly: upsert-checkpoint-verdict.mjs lines 1651-1652, detect-pr-gate-coordination-state.mjs lines 951-952 and detect-checkpoint-evidence.mjs lines 639-640 resolve draft AND preApproval, while ready-for-review.mjs line 71, reconcile-draft-gate.mjs line 201, close-gate-findings.mjs line 325 and consolidate-fanin.mjs line 1152 resolve one gate only. For a config where only gates.preApproval.blockCleanOnFindingSeverities is out of vocabulary, a draft_gate round still consolidates fan-in and still flips ready-for-review, then throws at verdict posting: two call sites reaching opposite conclusions about the same config object, which leaves a written ledger artifact and no verdict rather than one uniform refusal. resolveRefinement at config.mjs line 1825 inherits the same throw for a severity defect it never reads. Severity is held at low only because reachability is narrow: applyLayer drops a whole schema-invalid file layer before merge, so an out-of-vocabulary value cannot arrive from a .devloops file, only from a programmatically built config object. — judge: defer

packages/core/src/config/config.mjs
```diff
@@ -1839,9 +1846,30 @@ export function resolveRefinement(config) {
...
+        `Config validation failed: gates.${gate}.blockCleanOnFindingSeverities contains ` +
+        `out-of-vocabulary severity value(s) after normalization: ${invalid.map((s) => JSON.stringify(s)).join(", ")}. ` +
```
mfittko: **low** (`determinism`): The message builder formats each offending entry with JSON.stringify, which itself throws for some non-string entries: a BigInt raises a TypeError, and so do a circular structure and a throwing toJSON. Non-string entries are an explicitly pinned class here, since one new test asserts the refusal for the numeric entry 3. In those cases the intended fail-closed Error is replaced by an unrelated TypeError with no Config validation failed prefix and no gate-key naming, so the refusal contract the tests pin is not stable across the whole non-string input class. Ordering itself is deterministic: the invalid list keeps first-occurrence input order and Set iteration is insertion-ordered, so the same config object always yields the same message. — judge: act

packages/core/src/config/config.mjs
mfittko: **nit** (`determinism`): BLOCKING_SEVERITY_VOCABULARY is module-level mutable state: a Set has no freeze, so nothing prevents a later edit in this module from adding to it and widening the vocabulary for every consumer at once. The neighbouring severity constant LEGACY_SEVERITY_ALIASES in gate-fanin.mjs uses Object.freeze for exactly that reason. Nothing mutates it today and it is not exported, so this is cosmetic. — judge: reject

packages/core/src/config/config.mjs
mfittko: **medium** (`input-validation`): The new guard admits an explicitly-empty blockCleanOnFindingSeverities array, which is just as schema-invalid as an out-of-vocabulary token: the schema enum array carries min 1, pinned by the existing test GateConfig rejects empty blockCleanOnFindingSeverities. An empty array reaches resolveGateConfig only through the same failed-validation path the guard exists to close, normalizes to an empty list, and passes. Downstream this is worse than an unknown token, not merely equivalent: in scripts/github/upsert-checkpoint-verdict.mjs the clean-verdict precondition at lines 1846 to 1872 is gated on length greater than zero, so an empty list skips the whole verification - no --findings-severity-counts is required and no unresolved-blocker check runs, letting a clean verdict be posted while high findings are open. consolidateFanin in packages/core/src/loop/gate-fanin.mjs lines 760 to 765 instead coerces an empty list to the default high list, so the two severity consumers disagree on the same config. The PR body lists the empty array as a non-goal, but the linked issue's non-goal covers only merely-absent keys, and the issue text asks for refusal on a config that failed validation wholesale. — judge: act

mfittko: **low** (`input-validation`): A non-array blockCleanOnFindingSeverities still fails open to the default high list rather than refusing. A YAML scalar such as blockCleanOnFindingSeverities: medium, or an object or number from a programmatic caller, is schema-invalid and arrives by exactly the raw-merged-config route the guard targets, yet the Array.isArray branch skips validation entirely and resolves the built-in default. The operator asked for a wider blocking set and silently gets a narrower one, which is the same silent-passthrough failure mode as an unknown token, only expressed as a shape error instead of a value error. The rawBlocking truthiness check in front of Array.isArray is also redundant, since Array.isArray already excludes null and undefined. — judge: act

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `ff9527cd31d6b9a66d19c082724229481254ad2a`
**Verdict:** clean

**Findings summary:** 8 angles reviewed; 17 findings (see per-angle breakdown below).

- `determinism` → `findings_present` (3 findings)
- `input-validation` → `findings_present` (3 findings)

> **low** (`correctness`): Now that resolveGateConfig normalizes and validates at the resolve boundary, this in-place re-normalization of activeGateConfig.blockCleanOnFindingSeverities is unreachable work and its comment is false: it states that legacy config spellings compare against the canonical vocabulary here, but both draftGateConfig and preApprovalGateConfig come from resolveGateConfig at lines 1651 and 1652, so only canonical values can arrive. The PR's own rationale is that the resolve boundary is the single chokepoint every severity consumer shares; leaving a second normalization site keeps alive the impression that consumers may still receive raw config values. Adjacent to the diff, not in it. — judge: defer
> Location: `scripts/github/upsert-checkpoint-verdict.mjs:1795`


> **nit** (`gate-evidence`): The validation record for this head stamps depState status stale with detail that node_modules/.package-lock.json is absent, so by run-gate-validation's own definition the recorded suites ran against installed deps that cannot be verified against the committed lock. The worktree lives inside the primary checkout, so Node resolved modules from the parent tree, which is why verify still exited 0 with 4040 passing tests. No dependency changed in this diff, and resolveDepState documents the stale stamp as informational and non-blocking, so this is recorded for transparency only and needs no fix on this PR. — judge: reject


> **medium** (`pr-description`): The PR body's first acceptance criterion over-claims relative to the diff and contradicts the body's own second non-goal. Acceptance criterion 1 asserts that a schema-invalid config reaching gate verdict or severity paths produces an explicit refusal, while non-goal 2 states that an explicitly-empty array still passes through as configured. An empty array IS schema-invalid because the enum array carries minItems 1, so both statements cannot hold. The diff refuses exactly one class, an entry outside the high, medium, low vocabulary after normalization, and every other schema validation failure still reaches severity consumers through the raw merged config that loadDevLoopConfig returns alongside its errors. A reader ticking that checkbox would record a broader guarantee than the code delivers. — judge: act

**Next action:** Apply the 13 judge-act fixes (empty-array and non-array refusal, vocabulary set derived from the schema enum, PR-body AC alignment, test additions), push, rerun the draft fan-out at the new head

packages/core/src/config/config.mjs
mfittko: **nit** (`input-validation`): The refusal message builds its offending-value list with JSON.stringify, which throws a TypeError for a BigInt entry and for a circular object entry. The call still fails closed, but the surfaced error is a serialization TypeError that names neither the offending value nor the gates key, so the acceptance criterion of an explicit refusal naming the validation failure is not met for those entry types. Reachable only from a programmatic caller, since a parsed config file cannot produce either shape. — judge: act

packages/core/src/config/config.mjs
mfittko: **medium** (`scope`): The fail-closed guard covers only out-of-vocabulary entries, so a schema-invalid empty array still reaches gate severity paths unrefused and produces divergent consumer behavior. The zod enum carries .min(1), so blockCleanOnFindingSeverities set to an empty list fails safeParse; loadDevLoopConfig then returns the raw merged config and resolveGateConfig resolves it to an empty list because invalid.length is 0. Downstream the two consumers disagree: consolidateFanin in packages/core/src/loop/gate-fanin.mjs line 761 falls back to high, while the clean-verdict guard in scripts/github/upsert-checkpoint-verdict.mjs line 1848 is skipped entirely on a zero-length list, so a clean verdict can be recorded with no findings-severity-counts verification at all. That is the same silently-blocks-on-nothing class issue 1753 asks to refuse. The PR body declares empty-array passthrough a non-goal, but issue 1753 sanctions only the merely-absent-key fallback as a non-goal, so this narrowing of acceptance criterion 1 is not covered by the issue's stated non-goals. — judge: act

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `04eb9685d73fdf0c586f21066b11a7e5238eb578`
**Verdict:** clean

**Findings summary:** 8 angles reviewed; 17 findings (see per-angle breakdown below).

- `coverage` → `findings_present` (5 findings)
- `determinism` → `findings_present` (3 findings)
- `gate-evidence` → `clean`
- `scope` → `clean`

> **medium** (`input-validation`): The verdict poster loads the config and discards loadDevLoopConfig's errors array, so a config that failed schema validation for any reason other than the newly guarded key still posts a verdict. Concrete case: a misspelled sibling key such as blockCleanOnFindingSeverity, singular, is rejected by the strict-object schema, yet it leaves the guarded key absent, so resolveGateConfig falls back to the high default and the gate blocks on a severity set the operator never configured. The new resolver guard closes only this one key's shape, so the linked issue's first acceptance criterion, an explicit refusal when a schema-invalid config reaches gate verdict and severity paths, holds only for that key on the most severity-critical path. Sibling severity consumers already fail closed on a non-empty errors array: scripts/loop/consolidate-fanin.mjs line 1148, scripts/github/close-gate-findings.mjs lines 320 to 325, scripts/github/detect-checkpoint-evidence.mjs lines 773 to 774. The poster is the only outlier, and it is the one that writes the verdict. — judge: act
> Location: `scripts/github/upsert-checkpoint-verdict.mjs:1650`


> **low** (`pr-description`): Issue 1753 asked the author to decide between two fail-closed postures: refuse on an out-of-vocabulary blockCleanOnFindingSeverities, or refuse on a config that failed schema validation wholesale. The PR implements only the key-scoped posture, and the body never records that decision. The Non-goals list four items, none of which excludes wholesale refusal, so a reader cannot tell whether the broader arm was deliberately rejected or overlooked. Round 1 fixed the AC over-claim about which shapes refuse; this remaining gap is about the posture choice itself, not the shapes. — judge: act


> **nit** (`pr-description`): The Validation section lists three commands but the Definition of done lists four items: npm run schema:check has no matching entry in the reproduction block. The command text is present in the DoD line, so nothing is unreproducible, only inconsistent between the two sections. — judge: act

**Next action:** Apply the 11 judge-act round-2 fixes (verdict poster fails closed on loadDevLoopConfig errors, docblock cluster, formatConfigValue cluster, PR body), push, rerun draft fan-out

packages/core/src/config/config.mjs
```diff
@@ -1839,9 +1860,52 @@ export function resolveRefinement(config) {
...
+        `the schema requires at least one blocking severity, and an empty list would make the gate block on nothing. ` +
+        `Fix the config before gate operations can proceed.`
+      );
+    }
+    const invalid = blockCleanOnFindingSeverities.filter((s) => !BLOCKING_SEVERITY_VOCABULARY.has(s));
+    if (invalid.length > 0) {
+      throw new Error(
+        `Config validation failed: gates.${gate}.blockCleanOnFindingSeverities contains ` +
+        `out-of-vocabulary severity value(s) after normalization: ${invalid.map((s) => formatConfigValue(s)).join(", ")}. ` +
+        `Allowed: high, medium, low (legacy spellings must-fix, worth-fixing-now, nice-to-have, defer normalize to these). ` +
```
mfittko: **low** (`contract-surface`): The refusal message still hand-enumerates the vocabulary it now derives: it prints a literal allowed list of high, medium, low plus the four legacy spellings. Round 1 flagged that hand-duplication in the guard set; the fix derived the set from BLOCKING_SEVERITY_SPELLINGS but moved the same duplication into the operator-facing string, and no test pins that string's allowed list. The module comment at line 224 claims widening the enum can never leave the runtime guard behind; that now holds for the check but not for the guidance it prints, so adding a canonical severity to BLOCKING_SEVERITY_SPELLINGS silently makes the refusal message name a stale allowed set. — judge: defer

packages/core/src/config/config.mjs
```diff
@@ -1839,9 +1860,52 @@ export function resolveRefinement(config) {
  * @param {DevLoopConfig} config
  * @param {"draft"|"preApproval"|"spike"} gate
  * @returns {{ angles: string[]|null, excludeAngles: string[], mandatoryAngles: string[], required: boolean, requireCi: boolean, blockCleanOnFindingSeverities: string[], dynamicAngles: boolean, additiveAngles: boolean, mediumFixWindow: number, tiers: Array<{name: string, match: object, angles: string[]}> }}
+ * @throws {Error} when a PRESENT `blockCleanOnFindingSeverities` key is any
+ *   schema-invalid shape: a non-array value, an empty array, or an entry
+ *   outside the defect vocabulary after normalization. Every such shape can
+ *   only arrive through a config that failed schema validation (the schema
+ *   requires a min-1 array of the enum spellings), and passing it through
+ *   would make the gate block on the wrong severities or on nothing at all.
+ *   The refusal deliberately reaches EVERY caller of this resolver — including
```
mfittko: **low** (`contract-surface`): The new @throws overstates the blast radius in the opposite direction from round 1's understatement. It says the refusal reaches EVERY caller of this resolver, but the guard reads only the blockCleanOnFindingSeverities key of the gate it is passed, so it reaches only callers resolving that same gate. Its own cited example makes the gap visible: resolveRefinement at line 1839 resolves preApproval only, so a schema-invalid draft key never throws there, while print-gates and reconcile-draft-gate do throw. A reader taking the docblock literally will expect one broken gate stanza to halt all gate operations. — judge: act

packages/core/src/config/config.mjs
```diff
@@ -1839,9 +1860,52 @@ export function resolveRefinement(config) {
...
+ *   ones reading unrelated fields (resolveRefinement's requireCi read, angle
+ *   and tier resolution, state detection) — because no gate operation should
+ *   proceed on a config whose verdict-semantics key failed validation. This
+ *   is the stated boundary with the module's degrade-to-default convention
```
mfittko: **nit** (`contract-surface`): The boundary statement labels the sibling convention degrade-to-default and cites resolveMaxAnglesPerGroup and resolveFanoutGroups together, but only the former defaults. resolveFanoutGroups drops malformed entries and, per its own docblock, degrades to a smaller grouping table, mirroring normalizeAngleEntries rather than substituting a default. The umbrella term does not match the sibling wording it is meant to reconcile with. — judge: act

packages/core/src/config/config.mjs
```diff
@@ -219,6 +219,27 @@ const GateDynamicConfig = z.strictObject({
...
+// enum below consumes the spellings verbatim, and the fail-closed set in
+// resolveGateConfig derives from the same list through normalizeSeverity, so
+// widening the enum can never leave the runtime guard behind (drift would
+// have refused configs that pass schema validation).
+const BLOCKING_SEVERITY_SPELLINGS = ["high", "medium", "low", "must-fix", "worth-fixing-now", "nice-to-have", "defer"];
+const BLOCKING_SEVERITY_VOCABULARY = new Set(BLOCKING_SEVERITY_SPELLINGS.map((s) => normalizeSeverity(s)));
+
+// Render an offending config value for a refusal message without letting the
+// renderer itself throw: JSON.stringify raises on BigInt and circular
+// structures and returns undefined for undefined/symbol/function, so those
```
mfittko: **nit** (`contract-surface`): The formatConfigValue docblock claims it renders a value without letting the renderer itself throw, then enumerates only the JSON.stringify failure modes. String is the fallback, so a value with a throwing toString still throws out of the helper, outside the try. Contrived for a parsed config, reachable only for a programmatic caller, and the outcome is a thrown error either way, so this is wording precision rather than a defect. — judge: act

packages/core/src/config/config.mjs
mfittko: **low** (`correctness`): The out-of-vocabulary refusal message hardcodes the allowed list as high, medium, low plus the four legacy spellings, far from its single source at line 227. The comment at line 222 states that widening the enum can never leave the runtime guard behind. That is true of the check itself, because BLOCKING_SEVERITY_VOCABULARY is derived through normalizeSeverity, but it is not true of this message: widening BLOCKING_SEVERITY_SPELLINGS would leave the refusal text telling an operator that a now-allowed severity is disallowed. The new parity test pins only the current seven spellings, so it would not catch that drift either, and the drifted text points the operator away from the fix. — judge: defer

packages/core/test/config.test.mjs
```diff
@@ -2577,6 +2577,108 @@ describe("role resolution", () => {
...
+          /must be an array/,
+          String(value),
+        );
+      }
+    });
+
+    // Parity pin for the derived vocabulary: every spelling the schema enum
+    // accepts must resolve without refusal, so the guard can never drift
+    // ahead of the schema.
+    test("resolveGateConfig accepts every schema-enum blockCleanOnFindingSeverities spelling", () => {
```
mfittko, line 2687: **low** (`coverage`): The schema-enum parity pin duplicates the seven spellings as a literal in the test instead of deriving them from the shared source, so it does not detect the drift it names. If someone widens the schema enum by passing a fresh literal list to z.enum while leaving the guard vocabulary behind, this loop still iterates its own stale seven and passes. It also asserts only that the resolved list has length 1, never the canonical value, so accept-without-normalizing is invisible to it in isolation. — judge: defer

packages/core/test/config.test.mjs
```diff
@@ -2577,6 +2577,108 @@ describe("role resolution", () => {
...
+        version: 1,
+        gates: { draft: { blockCleanOnFindingSeverities: ["critical", "blocker", "high"] } },
+      };
+      assert.throws(
+        () => resolveGateConfig(config, "draft"),
+        (err) => err.message.includes('"critical"') && err.message.includes('"blocker"') && !err.message.includes('"high"'),
+      );
+    });
+
+    test("resolveGateConfig refuses an entry JSON.stringify cannot render instead of crashing the renderer", () => {
```
mfittko, line 2627: **low** (`coverage`): The unrenderable-entry pin covers only the throwing render path of formatConfigValue, via BigInt and a circular object, and asserts nothing about the rendered token. The documented second fallback, where JSON.stringify returns undefined for undefined, symbol and function values, has no pin. A regression that dropped that ternary would return undefined from the formatter, and Array join coerces undefined to an empty string, so the refusal message would name no offending value at all while still matching this pin's out-of-vocabulary regex. That silently breaks the acceptance criterion that the refusal names the offending value. — judge: act

packages/core/test/config.test.mjs
```diff
@@ -2577,6 +2577,108 @@ describe("role resolution", () => {
...
+
+    // The schema requires min 1; an empty list reaches this resolver only
+    // through a config that failed validation, and passing it through would
+    // make the gate block on nothing (and severity consumers diverge on it).
+    test("resolveGateConfig throws on an explicitly-empty blockCleanOnFindingSeverities array", () => {
+      const config = {
+        version: 1,
+        gates: { draft: { blockCleanOnFindingSeverities: [] } },
+      };
+      assert.throws(() => resolveGateConfig(config, "draft"), /is empty/);
```
mfittko, line 2668: **low** (`coverage`): The empty-array pin matches only the phrase is empty and the non-array pin at line 2653 matches only must be an array. Neither asserts the interpolated gates dot gate dot blockCleanOnFindingSeverities key that PR 1759 describes for all three refusal shapes, and both use the draft gate, so a hardcoded gate name or a dropped key reference in those two messages passes the suite. Only the out-of-vocabulary message has its key naming pinned, and that one is pinned for two gates. — judge: defer

packages/core/src/config/config.mjs
mfittko: **low** (`coverage`): The docblock makes the blast radius a deliberate contract: the refusal reaches every caller of the resolver, explicitly naming resolveRefinement's requireCi read. That documented behavior has no pin. resolveRefinement calls resolveGateConfig for the preApproval gate at line 1839, so a schema-invalid preApproval severities list now throws out of resolveRefinement, a function whose callers read nothing severity-related. Nothing in the suite locks that, so narrowing the guard later, for example by catching inside the resolver for non-severity readers, would pass green while reverting the stated posture. — judge: reject

packages/core/test/config.test.mjs
```diff
@@ -2577,6 +2577,108 @@ describe("role resolution", () => {
...
+    test("resolveGateConfig accepts every schema-enum blockCleanOnFindingSeverities spelling", () => {
+      for (const spelling of ["high", "medium", "low", "must-fix", "worth-fixing-now", "nice-to-have", "defer"]) {
+        const result = resolveGateConfig({ version: 1, gates: { draft: { blockCleanOnFindingSeverities: [spelling] } } }, "draft");
+        assert.equal(result.blockCleanOnFindingSeverities.length, 1, spelling);
+      }
+    });
+
+    // Case-sensitivity is deliberate (normalizeSeverity never lowercases):
+    // a mixed-case spelling is out of vocabulary and must refuse, not coerce.
+    test("resolveGateConfig throws on a mixed-case blockCleanOnFindingSeverities spelling", () => {
```
mfittko, line 2696: **low** (`coverage`): The mixed-case pin fixes one coercion decision as deliberate: an uppercase spelling is refused rather than lowercased. The sibling coercion has no pin and behaves the opposite way. normalizeSeverity trims, so a whitespace-padded spelling, which the schema enum also rejects, is silently accepted and normalized instead of refused, contradicting the docblock claim that every schema-invalid shape of a present key is refused. With no test either way, that asymmetry reads as incidental rather than chosen. — judge: defer

packages/core/src/config/config.mjs
```diff
@@ -219,6 +219,27 @@ const GateDynamicConfig = z.strictObject({
...
+// fall back to String().
+function formatConfigValue(value) {
+  try {
+    const rendered = JSON.stringify(value);
+    return rendered === undefined ? String(value) : rendered;
+  } catch {
+    return String(value);
```
mfittko: **low** (`determinism`): formatConfigValue is documented as a renderer that cannot itself throw, but the String fallback sits inside the catch block. A value where BOTH JSON.stringify and String throw escapes as a raw TypeError instead of the deterministic Config validation failed refusal. Concrete cases: a null-prototype object that is also circular, since JSON.stringify throws on the cycle and String throws because a null prototype has no toString or valueOf; and any object whose toString or Symbol.toPrimitive throws. The docblock names circular structures as a case that falls back to String, which holds only for normal-prototype cycles. The pinned test covers 10n and a normal-prototype circular object, both of which String renders, so the gap is unpinned. Fail-closed posture survives because the call still throws, but the refusal message is not deterministic for all inputs: the same guard yields either a named refusal or a renderer crash depending on the offending value's prototype. — judge: act

packages/core/src/config/config.mjs
```diff
@@ -1839,9 +1860,52 @@ export function resolveRefinement(config) {
...
+  let blockCleanOnFindingSeverities = ["high"];
+  if (rawBlocking !== undefined) {
+    if (!Array.isArray(rawBlocking)) {
+      throw new Error(
+        `Config validation failed: gates.${gate}.blockCleanOnFindingSeverities must be an array ` +
+        `of defect severities, got ${formatConfigValue(rawBlocking)}. ` +
+        `Fix the config before gate operations can proceed.`
+      );
+    }
+    blockCleanOnFindingSeverities = [...new Set(rawBlocking.map((s) => normalizeSeverity(s)))];
```
mfittko: **low** (`determinism`): normalizeSeverity trims strings before the alias lookup, so the derived vocabulary check accepts whitespace-varied spellings such as high followed by a trailing space. The schema enum matches exactly and rejects that same value, so the guard's accept set is a strict superset of the schema's. A config that failed schema validation in exactly that way resolves silently rather than refusing, which is the passthrough the acceptance criterion targets. Blast radius is benign because the resolved value is already canonical and every downstream consumer normalizes identically, so no verdict diverges; the issue is that the module comment presents the derivation as exact schema parity when parity holds in one direction only, schema-accepted implies guard-accepted and not the reverse. — judge: act

packages/core/src/config/config.mjs
mfittko: **low** (`determinism`): The docblock states the refusal deliberately reaches EVERY caller of this resolver, but the guard reads only the requested gate's entry via config.gates at that gate key. A config whose draft blocking severities are schema-invalid therefore still resolves cleanly when the same config object is passed with preApproval or spike, so callers resolving a different gate proceed on a config that failed validation. Outcome is a pure function of the config plus gate pair, so there is no non-determinism, but the documented blast radius overstates the guard: the refusal is per-gate, not per-config, and the linked issue's wording about a config that failed validation wholesale reads broader than what ships. — judge: act

packages/core/src/config/config.mjs
mfittko: **low** (`input-validation`): formatConfigValue's catch falls back to String of the value, but String can itself throw for the same class of exotic input the try block defends against. A null-prototype object holding a cycle makes JSON.stringify throw, and then String throws a TypeError because no toString, valueOf, or Symbol.toPrimitive exists on the chain; a value with a throwing Symbol.toPrimitive behaves the same way. In that case the refusal path surfaces an opaque TypeError instead of the config-validation error this round was added to guarantee, which is the same failure shape the round-1 finding flagged. The comment at line 230 claims the renderer cannot throw, which overstates what the code guarantees. Not reachable from YAML config text, but reachable from programmatic callers that hand resolveGateConfig a settings object, for instance the handoff-envelope path at packages/core/src/loop/handoff-envelope.mjs line 387. — judge: act

packages/core/src/config/config.mjs
```diff
@@ -219,6 +219,33 @@ const GateDynamicConfig = z.strictObject({
...
+// fail-closed guard exact-matches raw entries against the same list, so the
+// guard's accept set is byte-identical to the schema's (no trim/normalize
+// superset) and widening the enum can never leave the runtime guard behind.
+const BLOCKING_SEVERITY_SPELLINGS = ["high", "medium", "low", "must-fix", "worth-fixing-now", "nice-to-have", "defer"];
+const BLOCKING_SEVERITY_SPELLING_SET = new Set(BLOCKING_SEVERITY_SPELLINGS);
+
+// Render an offending config value for a refusal message without letting the
+// renderer itself throw: JSON.stringify raises on BigInt and circular
+// structures and returns undefined for undefined/symbol/function (those fall
+// back to String()), and String() itself can throw for exotic values (a
```
mfittko, line 233: **nit** (`contract-surface`): formatConfigValue comment lists a throwing Symbol.toPrimitive object as an example that reaches the literal placeholder, but JSON.stringify renders such an object as an empty object literal and never calls String on it, so that example never takes the placeholder path. The null-prototype cycle example is accurate and the never-throws guarantee holds either way. — judge: reject

scripts/github/upsert-checkpoint-verdict.mjs
```diff
@@ -1626,6 +1626,17 @@ async function applyGateFullLabel({ repo, pr }, { env, ghCommand, runChild = def
 
 export async function upsertCheckpointVerdict(options, { env = process.env, ghCommand = "gh", repoRoot = process.cwd(), runChild = defaultRunChild } = {}) {
   const gh = { env, ghCommand, repoRoot, runChild };
+  // loadDevLoopConfig never throws: on a validation failure it returns the raw
+  // merged config alongside its errors. Every other severity consumer
+  // (consolidate-fanin, close-gate-findings, detect-checkpoint-evidence) fails
```
mfittko, line 1631: **low** (`correctness`): The new comment claims all three sibling severity consumers fail closed on a non-empty errors array, but only consolidate-fanin refuses. close-gate-findings deliberately falls back to the built-in MEDIUM_FIX_WINDOW and its own comment calls that behavior fails open to the historic behavior; detect-checkpoint-evidence sets config to null and leaves fan-out enforcement disabled to preserve default behavior. The sentence also frames the poster as the one place a schema-invalid config still resolves gate config, which overstates the convention it is aligning to. — judge: defer

test/github/upsert-checkpoint-verdict.test.mjs
```diff
@@ -338,6 +338,36 @@ test("buildCoordinationEvaluatorInput threads postConvergenceReviewSuppressed fr
...
+      "--pr", "17",
+      "--gate", "draft_gate",
+      "--head-sha", "abc1234000000000000000000000000000000000",
+      "--verdict", "clean",
+      "--findings-summary", "no issues found",
+      "--next-action", "n/a",
+      "--inline-reason", "config refusal test",
+    ], { cwd: tempDir, env: runIdFreeEnv({ DEVLOOPS_RUN_ID: "" }) });
+    assert.notEqual(result.code, 0);
+    assert.match(result.stderr, /could not be fully loaded\/validated; refusing to post a gate verdict/);
```
mfittko, line 365: **low** (`coverage`): The poster refusal test pins THAT the refusal happens, not that it happens before any side effect. The code comment claims the check is made before any GitHub read so the refusal is immediate and side-effect free, but the test asserts only a non-zero exit and the refusal text, writes no gh stub that would record invocations, and clears DEVLOOPS_RUN_ID via runIdFreeEnv so the pre-existing stale-runner takeover path, which claims a coordination record on disk before the old config load point, is never entered. Moving the guard back below that block and below loadPrGateCoordinationContext would leave the ordering regression invisible except by incidental real-gh failure. — judge: reject

packages/core/src/config/config.mjs
```diff
@@ -1839,9 +1866,57 @@ export function resolveRefinement(config) {
  * @param {DevLoopConfig} config
  * @param {"draft"|"preApproval"|"spike"} gate
  * @returns {{ angles: string[]|null, excludeAngles: string[], mandatoryAngles: string[], required: boolean, requireCi: boolean, blockCleanOnFindingSeverities: string[], dynamicAngles: boolean, additiveAngles: boolean, mediumFixWindow: number, tiers: Array<{name: string, match: object, angles: string[]}> }}
+ * @throws {Error} when THIS gate's PRESENT `blockCleanOnFindingSeverities`
+ *   key is any schema-invalid shape: a non-array value, an empty array, or an
+ *   entry that is not one of the schema enum's exact spellings. Every such
+ *   shape can only arrive through a config that failed schema validation (the
+ *   schema requires a min-1 array of the enum spellings; the guard
+ *   exact-matches raw entries against the same spelling list, so its accept
+ *   set is byte-identical to the schema's), and passing it through would make
```
mfittko, line 1875: **low** (`coverage`): The new docblock states a per-gate blast radius as an explicit contract: the refusal reaches every caller resolving the affected gate, while a caller resolving a DIFFERENT gate of the same config is untouched. No test pins either half of that boundary. The added cases each corrupt the same gate they then resolve, and no case corrupts the draft gate and resolves preApproval, or corrupts preApproval and asserts that resolveRefinement propagates the refusal through its preApproval requireCi read. A future change that hoisted validation across all gates, or that caught the throw inside resolveRefinement, would keep the whole suite green while breaking the documented boundary. — judge: reject

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `a48ed437528c6780efa6649b93b8c5f8c1d3ddf0`
**Verdict:** clean

**Findings summary:** 8 angles reviewed; 5 findings (see per-angle breakdown below).

- `determinism` → `clean`
- `gate-evidence` → `clean`
- `pr-description` → `clean`
- `scope` → `clean`

> **nit** (`contract-surface`): Cross-file comment interaction. write-gate-context deliberately warns on a config with validation errors and then proceeds with the merged fallback config, reasoning that nulling it out would empty the angle set. Its angle resolution routes through resolveGateAnglesDynamic to resolveGateConfig, so with a schema-invalid blockCleanOnFindingSeverities the run now aborts instead of proceeding. The abort is the posture PR 1759 intends, but the comment's proceed-with-fallback rationale no longer covers that key. — judge: defer
> Location: `scripts/github/write-gate-context.mjs:2566`

**Next action:** Mark the PR ready for review (round-3 residuals all judge-deferred or rejected; follow-ups filed at close-out)

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint ba3e64e747e8977b, severity low, angle contract-surface): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint 86fbf67c9486dda6, severity low, angle contract-surface): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint 54eade37d7d5042c, severity low, angle contract-surface): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint 171b43fab4cb7f7e, severity low, angle correctness): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/test/config.test.mjs
mfittko: Deferred at gate close (round 3, fingerprint c6f40bae0b223efd, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint 8e8e97f92483994c, severity low, angle determinism): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint 78dc1dab239a8676, severity low, angle determinism): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint 9ea6815b87439ae5, severity nit, angle determinism): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint 839f6934fc326e8c, severity low, angle input-validation): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint df636624e1a693a6, severity nit, angle input-validation): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint eb124352dda203f4, severity low, angle contract-surface): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint 87fcc3f62ced7cc2, severity low, angle contract-surface): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint e488054c7159e7c5, severity nit, angle contract-surface): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint 10e982e32356fcbe, severity nit, angle contract-surface): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint 7f7af120b5f278b5, severity low, angle correctness): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/test/config.test.mjs
mfittko: Deferred at gate close (round 3, fingerprint ba29078367a8ba0b, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/test/config.test.mjs
mfittko: Deferred at gate close (round 3, fingerprint 6d88cdda44d17372, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/test/config.test.mjs
mfittko: Deferred at gate close (round 3, fingerprint b491bbfe78c44cc1, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint c9cc7bd1c3002f54, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/test/config.test.mjs
mfittko: Deferred at gate close (round 3, fingerprint 7276acc8388d33d3, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint e22535d4bf0a7d13, severity low, angle determinism): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint 7ff1f9d123f1eba2, severity low, angle determinism): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint b9af80d046e3fe0d, severity low, angle determinism): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint 7b8ce54de4b23def, severity low, angle input-validation): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint 075acec7681ad94e, severity nit, angle contract-surface): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred at gate close (round 3, fingerprint c5358fe80a5709d3, severity low, angle correctness): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

test/github/upsert-checkpoint-verdict.test.mjs
mfittko: Deferred at gate close (round 3, fingerprint 92b4cd52c9c8d755, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 3, fingerprint 94235da8488b8dc4, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Fixed in commit 04eb9685d73fdf0c586f21066b11a7e5238eb578. BLOCKING_SEVERITY_VOCABULARY now derives from BLOCKING_SEVERITY_SPELLINGS, the same spelling list the schema enum consumes, via normalizeSeverity, instead of being a separately hand-written literal. Widening the schema enum can no longer leave the runtime guard behind.

packages/core/src/config/config.mjs
mfittko: Fixed in commit 04eb9685d73fdf0c586f21066b11a7e5238eb578, the same code change as the sibling input-validation finding on this file: resolveGateConfig now refuses an explicitly-empty blockCleanOnFindingSeverities array instead of letting downstream consumers diverge on it, since gate-fanin was defaulting the empty case to high while the verdict poster's clean-verdict check was being skipped entirely on a zero-length list.

mfittko: Fixed in commit 04eb9685d73fdf0c586f21066b11a7e5238eb578. resolveGateConfig now checks the present key first and throws an explicit Config validation failed refusal when the value is not an array, instead of silently falling back to the default high blocking set. A pinned test covers the non-array shape.

mfittko: Fixed in commit 04eb9685d73fdf0c586f21066b11a7e5238eb578. resolveGateConfig now throws an explicit refusal when the normalized blockCleanOnFindingSeverities array is empty, closing the same failed-validation channel as an out-of-vocabulary token instead of silently resolving to an empty blocking set.

scripts/github/upsert-checkpoint-verdict.mjs
```diff
@@ -1626,6 +1626,17 @@ async function applyGateFullLabel({ repo, pr }, { env, ghCommand, runChild = def
...
+  // closed on a non-empty errors array; the poster that WRITES the verdict
+  // must not be the one place a schema-invalid config still resolves gate
+  // config and posts. Checked before any GitHub read so the refusal is
+  // immediate and side-effect free.
+  const { config, errors: configErrors } = await loadDevLoopConfig({ repoRoot });
+  if (Array.isArray(configErrors) && configErrors.length > 0) {
+    throw new Error(`This worktree's config (--repo-root ${JSON.stringify(repoRoot)}) could not be fully loaded/validated; refusing to post a gate verdict: ${JSON.stringify(configErrors)}`);
```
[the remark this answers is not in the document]
mfittko: Confirmed: parseUpsertCheckpointVerdictCliArgs has no repo-root flag; the refusal message wrongly implied one. Fixed in commit 38e05a86 to name the resolved repoRoot value instead of a nonexistent CLI flag. Verified with node --test packages/core/test/config.test.mjs and node --test test/github/upsert-checkpoint-verdict.test.mjs (all passing).

packages/core/src/config/config.mjs
```diff
@@ -1839,9 +1866,57 @@ export function resolveRefinement(config) {
  * @param {DevLoopConfig} config
  * @param {"draft"|"preApproval"|"spike"} gate
  * @returns {{ angles: string[]|null, excludeAngles: string[], mandatoryAngles: string[], required: boolean, requireCi: boolean, blockCleanOnFindingSeverities: string[], dynamicAngles: boolean, additiveAngles: boolean, mediumFixWindow: number, tiers: Array<{name: string, match: object, angles: string[]}> }}
+ * @throws {Error} when THIS gate's PRESENT `blockCleanOnFindingSeverities`
+ *   key is any schema-invalid shape: a non-array value, an empty array, or an
+ *   entry that is not one of the schema enum's exact spellings. Every such
+ *   shape can only arrive through a config that failed schema validation (the
```
mfittko, line 1872: **low** (`contradiction-lens`): The added resolveGateConfig docblock says every schema-invalid shape can only arrive through a config that failed schema validation, which reads as the config-load channel delivering the raw invalid value to this resolver. applyLayer validates each layer with FileConfigSchema, and FileGatesConfig keeps GateConfig's severity enum and min-1 constraint through .partial, so a file layer carrying an invalid blockCleanOnFindingSeverities is dropped whole and the merged config keeps the previously merged valid value. The guard is therefore reachable only from a programmatic caller that hand-builds a config object, which is the same reachability the deferred refusal-timing follow-up already records. — judge: reject

packages/core/test/config.test.mjs
```diff
@@ -2577,6 +2577,130 @@ describe("role resolution", () => {
       assert.deepEqual(result.blockCleanOnFindingSeverities, ["high", "medium"]);
     });
 
+    // Fail-closed posture: an out-of-vocabulary blocking severity can only
```
mfittko, line 2580: **low** (`contradiction-lens`): The comment above the out-of-vocabulary pin makes two claims its own fixture contradicts. First, it says loadDevLoopConfig returns the raw merged config carrying the invalid value, but the whole-layer drop in applyLayer means the invalid value never reaches the merged config. Second, it says passthrough would make the gate silently block on nothing, while this test's own fixture of high plus critical would have blocked on high under the old passthrough. — judge: reject

packages/core/src/config/config.mjs
```diff
@@ -1839,9 +1866,57 @@ export function resolveRefinement(config) {
...
+ *   schema requires a min-1 array of the enum spellings; the guard
+ *   exact-matches raw entries against the same spelling list, so its accept
+ *   set is byte-identical to the schema's), and passing it through would make
+ *   the gate block on the wrong severities or on nothing at all. The refusal
```
mfittko, line 1876: **low** (`contradiction-lens`): The added docblock justifies the refusal by the gate blocking on the wrong severities or on nothing at all, and the guard fires for the spike gate as well. The pre-existing comment above the schema states that blockCleanOnFindingSeverities is accepted but INERT for spike, so the stated rationale cannot hold for that gate: an invalid spike list refuses spike angle, tier, and requireCi resolution over a key no spike consumer reads. — judge: reject

packages/core/src/config/config.mjs
```diff
@@ -219,6 +219,33 @@ const GateDynamicConfig = z.strictObject({
 // accepted but INERT for spike (a findings-doc deliverable has no "clean
 // verdict" escalation path and no additive dynamic pool) rather than being
 // split into a second schema.
+// Single source for the blockCleanOnFindingSeverities vocabulary: the schema
```
mfittko, line 222: **low** (`contradiction-lens`): The added vocabulary comment is appended with no blank line to the pre-existing comment that documents the unified draft, preApproval, and spike gate schema, and the two new declarations push const GateConfig 27 lines further down. The pre-existing block now reads as prose about BLOCKING_SEVERITY_SPELLINGS, its closing phrase about not being split into a second schema flows straight into the new text, and the new comment's phrase enum below now points past an intervening function definition. — judge: reject

packages/core/src/config/config.mjs
```diff
@@ -1839,9 +1866,57 @@ export function resolveRefinement(config) {
...
+ *   reaches every caller that resolves the affected gate — including ones
+ *   reading unrelated fields from it (resolveRefinement's preApproval
+ *   requireCi read, angle and tier resolution, state detection) — because no
+ *   gate operation should proceed on a config whose verdict-semantics key
+ *   failed validation; a caller resolving a DIFFERENT gate of the same config
+ *   is untouched. This is the stated boundary with the module's
+ *   degrade-quietly convention for dispatch-ergonomics keys
+ *   (resolveMaxAnglesPerGroup substitutes its default, resolveFanoutGroups
+ *   drops malformed entries): those keys only shape dispatch, so they degrade;
+ *   the key that decides what blocks a clean verdict refuses. An ABSENT key
```
mfittko, line 1886: **low** (`pr-checklist-matrix`): The docblock carve-out says an ABSENT key still falls back to the default, and the PR body non-goal is worded as merely-absent keys, but neither distinguishes a key written in YAML with no value. A bare blockCleanOnFindingSeverities: line parses to null, which the new guard treats as present-but-non-array and refuses with the must-be-an-array message, a behavior the new non-array test pins for null. An author who reads that YAML form as absent now gets a hard refusal of every operation on that gate instead of the documented high default. — judge: reject

test/github/upsert-checkpoint-verdict.test.mjs
```diff
@@ -5261,13 +5291,17 @@ test("upsert-checkpoint-verdict refuses an inline pre_approval_gate verdict at p
...
+    // maxFiles/maxLines are schema-required on a present lightMode object, and
+    // the poster now refuses outright on a schema-invalid config.
     await writeFile(path.join(tempDir, ".devloops"), [
       "version: 1",
       "gates:",
       "  requireFanoutEvidence: true",
       "localImplementation:",
       "  lightMode:",
       "    enabled: false",
+      "    maxFiles: 5",
```
mfittko, line 5303: **low** (`pr-checklist-matrix`): The PR body Scope section describes this test file as only pinning the poster's refusal, omitting that a pre-existing test's config fixture had to be repaired here with maxFiles and maxLines because the new wholesale refusal rejects any schema-invalid config. That repair is the concrete evidence of the change's real blast radius: a worktree whose config fails schema validation for any unrelated reason, such as a partial lightMode block, can no longer post a gate verdict at all. The body carries no rollout or risk note for that. — judge: act

scripts/github/upsert-checkpoint-verdict.mjs
mfittko, line 1631: **nit** (`pr-checklist-matrix`): The PR body Summary repeats the added comment's claim that all three sibling severity consumers already fail closed on a non-empty errors array. Only consolidate-fanin throws; close-gate-findings documents itself as failing open to its built-in medium fix window, and detect-checkpoint-evidence nulls the config and leaves fan-out enforcement disabled to preserve default behavior. Issue 1763 already tracks the code-comment half of this as a judge-deferred round-3 finding, so this is recorded only because the PR body sentence carries the same inaccurate precedent and is outside that issue's comment-only scope. — judge: act

packages/core/src/config/config.mjs
```diff
@@ -219,6 +219,33 @@ const GateDynamicConfig = z.strictObject({
...
+const BLOCKING_SEVERITY_SPELLINGS = ["high", "medium", "low", "must-fix", "worth-fixing-now", "nice-to-have", "defer"];
+const BLOCKING_SEVERITY_SPELLING_SET = new Set(BLOCKING_SEVERITY_SPELLINGS);
+
+// Render an offending config value for a refusal message without letting the
+// renderer itself throw: JSON.stringify raises on BigInt and circular
+// structures and returns undefined for undefined/symbol/function (those fall
+// back to String()), and String() itself can throw for exotic values (a
+// null-prototype cycle, a throwing Symbol.toPrimitive) — those get a literal
+// placeholder so the refusal always surfaces as the refusal.
+function formatConfigValue(value) {
```
mfittko, line 236: **low** (`yagni`): formatConfigValue nests two fallback levels for value shapes no config channel produces. A value reaches resolveGateConfig either through validated config layers, where an invalid severity list is dropped with the whole layer, or from a programmatic caller, and YAML parsing yields no BigInt, symbol, or null-prototype value at all. The inner catch's unrenderable-value placeholder is asserted by no test: the two exotic-value pins only match the vocabulary sentence, so the nesting level they justify is unverified. — judge: reject

packages/core/src/config/config.mjs
```diff
@@ -219,6 +219,33 @@ const GateDynamicConfig = z.strictObject({
 // accepted but INERT for spike (a findings-doc deliverable has no "clean
 // verdict" escalation path and no additive dynamic pool) rather than being
 // split into a second schema.
+// Single source for the blockCleanOnFindingSeverities vocabulary: the schema
+// enum below consumes these spellings verbatim, and resolveGateConfig's
+// fail-closed guard exact-matches raw entries against the same list, so the
+// guard's accept set is byte-identical to the schema's (no trim/normalize
+// superset) and widening the enum can never leave the runtime guard behind.
+const BLOCKING_SEVERITY_SPELLINGS = ["high", "medium", "low", "must-fix", "worth-fixing-now", "nice-to-have", "defer"];
+const BLOCKING_SEVERITY_SPELLING_SET = new Set(BLOCKING_SEVERITY_SPELLINGS);
```
mfittko, line 228: **nit** (`yagni`): BLOCKING_SEVERITY_SPELLING_SET is a second module-level binding for a seven-element membership test, and it is the only consumer of the array it derives from besides the schema enum. — judge: reject

mfittko (commented):


**Reviewed head SHA:** `38e05a86492683a00b1414f3e59e477cedd80115`
**Verdict:** clean

**Findings summary:** 5 angles reviewed; 11 findings (see per-angle breakdown below).

- `contradiction-lens` → `findings_present` (4 findings)
- `pr-checklist-matrix` → `findings_present` (4 findings)
- `yagni` → `findings_present` (3 findings)

> **low** (`pr-checklist-matrix`): The AC / DoD / Non-goals matrix mis-anchors two evidence cells. Row 3, Refusal test pinned, cites schema:check green as verification; schema:check only re-generates and compares the JSON schema and cannot verify that a refusal test exists. Row 1 labels the upsert suite as evidence that consumers are unaffected, but that suite pins the poster's own wholesale refusal, which is arm 2 of the change rather than consumer non-regression. — judge: act


> **nit** (`yagni`): Same dead re-normalization class as the already-deferred poster item: the light-mode escalation scan re-maps resolveGateConfig's already-normalized blockCleanOnFindingSeverities through normalizeSeverity, which this PR's resolve-boundary normalization makes unreachable. The line is untouched by this PR, so this only widens the file list of the existing dead-re-normalization follow-up beyond the poster. — judge: reject
> Location: `packages/core/src/config/config.mjs:2130`

**Next action:** Apply the three judge-act PR-body corrections, then merge under the recorded standing authorization once gate-evidence status is green

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 1, fingerprint 9946a4c890a57e5d, severity low, angle contradiction-lens): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/test/config.test.mjs
mfittko: Deferred at gate close (round 1, fingerprint 37b81bfc6e3a41c2, severity low, angle contradiction-lens): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 1, fingerprint fe19a247bd8836d0, severity low, angle contradiction-lens): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 1, fingerprint 41e73c5626c9cb49, severity low, angle contradiction-lens): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 1, fingerprint e8c69cd37e6e0460, severity low, angle pr-checklist-matrix): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

test/github/upsert-checkpoint-verdict.test.mjs
mfittko: Deferred at gate close (round 1, fingerprint 3c3a1d1ce219bebc, severity low, angle pr-checklist-matrix): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/upsert-checkpoint-verdict.mjs
mfittko: Deferred at gate close (round 1, fingerprint 5bb68fa8bba17381, severity nit, angle pr-checklist-matrix): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 1, fingerprint d75e1c360ed5bd97, severity low, angle yagni): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/config/config.mjs
mfittko: Deferred at gate close (round 1, fingerprint 096f22216be49e73, severity nit, angle yagni): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

PR #1765, opened 2026-08-19
fix(guard): refuse the named hash entity; complete fallback boundary parity

### LINKED ISSUE #1755: comment-id guard: decide named-entity hash coverage; complete fallback-suite boundary parity (deferred from the guard hardening drive)
## Summary

Judge-deferred from the comment-id-guard hardening drive: the encoded-hash refusal covers decimal and hex NUMERIC character references (zero-padding included), but the docblock/AC wording says "an entity decoding to the hash character" — and the num named entity is a real HTML5 entity for the hash code point that GitHub also decodes. Decide and implement: either extend the refusal matcher to the named-entity form (both matcher copies plus tests), or narrow the documented claim to numeric references and record the named-entity form as out-of-threat-model with a reason.

Also fold in the fallback-suite boundary parity item deferred earlier: the fallback poster's synced guard copy should pin all three non-entity boundary shapes its docblock names, mirroring the core suite.

## Acceptance criteria

- [x] The named-entity hash form is either refused (with tests in both suites) or explicitly documented as out of scope with rationale in the guard docblock.
- [x] The fallback suite pins the same boundary shapes as the core suite (well-formed excluded; ampersand-no-semicolon refused; semicolon-no-ampersand refused; per-occurrence).
- [x] Both matcher copies stay semantically identical.

## Definition of done

- [x] node --test packages/core/test/comment-id-guard.test.mjs green.
- [x] node --test skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs green.
- [x] npm run verify green.

## Non-goals

- No change to the sanitizer or the allowlist contract.

mfittko:
## Summary

The comment-id guard's extraction is now decode-aware on both sides of the token. The issue asked for a decision between extending the matcher to the named hash entity and narrowing the documented claim: extended, because the named entity is a real HTML5 entity GitHub decodes into a live auto-link. The pre-approval gate's renderer-security review then found the sibling gap on the digit side — an id assembled from entity-encoded digits (a literal or encoded hash followed by numeric character references for the digits) also renders as an auto-link — so extraction now also scans the body after a single renderer-like entity decode (numeric character references plus the named hash entity; the single non-rescanning replace means a double-encoded form's output is never re-read). The fallback suite also gains the boundary-parity pins the core suite already had.

## Scope and context

Four files, two matcher copies kept semantically identical. `packages/core/src/github/comment-id-guard.mjs`: extraction scans the body as written AND after a single left-to-right decode of the entity forms the renderer resolves (`decodeRenderedText` over numeric character references and the named hash entity; undecodable references become the replacement character as on GitHub; the replace never rescans its own output, so a double-encoded form is never re-read as a fresh entity opener). The well-formed-entity exclusion applies only to the raw scan and is bounded at cmark-gfm's 8 digits — in decoded text the renderer's one decode is spent, so an ampersand-wrapped assembled id refuses rather than hides, and a 9-digit wrapped run (which the renderer does not decode) refuses too. This replaces the hash-side-only encoded pattern with one mechanism covering encoded hash, encoded digits, and every mix; the numeric bounds match cmark-gfm's 8-digit entity parser so nothing GitHub decodes escapes the pass, and named-entity case-variants decode too, as deliberate fail-closed over-refusal. `skills/dev-loop/scripts/post-gate-verdict-fallback.mjs`: same decode-aware extraction in the zero-dep inline copy, mirror comments. `packages/core/test/comment-id-guard.test.mjs`: named-entity refusal pin (decode form, case-variant over-refusal, inert no-digit form, unrelated-entity non-match, allowlist passthrough), digit-side assembly pins (literal hash plus encoded digit, encoded hash plus mixed literal/encoded digits, named hash plus encoded digit, allowlist through the assembled form), a double-encoded pin documenting that the single-pass decode never manufactures a new refusal (the numeric double-encoded form stays refused by the pre-existing raw scan; the named form stays inert), ampersand-wrapped assembly refusal pins, and exclusion-bound pins (9-digit wrapped run refuses; an undecodable 8-digit reference stays inert as the replacement character). `skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs`: named-entity refusal plus the missing boundary-parity pins — semicolon-without-ampersand refused, per-occurrence exclusion (same digit run well-formed in one place and bare in another still refuses), named-form case-variant over-refusal, and inert non-match shapes (no-digit named form, unrelated named entity followed by digits, both posting successfully); well-formed exclusion and ampersand-without-semicolon were already pinned.

## Acceptance criteria

- [ ] The named-entity hash form is refused, with tests in both suites (the extend arm of the issue's decision, rationale recorded above).
- [ ] The fallback suite pins the same boundary shapes as the core suite: well-formed excluded; ampersand-no-semicolon refused; semicolon-no-ampersand refused; per-occurrence.
- [ ] Both matcher copies stay semantically identical (same pattern, mirror comments).

## Definition of done

- [ ] `node --test packages/core/test/comment-id-guard.test.mjs` green.
- [ ] `node --test skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs` green.
- [ ] `npm run verify` green (its legs include the asset-sync check; no standalone evidence claim beyond the recorded gate validation).

|---|---|---|
| Named-entity form refused in both copies | core suite green (named-entity pin); fallback suite green (named-entity pin) | sanitizer and allowlist contract untouched |
| Fallback boundary parity | fallback suite green (all four boundary pins, named-form over-refusal, inert non-match, and digit-side assembly parity pins) | — |
| Matcher copies semantically identical | identical extraction code and patterns in both copies; each suite green on its own pins (the core suite pins more shapes than the fallback suite); npm run verify green | — |

## Non-goals

- No change to the sanitizer or the allowlist contract.
- No general HTML entity decoding: the decode pass covers exactly the forms that can assemble a rendered hash-digit auto-link (numeric character references at cmark-gfm's 8-digit bound, plus the named hash entity); arbitrary named entities stay untouched.

## Open questions

None.

## Validation

```sh
node --test packages/core/test/comment-id-guard.test.mjs
node --test skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs
npm run verify
```

Closes #1755

### DIFF
--- a/packages/core/src/github/comment-id-guard.mjs
+++ b/packages/core/src/github/comment-id-guard.mjs
@@ -26,8 +26,9 @@
  * skipped; the same digit run appearing elsewhere as a bare token still
  * refuses — this is the ONLY sanctioned way a generated comment body may
  * carry a `#<digits>` token. An entity that itself decodes to the hash
- * character followed by a digit run is treated as a bare id (it renders as
- * an auto-link). Keep the allowlist small and deliberate.
+ * character — numeric (`&#35;`/`&#x23;`, zero-padded variants included) or
+ * the named `&num;` — followed by a digit run is treated as a bare id (it
+ * renders as an auto-link). Keep the allowlist small and deliberate.
  */
 
 // Matches a bare GitHub auto-link issue/PR reference: `#<digits>`. Bound to
@@ -44,13 +45,17 @@ function isNumericCharacterReference(body, match) {
   return body[match.index - 1] === "&" && body[match.index + match[0].length] === ";";
 }
 
-// An entity that DECODES to the hash character (`&#35;` decimal or `&#x23;`
-// hex, case-insensitive, zero-padding accepted — CommonMark decodes padded
-// forms too), immediately followed by an issue-length digit run, renders as a
-// bare `#<digits>` auto-link on GitHub even though no literal hash-digit token
+// An entity that DECODES to the hash character (`&#35;` decimal, `&#x23;`
+// hex — case-insensitive, zero-padding accepted, CommonMark decodes padded
+// forms too — or the HTML5 named entity `&num;`, which GitHub also decodes),
+// immediately followed by an issue-length digit run, renders as a bare
+// `#<digits>` auto-link on GitHub even though no literal hash-digit token
 // exists in the raw body. Treated as a bare-id occurrence so the encoded form
-// cannot smuggle a reference past the guard.
-const ENCODED_HASH_ID_RE = /&#(?:0*35|x0*23);(\d{1,9})/gi;
+// cannot smuggle a reference past the guard. The `i` flag makes the named
+// form match case-variants like `&NUM;` too; only lowercase `&num;` actually
+// decodes, so that is deliberate over-refusal on a suspicious near-miss,
+// keeping the guard fail-closed.
+const ENCODED_HASH_ID_RE = /&(?:#(?:0*35|x0*23)|num);(\d{1,9})/gi;
 
 /**
  * Extract the raw issue/PR id tokens found in a body (as strings, deduped).
--- a/packages/core/src/github/comment-id-guard.mjs
+++ b/packages/core/src/github/comment-id-guard.mjs
@@ -53,7 +53,12 @@
 const ISSUE_PR_ID_RE = /#(\d{1,9})/g;
 
 function isNumericCharacterReference(body, match) {
-  return body[match.index - 1] === "&" && body[match.index + match[0].length] === ";";
+  // The digit bound mirrors cmark-gfm's 8-digit entity parser: a 9-digit
+  // ampersand-wrapped run is NOT decoded by the renderer, so it renders
+  // literally and must not be excluded as a spent entity.
+  return match[1].length <= 8
+    && body[match.index - 1] === "&"
+    && body[match.index + match[0].length] === ";";
 }
 
 // Entity forms the renderer resolves that can participate in assembling a
@@ -73,14 +78,17 @@ function decodeRenderedText(body) {
     try {
       return String.fromCodePoint(code);
     } catch {
-      return entity;
+      // cmark substitutes the replacement character for a reference it cannot
+      // decode; mirroring that keeps decodable-shaped text from lingering in
+      // the decoded scan, where it could masquerade as a spent entity.
+      return "�";
     }
   });
 }
 
-function collectBareIds(text, found) {
+function collectBareIds(text, found, { excludeEntities }) {
   for (const m of text.matchAll(ISSUE_PR_ID_RE)) {
-    if (isNumericCharacterReference(text, m)) continue;
+    if (excludeEntities && isNumericCharacterReference(text, m)) continue;
     found.add(m[1]);
   }
 }
@@ -88,15 +96,19 @@ function collectBareIds(text, found) {
 /**
  * Extract the raw issue/PR id tokens found in a body (as strings, deduped).
  * Scans the body as written AND after a single renderer-like entity decode,
- * so an id assembled from entity-encoded pieces is still found. Returns []
- * for non-string input (and for a body with no `#<digits>`).
+ * so an id assembled from entity-encoded pieces is still found. The entity
+ * exclusion applies only to the RAW scan: in decoded text the renderer's one
+ * decode is already spent, so an ampersand-then-digits-then-semicolon shape
+ * there is plain text a wrapper cannot re-protect (an ampersand-wrapped
+ * encoded hash plus digits must refuse, not hide). Returns [] for non-string
+ * input (and for a body with no `#<digits>`).
  */
 export function extractIssuePrIds(body) {
   if (typeof body !== "string" || body.length === 0) return [];
   const found = new Set();
-  collectBareIds(body, found);
+  collectBareIds(body, found, { excludeEntities: true });
   const decoded = decodeRenderedText(body);
-  if (decoded !== body) collectBareIds(decoded, found);
+  if (decoded !== body) collectBareIds(decoded, found, { excludeEntities: false });
   return [...found];
 }
 
--- a/packages/core/src/github/comment-id-guard.mjs
+++ b/packages/core/src/github/comment-id-guard.mjs
@@ -25,10 +25,17 @@
  * reference (`&#<digits>;`, e.g. `&#91;` for `[`) — each such OCCURRENCE is
  * skipped; the same digit run appearing elsewhere as a bare token still
  * refuses — this is the ONLY sanctioned way a generated comment body may
- * carry a `#<digits>` token. An entity that itself decodes to the hash
- * character — numeric (`&#35;`/`&#x23;`, zero-padded variants included) or
- * the named `&num;` — followed by a digit run is treated as a bare id (it
- * renders as an auto-link). Keep the allowlist small and deliberate.
+ * carry a `#<digits>` token. Extraction is decode-aware on BOTH sides of the
+ * token: the body is also scanned after a single left-to-right decode of the
+ * entity forms GitHub's renderer resolves (numeric character references,
+ * zero-padding and hex included, plus the named hash and ampersand entities),
+ * so a hash or any digit of the id smuggled as an entity — `&#35;123`,
+ * `&num;123`, `#&#49;23`, any mix — still refuses. The decode is single-pass
+ * like the renderer's, so a double-encoded form (`&amp;#35;123`), which
+ * renders as inert literal text, is NOT refused. Case-variants of the named
+ * entities are decoded too even where GitHub would not (`&NUM;`): deliberate
+ * over-refusal on a suspicious near-miss, keeping the guard fail-closed.
+ * Keep the allowlist small and deliberate.
  */
 
 // Matches a bare GitHub auto-link issue/PR reference: `#<digits>`. Bound to
@@ -45,32 +52,47 @@ function isNumericCharacterReference(body, match) {
   return body[match.index - 1] === "&" && body[match.index + match[0].length] === ";";
 }
 
-// An entity that DECODES to the hash character (`&#35;` decimal, `&#x23;`
-// hex — case-insensitive, zero-padding accepted, CommonMark decodes padded
-// forms too — or the HTML5 named entity `&num;`, which GitHub also decodes),
-// immediately followed by an issue-length digit run, renders as a bare
-// `#<digits>` auto-link on GitHub even though no literal hash-digit token
-// exists in the raw body. Treated as a bare-id occurrence so the encoded form
-// cannot smuggle a reference past the guard. The `i` flag makes the named
-// form match case-variants like `&NUM;` too; only lowercase `&num;` actually
-// decodes, so that is deliberate over-refusal on a suspicious near-miss,
-// keeping the guard fail-closed.
-const ENCODED_HASH_ID_RE = /&(?:#(?:0*35|x0*23)|num);(\d{1,9})/gi;
+// Entity forms the renderer resolves that can participate in assembling a
+// rendered `#<digits>` auto-link: numeric character references (any code
+// point — the hash AND the digits themselves are smuggleable) plus the named
+// hash and ampersand entities. `&amp;` is decoded (consumed) precisely so a
+// double-encoded form's output is never re-read as a fresh entity opener —
+// one pass, like the renderer.
+const DECODABLE_ENTITY_RE = /&(?:#(?:\d{1,7}|x[0-9a-f]{1,6})|num|amp);/gi;
+
+function decodeRenderedText(body) {
+  return body.replace(DECODABLE_ENTITY_RE, (entity) => {
+    const inner = entity.slice(1, -1).toLowerCase();
+    if (inner === "num") return "#";
+    if (inner === "amp") return "&";
+    const code = inner[1] === "x" ? Number.parseInt(inner.slice(2), 16) : Number.parseInt(inner.slice(1), 10);
+    try {
+      return String.fromCodePoint(code);
+    } catch {
+      return entity;
+    }
+  });
+}
+
+function collectBareIds(text, found) {
+  for (const m of text.matchAll(ISSUE_PR_ID_RE)) {
+    if (isNumericCharacterReference(text, m)) continue;
+    found.add(m[1]);
+  }
+}
 
 /**
  * Extract the raw issue/PR id tokens found in a body (as strings, deduped).
- * Returns [] for non-string input (and for a body with no `#<digits>`).
+ * Scans the body as written AND after a single renderer-like entity decode,
+ * so an id assembled from entity-encoded pieces is still found. Returns []
+ * for non-string input (and for a body with no `#<digits>`).
  */
 export function extractIssuePrIds(body) {
   if (typeof body !== "string" || body.length === 0) return [];
   const found = new Set();
-  for (const m of body.matchAll(ISSUE_PR_ID_RE)) {
-    if (isNumericCharacterReference(body, m)) continue;
-    found.add(m[1]);
-  }
-  for (const m of body.matchAll(ENCODED_HASH_ID_RE)) {
-    found.add(m[1]);
-  }
+  collectBareIds(body, found);
+  const decoded = decodeRenderedText(body);
+  if (decoded !== body) collectBareIds(decoded, found);
   return [...found];
 }
 
--- a/packages/core/src/github/comment-id-guard.mjs
+++ b/packages/core/src/github/comment-id-guard.mjs
@@ -28,14 +28,18 @@
  * carry a `#<digits>` token. Extraction is decode-aware on BOTH sides of the
  * token: the body is also scanned after a single left-to-right decode of the
  * entity forms GitHub's renderer resolves (numeric character references,
- * zero-padding and hex included, plus the named hash and ampersand entities),
+ * zero-padding and hex included at cmark-gfm's 8-digit bound, plus the named
+ * hash entity),
  * so a hash or any digit of the id smuggled as an entity — `&#35;123`,
  * `&num;123`, `#&#49;23`, any mix — still refuses. The decode is single-pass
- * like the renderer's, so a double-encoded form (`&amp;#35;123`), which
- * renders as inert literal text, is NOT refused. Case-variants of the named
- * entities are decoded too even where GitHub would not (`&NUM;`): deliberate
- * over-refusal on a suspicious near-miss, keeping the guard fail-closed.
- * Keep the allowlist small and deliberate.
+ * like the renderer's: a double-encoded form (`&amp;#35;123`) renders as
+ * inert literal text and the decode pass never manufactures a refusal of its
+ * INNER id — though the raw scan still refuses the outer digit run of the
+ * numeric form as a pre-existing fail-closed near-miss (the outer entity's
+ * semicolon sits before the hash, so the well-formed-entity exclusion does
+ * not apply). Case-variants of the named hash entity are decoded too even
+ * where GitHub would not (`&NUM;`): deliberate over-refusal, keeping the
+ * guard fail-closed. Keep the allowlist small and deliberate.
  */
 
 // Matches a bare GitHub auto-link issue/PR reference: `#<digits>`. Bound to
@@ -55,16 +59,16 @@ function isNumericCharacterReference(body, match) {
 // Entity forms the renderer resolves that can participate in assembling a
 // rendered `#<digits>` auto-link: numeric character references (any code
 // point — the hash AND the digits themselves are smuggleable) plus the named
-// hash and ampersand entities. `&amp;` is decoded (consumed) precisely so a
-// double-encoded form's output is never re-read as a fresh entity opener —
-// one pass, like the renderer.
-const DECODABLE_ENTITY_RE = /&(?:#(?:\d{1,7}|x[0-9a-f]{1,6})|num|amp);/gi;
+// hash entity. The digit bounds match cmark-gfm's numeric-entity parser
+// (up to 8 digits, decimal or hex) so nothing GitHub decodes escapes the
+// pass. Single non-rescanning replace = one decode, like the renderer, so a
+// double-encoded form's output is never re-read as a fresh entity.
+const DECODABLE_ENTITY_RE = /&(?:#(?:\d{1,8}|x[0-9a-f]{1,8})|num);/gi;
 
 function decodeRenderedText(body) {
   return body.replace(DECODABLE_ENTITY_RE, (entity) => {
     const inner = entity.slice(1, -1).toLowerCase();
     if (inner === "num") return "#";
-    if (inner === "amp") return "&";
     const code = inner[1] === "x" ? Number.parseInt(inner.slice(2), 16) : Number.parseInt(inner.slice(1), 10);
     try {
       return String.fromCodePoint(code);
--- a/skills/dev-loop/scripts/post-gate-verdict-fallback.mjs
+++ b/skills/dev-loop/scripts/post-gate-verdict-fallback.mjs
@@ -170,17 +170,16 @@ function collapseWhitespace(value) {
 // reference — preceded by `&` AND immediately followed by `;` (e.g. `&#91;`,
 // the entity-encoded form of `[`). Extraction mirrors the shared copy's
 // decode-aware behavior: the body is also scanned after a single
-// renderer-like entity decode (numeric references, the named hash and
-// ampersand entities — `&amp;` consumed so double-encoded forms stay inert;
-// named-entity case-variants decoded as deliberate over-refusal), so a hash
-// or digit smuggled as an entity still refuses.
+// renderer-like entity decode (numeric references at cmark-gfm's 8-digit
+// bound, plus the named hash entity; named case-variants decoded as
+// deliberate over-refusal), so a hash or digit smuggled as an entity still
+// refuses, while a double-encoded form's output is never re-read.
 const ISSUE_PR_ID_RE = /#(\d{1,9})/gu;
-const DECODABLE_ENTITY_RE = /&(?:#(?:\d{1,7}|x[0-9a-f]{1,6})|num|amp);/giu;
+const DECODABLE_ENTITY_RE = /&(?:#(?:\d{1,8}|x[0-9a-f]{1,8})|num);/giu;
 function decodeRenderedText(body) {
   return body.replace(DECODABLE_ENTITY_RE, (entity) => {
     const inner = entity.slice(1, -1).toLowerCase();
     if (inner === "num") return "#";
-    if (inner === "amp") return "&";
     const code = inner[1] === "x" ? Number.parseInt(inner.slice(2), 16) : Number.parseInt(inner.slice(1), 10);
     try {
       return String.fromCodePoint(code);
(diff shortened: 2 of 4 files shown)


skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs
```diff
@@ -618,6 +618,47 @@ test("runCli's id guard does not mistake a well-formed HTML numeric character re
...
+    await assert.rejects(
+      () => runCli(
+        [
+          "--repo", "owner/repo", "--pr", "17", "--head-sha", "abc1234000000000000000000000000000000000",
+          "--verdict", "clean", "--findings-summary", "see &num;123 there", "--next-action", "go",
+        ],
+        { env: stubNamedEntity.env, spawn: spawn, ghCommand: "gh", stdoutSink: [], stderrSink: [] },
+      ),
+      /#123/,
+    );
```
mfittko, line 661: **low** (`coverage`): The fallback suite pins only the positive named-entity shape for the hash entity, so the over-refusal direction of the hand-copied matcher is undefended. The core suite pins five named-form shapes: decode refusal, case-variant over-refusal, inert no-digit form, unrelated-entity non-match, allowlist passthrough. The fallback pins just the decode refusal. Because the fallback pattern is a hand-copied duplicate with no import and no cross-copy parity test anywhere in the repo, a widening drift in the fallback copy alone, for example replacing the num alternative with a generic letter run, still passes the fallback decode pin and ships silently. Its effect is a false refusal in the emergency posting path, which fails closed and blocks the degraded verdict post. The PR matrix row claiming both suites are green on the same shapes therefore overstates the parity actually pinned. — judge: act

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `e8a2e11ec1ac331c5afb51905e2349086b3fda08`
**Verdict:** clean

**Findings summary:** 8 angles reviewed; 1 finding (see per-angle breakdown below).

- `correctness` → `clean`
- `determinism` → `clean`
- `gate-evidence` → `clean`
- `pr-description` → `clean`
- `scope` → `clean`

**Next action:** Apply the judge-act fallback parity pins, push, rerun the draft chain at the new head

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `ebe84449dc89b9f87c7b572dde525f5437f0411d`
**Verdict:** clean

**Findings summary:** 8 angles reviewed; 4 findings (see per-angle breakdown below).

- `determinism` → `clean`
- `gate-evidence` → `clean`
- `pr-description` → `clean`
- `scope` → `clean`

**Next action:** Mark the PR ready for review (round-2 residuals judge-deferred or rejected; follow-ups filed at close-out)

skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs
```diff
@@ -618,6 +618,87 @@ test("runCli's id guard does not mistake a well-formed HTML numeric character re
...
+
+    // Non-match parity with the shared copy: the named entity with no digit
+    // run stays inert, and a different named entity followed by digits does
+    // not match — both bodies must POST successfully.
+    for (const inertSummary of ["a literal hash: &num; alone", "&nbsp;123 stays"]) {
+      const stubInert = await writeGhStub(
+        tempDir,
+        [
+          {
+            assertArgIncludes: ["api", "repos/owner/repo/issues/17/comments"],
```
mfittko, line 713: **nit** (`correctness`): The inert-body success stub asserts only the POST target args, not the posted body. Unlike the well-formed pin earlier in the same test, which pins assertStdinIncludes on the entity text, the loop's stub omits assertStdinIncludes, so the pin cannot tell apart 'the guard let the intact named entity through' from 'the renderer mangled the entity before the guard saw it'. The paired refusal pins keep the suite sound overall, so this is diagnostic strength only. — judge: defer

skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs
```diff
@@ -618,6 +618,87 @@ test("runCli's id guard does not mistake a well-formed HTML numeric character re
       ),
       /#123/,
     );
+
+    // Semicolon with no leading `&` is not a well-formed entity: still refused.
```
mfittko, line 622: **nit** (`correctness`): The five new pins are appended to one pre-existing test whose title names only the numeric-character-reference and non-semicolon cases, so it now covers nine scenarios under a name that never mentions the named entity form, per-occurrence exclusion, or over-refusal. node --test reports a single failure with a misleading title, and because assertions run fail-fast in sequence, an early break hides the later named-form pins. Only the inert loop carries a per-case assertion message. — judge: reject

skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs
```diff
@@ -618,6 +618,87 @@ test("runCli's id guard does not mistake a well-formed HTML numeric character re
...
+    );
+
+    // The HTML5 named entity for the hash character followed by digits decodes
+    // to a bare auto-link: refused (mirrors the shared guard copy).
+    const stubNamedEntity = await writeGhStub(tempDir, [], {});
+    await assert.rejects(
+      () => runCli(
+        [
+          "--repo", "owner/repo", "--pr", "17", "--head-sha", "abc1234000000000000000000000000000000000",
+          "--verdict", "clean", "--findings-summary", "see &num;123 there", "--next-action", "go",
```
mfittko, line 656: **low** (`coverage`): The fallback suite still has no pin for the hex or zero-padded numeric alternatives of the encoded-hash matcher. Its only numeric pin is the unpadded decimal form, so the hand-copied fallback pattern could lose the hex alternative or the zero-padding tolerance and the fallback suite would stay green. The core suite pins all four numeric shapes, and no automated parity check compares the two matcher copies, so suite-level shape parity is the only drift detector there is. This gap predates PR 1765 and sits outside the stated acceptance criteria, which name only the four non-entity boundary shapes. — judge: defer

skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs
```diff
@@ -618,6 +618,87 @@ test("runCli's id guard does not mistake a well-formed HTML numeric character re
...
+        ],
+        { env: stubNamedCaseVariant.env, spawn: spawn, ghCommand: "gh", stdoutSink: [], stderrSink: [] },
+      ),
+      /#456/,
+    );
+
+    // Non-match parity with the shared copy: the named entity with no digit
+    // run stays inert, and a different named entity followed by digits does
+    // not match — both bodies must POST successfully.
+    for (const inertSummary of ["a literal hash: &num; alone", "&nbsp;123 stays"]) {
```
mfittko, line 708: **nit** (`coverage`): The two new inert non-match pins assert only that the run exits zero and reports ok, not that the inert entity text survived into the posted body. The sibling well-formed-entity pin earlier in the same test uses assertStdinIncludes to pin posted body content, so these new pins are slightly weaker than the established pattern beside them. — judge: defer

skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs
mfittko: Deferred at gate close (round 2, fingerprint 0fd7f2a5dd749357, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs
mfittko: Deferred at gate close (round 2, fingerprint 8f7d45a6dbec38ff, severity nit, angle correctness): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs
mfittko: Deferred at gate close (round 2, fingerprint 5276f115f51436e5, severity nit, angle correctness): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs
mfittko: Deferred at gate close (round 2, fingerprint 9fa1f880789e9395, severity low, angle coverage): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs
mfittko: Deferred at gate close (round 2, fingerprint f7635dce9e222e11, severity nit, angle coverage): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

skills/dev-loop/scripts/post-gate-verdict-fallback.mjs
```diff
@@ -171,9 +171,11 @@ function collapseWhitespace(value) {
...
 // entity exclusion so the two copies stay behaviorally identical.
 const ISSUE_PR_ID_RE = /#(\d{1,9})/gu;
-// An entity decoding to the hash character followed by a digit run renders as
-// a bare auto-link; treated as a bare-id occurrence (mirrors the shared copy).
-const ENCODED_HASH_ID_RE = /&#(?:0*35|x0*23);(\d{1,9})/giu;
+// An entity decoding to the hash character (numeric forms, or the HTML5 named
+// entity form) followed by a digit run renders as a bare auto-link; treated
+// as a bare-id occurrence (mirrors the shared copy, including its deliberate
+// case-insensitive over-refusal of non-decoding variants of the named form).
+const ENCODED_HASH_ID_RE = /&(?:#(?:0*35|x0*23)|num);(\d{1,9})/giu;
```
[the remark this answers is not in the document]
mfittko: Fixed in commit 4cbe3268. The inline guard comment in this fallback file now names the num entity explicitly and states that only its lowercase form actually decodes, with the i flag's case-insensitive over-refusal called out as deliberate, matching the wording in the shared core comment-id-guard doc more closely so the two copies do not drift apart. Verified with node --test on both the core comment-id-guard test file and this fallback's test file; all tests pass.

skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs
```diff
@@ -618,6 +618,87 @@ test("runCli's id guard does not mistake a well-formed HTML numeric character re
...
+      const stubInert = await writeGhStub(
+        tempDir,
+        [
+          {
+            assertArgIncludes: ["api", "repos/owner/repo/issues/17/comments"],
+            stdout: '{"id":301,"html_url":"https://github.com/owner/repo/pull/17#issuecomment-301"}\n',
+          },
```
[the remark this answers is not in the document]
mfittko: Deferring this one. It is a real coverage gap, not a defect in the guard itself, and it sits outside this pull request's scope of comment-id-guard named-entity wording and fallback boundary parity: the sanitizer and allowlist behavior stay untouched here. Strengthening the inert non-match assertions to check the posted body content, alongside the related numeric-shape parity coverage, is already tracked as a judge-deferred follow-up to be filed separately rather than folded into this fix pass.

mfittko (commented):


**Reviewed head SHA:** `4cbe3268da22f696a1accc717160ad97e143f82b`
**Verdict:** clean

**Findings summary:** 5 angles reviewed; 5 findings (see per-angle breakdown below).

- `renderer-security` → `findings_present` (1 finding)

> **low** (`acceptance-criteria`): Definition-of-done item three, npm run assets:check green with no vendored copy of either file drifting, has no evidence at this head and also verifies nothing for this change. The recorded validation at this head lists only the verify and schema:check suites, and the verify script fans out to test:assets, test:extension, test:scripts, test:core, test:docs, test:pack, test:dev-loop and test:workflows, none of which is the generator's own check mode; test:assets is a similarly named but different node test suite, which is an easy substitution to make. Separately the drift risk the item names does not exist here: git ls-files shows exactly one copy of each changed file, and the asset generator mirrors skills/dev-loop/templates, not skills/dev-loop/scripts, so neither changed file has a vendored copy. The other three definition-of-done items are covered, since verify includes test:core and test:dev-loop which run both named suites, and the record reports allPassed true. All three acceptance criteria from issue 1755 are genuinely met on the extend arm: the named form is refused with pins in both suites, the fallback suite now pins well-formed exclusion, ampersand-without-semicolon, semicolon-without-ampersand and a true per-occurrence case reusing the same digit run in both shapes, and the two matcher patterns are character-identical apart from the pre-existing unicode flag on the fallback copy. — judge: act


> **low** (`pr-checklist-matrix`): The PR body's fourth definition-of-done item, npm run assets:check green with the parenthetical claim that no vendored copy of either file drifts, has no recorded evidence at this head and does not apply to this change. The gate validation record for this head records only the verify and schema:check suites, and the CI verify-suite matrix runs test:assets, test:extension, test:scripts, test:core, test:docs, test:pack, test:dev-loop and test:workflows, none of which is assets:check. The asset generator mirrors markdown sources such as skills/dev-loop/SKILL.md into .claude/, and neither changed source file has a vendored copy in the tree, so no markdown changed here and the check is vacuous for this diff. The matrix's third row then cites assets:check green as evidence for the criterion that both matcher copies stay semantically identical, which that check cannot establish because the two matcher copies are independently hand-maintained sources rather than one generated from the other. The remaining evidence for that criterion, mirrored pins in both suites plus the aligned mirror comments, does hold, so the criterion itself is met and only the cited evidence is misattributed. — judge: act

**Next action:** Apply the 4 judge-act fixes (decode-aware digit-side extraction in both matcher copies, comment dedup, PR-body evidence corrections), push, re-run the affected gate chain at the new head

skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs
```diff
@@ -618,6 +618,87 @@ test("runCli's id guard does not mistake a well-formed HTML numeric character re
...
+        [
+          "--repo", "owner/repo", "--pr", "17", "--head-sha", "abc1234000000000000000000000000000000000",
+          "--verdict", "clean", "--findings-summary", "&#91;ok&#93; but bare #91 too", "--next-action", "go",
+        ],
+        { env: stubPerOccurrence.env, spawn: spawn, ghCommand: "gh", stdoutSink: [], stderrSink: [] },
+      ),
+      /#91/,
+    );
+
+    // The HTML5 named entity for the hash character followed by digits decodes
```
mfittko, line 649: **low** (`contradiction-lens`): The named-entity refusal, the case-variant over-refusal and the two inert named-entity non-match posts are appended inside the test at line 557, whose name promises only the well-formed numeric-character-reference exclusion plus a non-semicolon-terminated or bare digit run. The headline behaviour of PR 1765 is therefore pinned under a test name that does not describe it. The core suite splits the same behaviour classes into separately named tests at its lines 59, 80, 100 and 123, including a dedicated named-entity test, so the mirroring the PR body claims holds at assertion level but not at test structure. Because all nine shapes now run sequentially in one test, a failure in an early shape reports the numeric-reference test name and aborts every shape after it. — judge: reject

packages/core/src/github/comment-id-guard.mjs
```diff
@@ -44,13 +45,17 @@ function isNumericCharacterReference(body, match) {
...
+// immediately followed by an issue-length digit run, renders as a bare
+// `#<digits>` auto-link on GitHub even though no literal hash-digit token
 // exists in the raw body. Treated as a bare-id occurrence so the encoded form
-// cannot smuggle a reference past the guard.
-const ENCODED_HASH_ID_RE = /&#(?:0*35|x0*23);(\d{1,9})/gi;
+// cannot smuggle a reference past the guard. The `i` flag makes the named
+// form match case-variants like `&NUM;` too; only lowercase `&num;` actually
+// decodes, so that is deliberate over-refusal on a suspicious near-miss,
+// keeping the guard fail-closed.
+const ENCODED_HASH_ID_RE = /&(?:#(?:0*35|x0*23)|num);(\d{1,9})/gi;
```
mfittko: **medium** (`renderer-security`): The widened matcher is decode-aware on the hash side only, never on the digit side, so an entity-encoded digit run still smuggles a live auto-link past the guard. Take a body that opens with a literal hash, then the decimal numeric reference for the digit one, then the literal characters two and three. ENCODED_HASH_ID_RE requires a literal digit immediately after the entity terminator, and here the next character is an ampersand, so it does not match. ISSUE_PR_ID_RE does find the digit run inside that inner reference, but isNumericCharacterReference skips it as a well-formed entity, and the leading literal hash has no literal digit after it. The body therefore extracts to an empty id list and posts. GitHub decodes the inner reference during inline parsing, so the rendered text node is a hash followed by one two three and the cross-reference filter auto-links it. This is the same premise the PR relies on for the hash side: if the decimal reference for the hash character auto-links, the auto-link pass runs on decoded text, which necessarily makes decoded digits auto-link too. The same shape works after an encoded hash as well, both numeric and the new named form. Beyond being an open bypass, the PR newly asserts completeness it does not deliver: the docblock now claims an entity decoding to the hash character followed by a digit run is treated as a bare id, and the non-goals state that only the hash-decoding entity is a leak vector for the auto-link rule. Both are incorrect while digit-side encoding is unhandled. The other candidate shapes for this angle are genuinely closed or out of threat model, see the review text. — judge: act

packages/core/src/github/comment-id-guard.mjs
```diff
@@ -44,13 +45,17 @@ function isNumericCharacterReference(body, match) {
   return body[match.index - 1] === "&" && body[match.index + match[0].length] === ";";
 }
 
-// An entity that DECODES to the hash character (`&#35;` decimal or `&#x23;`
-// hex, case-insensitive, zero-padding accepted — CommonMark decodes padded
-// forms too), immediately followed by an issue-length digit run, renders as a
-// bare `#<digits>` auto-link on GitHub even though no literal hash-digit token
+// An entity that DECODES to the hash character (`&#35;` decimal, `&#x23;`
```
mfittko: **nit** (`yagni`): The one-token matcher change carries a twelve-line inline comment that restates what the file docblock twenty lines above already says about the same three entity classes: the decimal form, the hex form with zero-padding, and the named &num; form. The same fact now lives twice in this file and a third time in the fallback copy's inline comment, so a future entity addition needs three prose edits to stay consistent. — judge: act

mfittko: Fixed in commits e0f56358872beb8f2a381ff5d87a3cd2f72d5f49 and 5e85d0d753443b8c78e30fe5479c259555c40ef8: the duplicated twelve-line encoded-pattern comment is gone — the encoded-hash pattern itself was replaced by the decode-aware extraction, whose comment describes only the decode mechanics while the file docblock carries the entity-class rationale once.

skills/dev-loop/scripts/post-gate-verdict-fallback.test.mjs
mfittko: Dismissing per the gate judge's disposition: the fallback suite deliberately appends these shapes to the existing boundary test rather than mirroring the core suite's per-shape test structure — assertion-level parity is what the acceptance criteria require, and the PR body's matrix now states the structural difference explicitly rather than claiming same-shape suites. Restructuring the fallback test file is tracked with the other pin-depth items in the filed follow-up on refusal-message pins.

packages/core/src/github/comment-id-guard.mjs
mfittko: Fixed in commits e0f56358872beb8f2a381ff5d87a3cd2f72d5f49 and 5e85d0d753443b8c78e30fe5479c259555c40ef8: extraction is now decode-aware on the digit side too — the body is also scanned after a single renderer-like entity decode covering numeric character references at the cmark-gfm 8-digit bound plus the named hash entity, so a literal or encoded hash followed by entity-encoded digits (and every mix) refuses. Pinned in both suites, including the maximally zero-padded decimal and hex hash forms.

packages/core/src/github/comment-id-guard.mjs
```diff
@@ -44,28 +56,47 @@ function isNumericCharacterReference(body, match) {
...
+      return String.fromCodePoint(code);
+    } catch {
+      return entity;
+    }
+  });
+}
+
+function collectBareIds(text, found) {
+  for (const m of text.matchAll(ISSUE_PR_ID_RE)) {
+    if (isNumericCharacterReference(text, m)) continue;
```
mfittko: **medium** (`renderer-security`): collectBareIds applies the numeric-character-reference exclusion to the DECODED text as well as to the raw body. After the single renderer-like decode, an ampersand-then-digits-then-semicolon shape is no longer an entity: the renderer already spent its one pass, exactly as this PR's own docblock argues. So a body made of a literal ampersand, then the decimal reference for the hash character, then the id digits, then a semicolon escapes BOTH scans. The raw scan skips the reference's own digit run as a well-formed entity, and the decoded scan then skips the assembled id because the decoded text now carries a leading ampersand and a trailing semicolon. extractIssuePrIds returns empty, yet GitHub renders that body as text whose hash-digits token is a live auto-link. The pre-PR encoded-hash loop had NO exclusion and refused this exact shape, so this is a fail-closed regression introduced by the decode rework; the named-hash-entity variant of the same wrapper is also unrefused, which defeats acceptance criterion one in a constructible shape. Note that in the decoded pass the exclusion can only ever fire when the hash character itself came from a decode, which is precisely the smuggling case it is supposed to catch, so it protects no legitimate body. Same defect in the fallback mirror at line 196. — judge: act

mfittko (commented):


**Reviewed head SHA:** `5e85d0d753443b8c78e30fe5479c259555c40ef8`
**Verdict:** clean

**Findings summary:** 5 angles reviewed; 5 findings (see per-angle breakdown below).

- `renderer-security` → `findings_present` (2 findings)
- `yagni` → `clean`

> **medium** (`contradiction-lens`): PR body Summary and Scope sections still claim the renderer-like decode covers the named ampersand entity, and that the ampersand entity is consumed so double-encoded forms are never re-read. The shipped decode set at this head is numeric character references plus the named hash entity only; the ampersand branch was removed in round 2. The never-re-read property is achieved by the single non-rescanning replace, not by consuming an ampersand entity. The same PR body's non-goal 2 already states the correct narrower scope, so the body also contradicts itself. Both source docblocks are accurate; only the PR body prose is stale. — judge: act


> **low** (`pr-checklist-matrix`): The AC/DoD/non-goals matrix still represents only one of the two stated non-goals. The Non-goal boundary column cites the sanitizer-and-allowlist boundary on row 1 and an em dash on rows 2 and 3, so the second non-goal, no general HTML entity decoding, appears nowhere in the matrix. Everything else this angle requires is complete and accurate at this head: the matrix has a header row plus three content rows, every acceptance criterion maps to at least one definition-of-done item, and all three definition-of-done items are cited as evidence. The non-goal text itself is now accurate against the diff, naming numeric character references at cmark-gfm's 8-digit bound plus the named hash entity, which matches DECODABLE_ENTITY_RE exactly after the ampersand branch removal, so only the matrix cross-reference is missing, not the boundary. This is the unfixed half of a round-2 low finding whose other half, the stale rationale clause, was corrected. — judge: act


> **nit** (`pr-checklist-matrix`): Matrix row 2 still overstates the fallback suite's digit-side coverage in the plural: it credits digit-side assembly parity pins where the fallback suite has exactly one such pin at this head, the encoded-first-digit case in the summary text see hash entity-49 23. The core suite pins three assembly shapes plus the padded decimal and hex hash forms. Row 3 was rewritten to disclose this asymmetry honestly, so only row 2's wording lags. Raised in round 2 and deferred. — judge: act


> **low** (`renderer-security`): Bound asymmetry between the two entity gates. ISSUE_PR_ID_RE captures up to 9 digits, so isNumericCharacterReference treats a 9-digit ampersand-digits-semicolon run as a well-formed reference and skips it, while DECODABLE_ENTITY_RE correctly stops at cmark-gfm's 8-digit cap. cmark-gfm refuses to decode a 9-digit reference, so it renders literally and the rendered text still carries a hash followed by that zero-padded digit run. A maximally padded 9-digit form of a real id is therefore excluded by the raw scan and never reached by the decode pass. Exactly 9 digits is the only slipping width: 10 or more digits already fails the exclusion and refuses. This is pre-existing on the exclusion side rather than introduced by this diff, but the PR's own cmark-parity rationale for the decode bound makes the mismatch on the exclusion bound worth closing in the same change. Impact depends on whether GitHub resolves a zero-padded id, which I could not verify from the worktree. — judge: act
> Location: `packages/core/src/github/comment-id-guard.mjs:53`

**Next action:** Apply the 5 judge-act fixes (decoded-scan exclusion regression in both copies with cmark replacement-char parity, 9-digit exclusion bound, PR-body prose), push, re-run the PA chain at the new head

packages/core/src/github/comment-id-guard.mjs
mfittko: Deferred at gate close (round 4, fingerprint 228439079c1a3955, severity medium, angle renderer-security): stayed open past this gate's round-3 medium fix window; the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/github/comment-id-guard.mjs
```diff
@@ -25,9 +25,21 @@
...
  * skipped; the same digit run appearing elsewhere as a bare token still
  * refuses — this is the ONLY sanctioned way a generated comment body may
- * carry a `#<digits>` token. An entity that itself decodes to the hash
- * character followed by a digit run is treated as a bare id (it renders as
- * an auto-link). Keep the allowlist small and deliberate.
+ * carry a `#<digits>` token. Extraction is decode-aware on BOTH sides of the
+ * token: the body is also scanned after a single left-to-right decode of the
+ * entity forms GitHub's renderer resolves (numeric character references,
+ * zero-padding and hex included at cmark-gfm's 8-digit bound, plus the named
+ * hash entity),
```
mfittko, line 32: **nit** (`contradiction-lens`): Cosmetic wrapping residue from removing the ampersand-branch clause: the docblock line ends mid-sentence after "hash entity)," leaving an orphan short line before the continuation. Prose content is accurate and matches the shipped decode set. — judge: reject

mfittko (commented):


**Reviewed head SHA:** `2e78ba3751955dcb2820b1c1be02979e39ca02db`
**Verdict:** clean

**Findings summary:** 5 angles reviewed; 1 finding (see per-angle breakdown below).

- `yagni` → `clean`

**Gate evidence note:** Copilot review rounds exhausted (3/2); current head has zero unresolved threads and green CI, so pre_approval_gate fallback is allowed without another Copilot re-request.

**Next action:** Merge under the recorded standing authorization once gate-evidence status is green (round-4 residual judge-rejected)

packages/core/src/github/comment-id-guard.mjs
mfittko: Dismissing as cosmetic: the guard already returns early for non-string bodies, so the repeated String coercions are harmless no-op noise rather than a defect, and this degraded fallback path deliberately favors leaving reviewed code untouched at this stage over another gate cycle for a style tweak. The fallback-copy simplification is tracked with the other pin-depth and structure items in the filed refusal-message follow-up issue 1762.

PR #1775, opened 2026-08-20
fix(gate): route the exported renderer through validation; refuse silent file-ref loss

### LINKED ISSUE #1615: post-gate-findings: unguarded exported renderer, silent file-ref loss, and consistency follow-ups from the gate-surface sweep
## Summary

Follow-ups deferred from the gate-surface sweep's draft gate. None block that change; all are in `scripts/github/post-gate-findings.mjs` unless noted, and all were found by adversarial review or mutation testing rather than by failing tests.

### 1. `renderFindingsCommentBody` is exported and reachable with zero guards

The sweep consolidated validation into `validateAndSanitizeRenderInputs`, but that seam lives inside the *bounded* entry point. The still-exported `renderFindingsCommentBody` reaches the same render with no guards at all, so a direct caller emits `gate=undefined` markers. The "no field bypasses validation by omission" property therefore holds only for the bounded path.

Either route the unbounded export through the same seam, or stop exporting it if the bounded entry point is the intended public surface.

### 2. `files` entries are validated raw but dropped after sanitisation

An entry like `` files: ["`"] `` passes validation, then sanitises to empty and is filtered out, so the file reference is silently lost from the rendered comment. Validate the post-sanitisation value, or keep a placeholder so the loss is visible.

### 3. `angle` has no bracketed-path legibility pin

`files` has a contract test pinning how bracketed paths render; `angle` goes through a different sanitiser (`sanitizeCodeSpan` vs `sanitizeInline`) with no equivalent pin, so a change there would not be caught. Pre-existing, not introduced by the sweep.

### 4. The `files` inner loop uses `.forEach`

The same function's outer loop was deliberately changed to `for...of` because `.forEach` skips array holes, with a comment mandating it. The inner `files` loop still uses `.forEach`. Benign today because the render also skips holes, but the two now contradict the stated rule.

### 5. `SEVERITY_ORDER` / `VALID_SEVERITIES` are not frozen

Three sibling exports in `packages/core/src/loop/gate-fanin.mjs` are `Object.freeze`d; these two are not. Nothing mutates them and the sweep's new code defends by spread-copying before `.reverse()`, but the asymmetry is an easy footgun.

### 6. Skipped nits

Delimiter `\s*` padding in the title-marker constructions, the degradation fast-path `<=` boundary, and the `GITHUB_COMMENT_MAX_CHARS` literal. Recorded for completeness; each is cosmetic.

## Acceptance criteria

- [x] No exported render entry point can be called in a way that bypasses input validation, or the unguarded export is removed.
- [x] A file reference that survives validation cannot vanish silently from the rendered comment.
- [x] `angle` rendering is pinned the same way `files` rendering is.
- [x] The hole-handling rule stated in the function's own comment is applied consistently within that function.
- [x] The severity vocabulary exports are frozen consistently with their siblings.

## Definition of done

- [x] `npm run verify` green with zero new failures.
- [x] Each change carries a test that fails when the behavior is reverted.

## Non-goals

- Changing the severity vocabulary, the degradation algorithm, or the title-marker constructions.
- Revisiting the acknowledged fullwidth/non-ASCII lookalike class or the deliberate `WIP foo bar` loosening.

mfittko:
## Summary

Hardening batch for the gate-findings renderer, closing the five actionable items deferred from the gate-surface sweep: the exported render entry point can no longer bypass input validation, a file reference that survives validation can no longer vanish silently, angle rendering gains the same legibility pin files has, the function's stated hole-handling rule now holds throughout the function, and the severity vocabulary exports are frozen like their siblings.

## Scope and context

Four files. `scripts/github/post-gate-findings.mjs`: `renderFindingsCommentBody` (kept exported — no caller outside the module and its own test file) now routes through the same `validateAndSanitizeRenderInputs` seam, and is the module's only render path (`renderBoundedFindingsCommentBody` composes it and re-runs the seam in its own scope so its closures see normalized values); the seam now also validates the caller-supplied `omittedCounts` (known severity, positive integer count) and each finding's `judgeDisposition`, and rejects a `headSha` that sanitizes to an empty string — completing the sanitize-to-empty guard family this diff itself builds out (angle/summary/disposition were guarded before it; the `files`, `judgeDisposition`, and `headSha` guards are all introduced here); every refusal names the entry point actually invoked (an `entryPoint` parameter replaces the hardcoded function name at every throw site); the inner `files` loop converted to `for...of` per the function's own hole-handling comment. `packages/core/src/loop/gate-fanin.mjs`: `SEVERITY_ORDER`, `NON_DEFECT_SEVERITIES`, and `VALID_SEVERITIES` frozen to match their frozen siblings, with a comment noting `Object.freeze` does not block Set mutation methods (done for consistency; nothing mutates any of the three exports, and the renderer's spread-copy-before-reverse stays). Tests (fourteen added): a direct-export bypass pin, a files-sanitizes-to-empty refusal pin, a headSha-sanitizes-to-empty refusal pin (both entry points), a sparse-files-array hole pin, an angle bracketed-path legibility pin mirroring the files pin, frozen-export pins, four omittedCounts pins (unknown severity, non-array, non-object element, non-positive-integer count), three judgeDisposition pins (non-string or blank, sanitizes-to-empty, null accepted unchanged), and a bounded-path pin asserting the fail-closed re-validation. Mutation checks ran for the guard routing (twenty tests fail on revert), the files and headSha sanitize-to-empty refusals, the hole-handling conversion, the omittedCounts and judgeDisposition validation blocks, the bounded-path re-validation, and both freezes — each revert failed its pin and each restore was verified byte-identical.

## Acceptance criteria

- [x] No exported render entry point can be called in a way that bypasses input validation, or the unguarded export is removed.
- [x] A file reference that survives validation cannot vanish silently from the rendered comment.
- [x] `angle` rendering is pinned the same way `files` rendering is.
- [x] The hole-handling rule stated in the function's own comment is applied consistently within that function.
- [x] The severity vocabulary exports are frozen consistently with their siblings.

## Definition of done

- [x] `npm run verify` green with zero new failures.
- [x] Each change carries a test that fails when the behavior is reverted (mutation-checked across the batch: guard routing, files and headSha sanitize-to-empty refusals, hole-handling conversion, omittedCounts and judgeDisposition validation blocks, bounded-path re-validation, and both freezes).

|---|---|---|
| Export cannot bypass validation | post-gate-findings suite green (direct-export pin; 20 tests fail on revert) | renderer output format unchanged |
| No silent file-ref loss | post-gate-findings suite green (sanitize-empty refusal pin) | no placeholder rendering added (refusal chosen per file convention) |
| Angle legibility pinned | post-gate-findings suite green (bracketed-path pin) | — |
| Consistent hole handling | post-gate-findings suite green (sparse-array pin) | — |
| Frozen vocabulary exports | gate-fanin suite green (frozen pins); npm run verify green | Set mutation methods not blockable by freeze (noted in comment) |
| Seam completeness: omittedCounts validated (gate-round addition under the first criterion) | post-gate-findings suite green (four omittedCounts pins, mutation-checked) | — |
| Seam completeness: judgeDisposition validated (gate-round addition under the first criterion) | post-gate-findings suite green (three judgeDisposition pins, mutation-checked) | — |
| Seam completeness: headSha sanitize-to-empty refusal (gate-round addition under the first criterion) | post-gate-findings suite green (both-entry-point pin, mutation-checked) | — |
| Refusal attribution names the invoked entry point (gate-round addition under the first criterion) | post-gate-findings suite green (misattribution pin via the bounded path) | — |

## Non-goals

- The issue's item six nits (delimiter padding, degradation boundary, the comment-size literal) stay skipped as recorded there.
- No renderer output-format changes.

## Open questions

None.

## Validation

```sh
node --test test/github/post-gate-findings.test.mjs packages/core/test/gate-fanin.test.mjs
npm run verify
```

Closes #1615

### DIFF
--- a/packages/core/src/loop/gate-fanin.mjs
+++ b/packages/core/src/loop/gate-fanin.mjs
@@ -175,7 +175,7 @@ export function reviewerBudgetPreflight(dispatchGroups, availableReviewers, { co
 // finding via the fix loop, a question via never being auto-deferred) — it
 // outranks "medium"/"low", which both eventually defer. "nit" trails last:
 // it defers immediately, with no fixer cycle at all.
-export const SEVERITY_ORDER = ["high", "question", "medium", "low", "nit"];
+export const SEVERITY_ORDER = Object.freeze(["high", "question", "medium", "low", "nit"]);
 
 // The non-defect subset of SEVERITY_ORDER: a "question" is answered (never
 // fixed or deferred like a defect — see deriveDisposition), and a "nit"
@@ -191,7 +191,12 @@ export const NON_DEFECT_SEVERITIES = new Set(["question", "nit"]);
 // resolveFanoutGroups maps the same way; passing the marker name verbatim
 // resolves no groups and silently downgrades pairing enforcement.
 export const GATE_CONFIG_KEY = Object.freeze({ draft_gate: "draft", pre_approval_gate: "preApproval" });
-export const VALID_SEVERITIES = new Set(SEVERITY_ORDER);
+// Object.freeze on a Set only locks its OWN properties, not the add/delete
+// methods that mutate its internal collection — freezing is still applied
+// here for consistency with SEVERITY_ORDER and this file's other frozen
+// exports (GATE_CONFIG_KEY, LEGACY_SEVERITY_ALIASES, etc.), and it does stop
+// a caller from attaching a stray own property to the Set object itself.
+export const VALID_SEVERITIES = Object.freeze(new Set(SEVERITY_ORDER));
 
 // Pre-rename spellings. Old ledgers, markers, and configs still carry them;
 // every read boundary normalizes through this map. Every SANCTIONED producer
--- a/packages/core/src/loop/gate-fanin.mjs
+++ b/packages/core/src/loop/gate-fanin.mjs
@@ -185,7 +185,7 @@ export const SEVERITY_ORDER = Object.freeze(["high", "question", "medium", "low"
 // partition so a consumer (e.g. config.mjs's BLOCKING_SEVERITY_SPELLINGS
 // vocabulary contract test) derives "defect severities" as
 // SEVERITY_ORDER minus this set, rather than re-hand-listing "question"/"nit".
-export const NON_DEFECT_SEVERITIES = new Set(["question", "nit"]);
+export const NON_DEFECT_SEVERITIES = Object.freeze(new Set(["question", "nit"]));
 
 // Marker gate name → gates.<key> config key. Owned here so every caller of
 // resolveFanoutGroups maps the same way; passing the marker name verbatim
--- a/packages/core/src/loop/gate-fanin.mjs
+++ b/packages/core/src/loop/gate-fanin.mjs
@@ -185,6 +185,11 @@ export const SEVERITY_ORDER = Object.freeze(["high", "question", "medium", "low"
 // partition so a consumer (e.g. config.mjs's BLOCKING_SEVERITY_SPELLINGS
 // vocabulary contract test) derives "defect severities" as
 // SEVERITY_ORDER minus this set, rather than re-hand-listing "question"/"nit".
+// Object.freeze on a Set only locks its OWN properties, not the add/delete
+// methods that mutate its internal collection — freezing is still applied
+// here for consistency with SEVERITY_ORDER and this file's other frozen
+// exports (GATE_CONFIG_KEY, LEGACY_SEVERITY_ALIASES, etc.), and it does stop
+// a caller from attaching a stray own property to the Set object itself.
 export const NON_DEFECT_SEVERITIES = Object.freeze(new Set(["question", "nit"]));
 
 // Marker gate name → gates.<key> config key. Owned here so every caller of
--- a/packages/core/test/gate-fanin.test.mjs
+++ b/packages/core/test/gate-fanin.test.mjs
@@ -29,9 +29,13 @@ import {
 
 // SEVERITY_ORDER, VALID_SEVERITIES, and NON_DEFECT_SEVERITIES must be frozen
 // like this file's other exported vocabulary constants (GATE_CONFIG_KEY,
-// LEGACY_SEVERITY_ALIASES, FANIN_SYNTHETIC_ANGLES, JUDGE_DISPOSITIONS) —
-// nothing in the repo mutates them today, but an unfrozen shared array/set is
-// a footgun a later change could silently corrupt for every consumer.
+// LEGACY_SEVERITY_ALIASES, FANIN_SYNTHETIC_ANGLES, JUDGE_DISPOSITIONS) — an
+// unfrozen shared array is a footgun a later change could silently corrupt
+// for every consumer. Object.freeze on VALID_SEVERITIES/NON_DEFECT_SEVERITIES
+// (both Sets) only locks their own properties, not Set.prototype.add/delete,
+// so it doesn't block content mutation the way it blocks Array.prototype.push
+// on SEVERITY_ORDER below — it's applied for export consistency and to stop
+// a stray own-property assignment on the Set object itself.
 test("SEVERITY_ORDER, VALID_SEVERITIES, and NON_DEFECT_SEVERITIES are frozen, matching their frozen sibling exports", () => {
   assert.equal(Object.isFrozen(SEVERITY_ORDER), true, "SEVERITY_ORDER must be frozen");
   assert.equal(Object.isFrozen(VALID_SEVERITIES), true, "VALID_SEVERITIES must be frozen");
--- a/packages/core/test/gate-fanin.test.mjs
+++ b/packages/core/test/gate-fanin.test.mjs
@@ -23,7 +23,19 @@ import {
   tallySeverities,
   zeroSeverityCounts,
   SEVERITY_ORDER,
+  VALID_SEVERITIES,
 } from "../src/loop/gate-fanin.mjs";
+
+// SEVERITY_ORDER and VALID_SEVERITIES must be frozen like this file's other
+// exported vocabulary constants (GATE_CONFIG_KEY, LEGACY_SEVERITY_ALIASES,
+// FANIN_SYNTHETIC_ANGLES, JUDGE_DISPOSITIONS) — nothing in the repo mutates
+// them today, but an unfrozen shared array/set is a footgun a later change
+// could silently corrupt for every consumer.
+test("SEVERITY_ORDER and VALID_SEVERITIES are frozen, matching their frozen sibling exports", () => {
+  assert.equal(Object.isFrozen(SEVERITY_ORDER), true, "SEVERITY_ORDER must be frozen");
+  assert.equal(Object.isFrozen(VALID_SEVERITIES), true, "VALID_SEVERITIES must be frozen");
+  assert.throws(() => { SEVERITY_ORDER.push("bogus"); }, TypeError);
+});
 
 function cleanAngle(angle) {
   return { angle, verdict: "clean", findings: [] };
--- a/packages/core/test/gate-fanin.test.mjs
+++ b/packages/core/test/gate-fanin.test.mjs
@@ -24,17 +24,29 @@ import {
   zeroSeverityCounts,
   SEVERITY_ORDER,
   VALID_SEVERITIES,
+  NON_DEFECT_SEVERITIES,
 } from "../src/loop/gate-fanin.mjs";
 
-// SEVERITY_ORDER and VALID_SEVERITIES must be frozen like this file's other
-// exported vocabulary constants (GATE_CONFIG_KEY, LEGACY_SEVERITY_ALIASES,
-// FANIN_SYNTHETIC_ANGLES, JUDGE_DISPOSITIONS) — nothing in the repo mutates
-// them today, but an unfrozen shared array/set is a footgun a later change
-// could silently corrupt for every consumer.
-test("SEVERITY_ORDER and VALID_SEVERITIES are frozen, matching their frozen sibling exports", () => {
+// SEVERITY_ORDER, VALID_SEVERITIES, and NON_DEFECT_SEVERITIES must be frozen
+// like this file's other exported vocabulary constants (GATE_CONFIG_KEY,
+// LEGACY_SEVERITY_ALIASES, FANIN_SYNTHETIC_ANGLES, JUDGE_DISPOSITIONS) —
+// nothing in the repo mutates them today, but an unfrozen shared array/set is
+// a footgun a later change could silently corrupt for every consumer.
+test("SEVERITY_ORDER, VALID_SEVERITIES, and NON_DEFECT_SEVERITIES are frozen, matching their frozen sibling exports", () => {
   assert.equal(Object.isFrozen(SEVERITY_ORDER), true, "SEVERITY_ORDER must be frozen");
   assert.equal(Object.isFrozen(VALID_SEVERITIES), true, "VALID_SEVERITIES must be frozen");
-  assert.throws(() => { SEVERITY_ORDER.push("bogus"); }, TypeError);
+  assert.equal(Object.isFrozen(NON_DEFECT_SEVERITIES), true, "NON_DEFECT_SEVERITIES must be frozen");
+  try {
+    assert.throws(() => { SEVERITY_ORDER.push("bogus"); }, TypeError);
+  } finally {
+    // If the freeze above ever regresses, the push actually succeeds BEFORE
+    // assert.throws fails — clean up the pushed entry so that regression
+    // fails only THIS test, instead of leaving "bogus" in the shared
+    // module-level SEVERITY_ORDER where every later order-dependent test in
+    // this file (zeroSeverityCounts/tallySeverities below) would fail too.
+    const bogusIndex = SEVERITY_ORDER.indexOf("bogus");
+    if (bogusIndex !== -1) SEVERITY_ORDER.splice(bogusIndex, 1);
+  }
 });
 
 function cleanAngle(angle) {
--- a/scripts/github/post-gate-findings.mjs
+++ b/scripts/github/post-gate-findings.mjs
@@ -321,7 +321,9 @@ export const GITHUB_COMMENT_MAX_CHARS = 65536;
 // goes through: a value newly added to what gets rendered can never bypass
 // validation by calling this export directly instead of the bounded one.
 export function renderFindingsCommentBody({ gate, headSha, findings, omittedCounts = [], maxChars = GITHUB_COMMENT_MAX_CHARS }) {
-  ({ gate, headSha } = validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars }));
+  ({ gate, headSha } = validateAndSanitizeRenderInputs({
+    gate, headSha, findings, omittedCounts, maxChars, entryPoint: "renderFindingsCommentBody",
+  }));
   const marker = buildFindingsMarker({ gate });
   const lines = [
     marker,
@@ -452,19 +454,25 @@ function describeInvalidValue(value) {
 }
 
 // Validates and sanitizes every caller-supplied value that
-// renderFindingsCommentBody renders, in ONE place, called from
-// renderFindingsCommentBody itself — so it also covers renderBoundedFindingsCommentBody,
-// which composes it rather than duplicating a render path — and a value newly
-// added to what gets rendered can never bypass validation by omission — four
-// consecutive review rounds each added one more one-off guard here before
-// this consolidation. Returns the comment's identity-key fields ready to
-// render: gate NORMALIZED (trim + lowercase, constrained to KNOWN_GATES —
-// never sanitized, since it is a closed two-value vocabulary, not free
-// text) and headSha SANITIZED (see sanitizeInline above; it is not a closed
-// set); findings are validated only here, since their own free-text fields
-// (summary/angle/files/disposition) are sanitized later, at render time, by
-// renderFindingsCommentBody itself.
-function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars }) {
+// renderFindingsCommentBody renders (including omittedCounts, its own
+// caller-supplied parameter — see summarizeDroppedBySeverity above), in ONE
+// place — called from renderFindingsCommentBody itself, and again from
+// renderBoundedFindingsCommentBody before it composes renderFindingsCommentBody,
+// so its own later use of the SAME gate/headSha (the fail-closed throw
+// further down) sees the normalized/sanitized values, never the caller's raw
+// input — so a value newly added to what gets rendered can never bypass
+// validation by omission on EITHER render entry point — four consecutive
+// review rounds each added one more one-off guard here before this
+// consolidation. `entryPoint` names the export that actually ran this call
+// (never hand-copied per throw site), so a rejection is always attributed to
+// the function the caller actually invoked, not a hardcoded stand-in.
+// Returns the comment's identity-key fields ready to render: gate NORMALIZED
+// (trim + lowercase, constrained to KNOWN_GATES — never sanitized, since it
+// is a closed two-value vocabulary, not free text) and headSha SANITIZED
+// (see sanitizeInline above; it is not a closed set); findings are validated
+// only here, since their own free-text fields (summary/angle/files/disposition)
+// are sanitized later, at render time, by renderFindingsCommentBody itself.
+function validateAndSanitizeRenderInputs({ gate, headSha, findings, omittedCounts, maxChars, entryPoint }) {
   // gate is the comment's IDENTITY key (buildFindingsMarker): an unvalidated
   // undefined/blank value would render `gate=undefined` and thereafter match
   // (and keep updating) that bogus marker on every later run. headSha is
@@ -478,17 +486,17 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars }) 
   // no collision surface between them left to sanitize away, so gate never
   // needs sanitizeInline at render time the way headSha does.
   if (typeof gate !== "string") {
-    throw new Error(`renderBoundedFindingsCommentBody: gate must be a non-empty string, got ${describeInvalidValue(gate)}`);
+    throw new Error(`${entryPoint}: gate must be a non-empty string, got ${describeInvalidValue(gate)}`);
   }
   const normalizedGate = gate.trim().toLowerCase();
   if (!KNOWN_GATES.has(normalizedGate)) {
-    throw new Error(`renderBoundedFindingsCommentBody: gate must be one of: ${[...KNOWN_GATES].join(", ")}, got ${describeInvalidValue(gate)}`);
+    throw new Error(`${entryPoint}: gate must be one of: ${[...KNOWN_GATES].join(", ")}, got ${describeInvalidValue(gate)}`);
   }
   if (typeof headSha !== "string" || headSha.trim().length === 0) {
-    throw new Error(`renderBoundedFindingsCommentBody: headSha must be a non-empty string, got ${describeInvalidValue(headSha)}`);
+    throw new Error(`${entryPoint}: headSha must be a non-empty string, got ${describeInvalidValue(headSha)}`);
   }
   if (!Array.isArray(findings)) {
-    throw new Error(`renderBoundedFindingsCommentBody: findings must be an array, got ${describeInvalidValue(findings)}`);
+    throw new Error(`${entryPoint}: findings must be an array, got ${describeInvalidValue(findings)}`);
   }
   // for...of (never .forEach, which SKIPS array holes) so a sparse findings
   // array produces the same named, index-bearing error as any other bad
@@ -499,10 +507,10 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars }) 
   let i = 0;
   for (const finding of findings) {
     if (!finding || typeof finding !== "object" || Array.isArray(finding)) {
-      throw new Error(`renderBoundedFindingsCommentBody: findings[${i}] must be an object, got ${describeInvalidValue(finding)}`);
+      throw new Error(`${entryPoint}: findings[${i}] must be an object, got ${describeInvalidValue(finding)}`);
     }
     if (!SEVERITY_ORDER.includes(normalizeSeverity(finding.severity))) {
-      throw new Error(`renderBoundedFindingsCommentBody: findings[${i}].severity must be one of: ${SEVERITY_ORDER.join(", ")}, got ${describeInvalidValue(finding.severity)}`);
+      throw new Error(`${entryPoint}: findings[${i}].severity must be one of: ${SEVERITY_ORDER.join(", ")}, got ${describeInvalidValue(finding.severity)}`);
     }
     // angle/summary are rendered directly into the comment (as a code span /
     // bare prose respectively); an unvalidated caller passing neither would
@@ -513,16 +521,16 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars }) 
     // the raw string would let it through and render an empty code span /
     // empty prose run.
     if (typeof finding.angle !== "string" || finding.angle.trim().length === 0) {
-      throw new Error(`renderBoundedFindingsCommentBody: findings[${i}].angle must be a non-empty string, got ${describeInvalidValue(finding.angle)}`);
+      throw new Error(`${entryPoint}: findings[${i}].angle must be a non-empty string, got ${describeInvalidValue(finding.angle)}`);
     }
     if (sanitizeCodeSpan(finding.angle).length === 0) {
-      throw new Error(`renderBoundedFindingsCommentBody: findings[${i}].angle sanitizes to an empty code span, got ${describeInvalidValue(finding.angle)}`);
+      throw new Error(`${entryPoint}: findings[${i}].angle sanitizes to an empty code span, got ${describeInvalidValue(finding.angle)}`);
     }
     if (typeof finding.summary !== "string" || finding.summary.trim().length === 0) {
-      throw new Error(`renderBoundedFindingsCommentBody: findings[${i}].summary must be a non-empty string, got ${describeInvalidValue(finding.summary)}`);
+      throw new Error(`${entryPoint}: findings[${i}].summary must be a non-empty string, got ${describeInvalidValue(finding.summary)}`);
     }
     if (sanitizeInline(finding.summary).length === 0) {
-      throw new Error(`renderBoundedFindingsCommentBody: findings[${i}].summary sanitizes to an empty string, got ${describeInvalidValue(finding.summary)}`);
+      throw new Error(`${entryPoint}: findings[${i}].summary sanitizes to an empty string, got ${describeInvalidValue(finding.summary)}`);
     }
     // disposition is rendered directly into the comment (bare prose, for any
     // truthy value, when present — see renderFindingsCommentBody's own `?:`
@@ -536,7 +544,7 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars }) 
       finding.disposition !== undefined && finding.disposition !== null && finding.disposition !== ""
       && (typeof finding.disposition !== "string" || finding.disposition.trim().length === 0)
     ) {
-      throw new Error(`renderBoundedFindingsCommentBody: findings[${i}].disposition must be a non-empty string when present, got ${describeInvalidValue(finding.disposition)}`);
+      throw new Error(`${entryPoint}: findings[${i}].disposition must be a non-empty string when present, got ${describeInvalidValue(finding.disposition)}`);
     }
     // The render uses the SANITIZED disposition (sanitizeInline), never the
     // raw one, so a non-blank string that sanitizes to nothing (e.g. a bare
@@ -546,14 +554,32 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars }) 
       typeof finding.disposition === "string" && finding.disposition.trim().length > 0
       && sanitizeInline(finding.disposition).length === 0
     ) {
-      throw new Error(`renderBoundedFindingsCommentBody: findings[${i}].disposition sanitizes to an empty string, got ${describeInvalidValue(finding.disposition)}`);
+      throw new Error(`${entryPoint}: findings[${i}].disposition sanitizes to an empty string, got ${describeInvalidValue(finding.disposition)}`);
+    }
+    // judgeDisposition is rendered directly into the comment (bare prose, for
+    // any truthy value, when present — see renderFindingsCommentBody's own
+    // judgeSuffix truthiness check), the exact same failure mode as
+    // disposition just above: an unvalidated truthy non-string is
+    // String-coerced by sanitizeInline and posted as junk, and a
+    // whitespace-only string collapses to an empty italic run.
+    if (
+      finding.judgeDisposition !== undefined && finding.judgeDisposition !== null && finding.judgeDisposition !== ""
+      && (typeof finding.judgeDisposition !== "string" || finding.judgeDisposition.trim().length === 0)
+    ) {
+      throw new Error(`${entryPoint}: findings[${i}].judgeDisposition must be a non-empty string when present, got ${describeInvalidValue(finding.judgeDisposition)}`);
+    }
+    if (
+      typeof finding.judgeDisposition === "string" && finding.judgeDisposition.trim().length > 0
+      && sanitizeInline(finding.judgeDisposition).length === 0
+    ) {
+      throw new Error(`${entryPoint}: findings[${i}].judgeDisposition sanitizes to an empty string, got ${describeInvalidValue(finding.judgeDisposition)}`);
     }
     // files entries are rendered directly as code-span file refs (see
     // renderFindingsCommentBody); an unvalidated element would otherwise post
     // the literal string "undefined"/"null"/"[object Object]" into the comment.
     if (finding.files !== undefined) {
       if (!Array.isArray(finding.files)) {
-        throw new Error(`renderBoundedFindingsCommentBody: findings[${i}].files must be an array, got ${describeInvalidValue(finding.files)}`);
+        throw new Error(`${entryPoint}: findings[${i}].files must be an array, got ${describeInvalidValue(finding.files)}`);
       }
       // for...of (never .forEach, which SKIPS array holes) — same rationale
       // as the outer findings loop above: a sparse files array must produce
@@ -561,7 +587,7 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars }) 
       let j = 0;
       for (const file of finding.files) {
         if (typeof file !== "string" || file.trim().length === 0) {
-          throw new Error(`renderBoundedFindingsCommentBody: findings[${i}].files[${j}] must be a non-empty string, got ${describeInvalidValue(file)}`);
+          throw new Error(`${entryPoint}: findings[${i}].files[${j}] must be a non-empty string, got ${describeInvalidValue(file)}`);
         }
         // The render uses the SANITIZED file ref (sanitizeCodeSpan), never
         // the raw one, and silently DROPS a ref that sanitizes to empty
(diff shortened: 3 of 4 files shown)


mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `f2a72973df138a524a075895baf9bafb23f86736`
**Verdict:** clean

**Findings summary:** 8 angles reviewed; 15 findings (see per-angle breakdown below).

- `determinism` → `findings_present` (2 findings)

> **low** (`coverage`): Removing the validation call from renderBoundedFindingsCommentBody is the one change in this PR with no test that fails when it is reverted, and it silently changed observable behavior. The bounded function's local gate and headSha are now the RAW caller values, and their only remaining in-scope use is the fail-closed throw message at line 643, so that message now reports an un-normalized gate and an un-sanitized headSha where it previously reported the normalized and sanitizeInline'd forms. The two tests that exercise this throw match only the substring 'cannot be rendered within', so neither the old nor the new form is pinned, leaving definition-of-done item two unmet for this change. — judge: act
> Location: `scripts/github/post-gate-findings.mjs:643`


> **nit** (`gate-evidence`): The validation record is complete, head-stamped correctly, and green (allPassed true; verify and schema:check both exitCode 0; 4057 tests, 0 fail; verify's test:core task covers the new frozen-export pins in packages/core/test), but the same record stamps depState status stale with node_modules/.package-lock.json absent, so the green run used dependencies not materialized in this worktree. Zero risk for this diff, which touches no manifest or lockfile. — judge: reject


> **low** (`pr-description`): The PR body's Scope and context states the export was kept because tests import it at 84 call sites; the reviewed tree has 40 call-form occurrences in one test file and 58 total mentions there, with no caller outside the module and its own test file. The justification still holds, but the figure in the body is not supported by the tree. — judge: act

**Next action:** Apply the 14 judge-act fixes (omittedCounts validation in the seam foremost), push, rerun the draft chain

scripts/github/post-gate-findings.mjs
```diff
@@ -446,7 +452,9 @@ function describeInvalidValue(value) {
 }
 
 // Validates and sanitizes every caller-supplied value that
-// renderBoundedFindingsCommentBody renders, in ONE place, so a value newly
+// renderFindingsCommentBody renders, in ONE place, called from
```
mfittko: **medium** (`contract-surface`): The validation seam's docblock was reworded from "every caller-supplied value that renderBoundedFindingsCommentBody renders, in ONE place" to "...that renderFindingsCommentBody renders, in ONE place", and that new claim is untrue. renderFindingsCommentBody also renders a caller-supplied omittedCounts parameter (the omission Note built at lines 336 to 346: omittedTotal, SEVERITY_LABELS lookup, raw count interpolation) which validateAndSanitizeRenderInputs never receives and never checks. The pre-diff wording was accurate precisely because renderBoundedFindingsCommentBody builds omittedCounts itself and never takes it from a caller; widening the subject to the now-directly-guarded export makes the module's central safety invariant overstated on the exact bypass-by-omission class it exists to prevent. — judge: act

scripts/github/post-gate-findings.mjs
```diff
@@ -577,7 +599,10 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars })
 // kept still cannot fit, so a round that truly cannot be posted is
 // never reported as a success.
 export function renderBoundedFindingsCommentBody({ gate, headSha, findings, maxChars = GITHUB_COMMENT_MAX_CHARS }) {
-  ({ gate, headSha } = validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars }));
+  // renderFindingsCommentBody now runs validateAndSanitizeRenderInputs itself
+  // (see its own doc comment), so this first call both validates and
+  // normalizes gate/headSha; every later call in this function is idempotent
```
mfittko: **low** (`contract-surface`): The new comment says the first render call "both validates and normalizes gate/headSha", but the destructuring assignment that propagated the normalized values into this scope was removed, so renderBoundedFindingsCommentBody's own gate/headSha stay exactly as the caller passed them. The observable consequence is the fail-closed throw at line 643, which interpolates gate and headSha directly: it used to report the normalized gate (trim plus lowercase) and the sanitized headSha, and now reports raw caller input, so a padded or mixed-case gate and a newline-bearing headSha reach that error string unnormalized. The same stale attribution appears at line 826, whose comment still says "Normalized the same way renderBoundedFindingsCommentBody normalizes gate" although this function no longer performs that normalization. — judge: act

packages/core/src/loop/gate-fanin.mjs
```diff
@@ -191,7 +191,12 @@ export const NON_DEFECT_SEVERITIES = new Set(["question", "nit"]);
 // resolveFanoutGroups maps the same way; passing the marker name verbatim
 // resolves no groups and silently downgrades pairing enforcement.
 export const GATE_CONFIG_KEY = Object.freeze({ draft_gate: "draft", pre_approval_gate: "preApproval" });
-export const VALID_SEVERITIES = new Set(SEVERITY_ORDER);
+// Object.freeze on a Set only locks its OWN properties, not the add/delete
+// methods that mutate its internal collection — freezing is still applied
+// here for consistency with SEVERITY_ORDER and this file's other frozen
+// exports (GATE_CONFIG_KEY, LEGACY_SEVERITY_ALIASES, etc.), and it does stop
+// a caller from attaching a stray own property to the Set object itself.
+export const VALID_SEVERITIES = Object.freeze(new Set(SEVERITY_ORDER));
```
mfittko, line 204: **low** (`contract-surface`): The freeze rationale states the change is made "for consistency with SEVERITY_ORDER and this file's other frozen exports", but the nearest peer severity-vocabulary export, NON_DEFECT_SEVERITIES at line 188, is still a bare unfrozen Set sitting between the two exports this change freezes. It is a first-class member of the same vocabulary surface (test/contracts/gate-severity-vocabulary-contract.test.mjs consumes it alongside SEVERITY_ORDER to derive the defect/non-defect partition), so the stated consistency property and the acceptance criterion wording about freezing consistently with siblings are only partly satisfied. — judge: act

scripts/github/post-gate-findings.mjs
```diff
@@ -315,7 +315,13 @@ export function sanitizeInline(value) {
 // number rather than a second hand-copied literal.
 export const GITHUB_COMMENT_MAX_CHARS = 65536;
 
+// This is the module's ONLY render entry point (renderBoundedFindingsCommentBody
+// below composes it rather than duplicating a render path), and it runs the
+// SAME validateAndSanitizeRenderInputs seam every caller — bounded or direct —
+// goes through: a value newly added to what gets rendered can never bypass
+// validation by calling this export directly instead of the bounded one.
 export function renderFindingsCommentBody({ gate, headSha, findings, omittedCounts = [], maxChars = GITHUB_COMMENT_MAX_CHARS }) {
+  ({ gate, headSha } = validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars }));
```
mfittko: **nit** (`contract-surface`): Every throw inside the seam is prefixed "renderBoundedFindingsCommentBody:", and those errors are now raised from direct renderFindingsCommentBody calls too, so the prefix can name a function the caller never invoked. Confirmed no contract surface pins the old behavior: no markdown doc under the worktree mentions either render entry point, and the new direct-export test matches on the message body without the prefix, so this stays cosmetic legibility rather than a contract break. — judge: act

scripts/github/post-gate-findings.mjs
mfittko: **low** (`correctness`): Removing the destructuring assignment leaves the enclosing gate and headSha bindings raw for the rest of renderBoundedFindingsCommentBody, but the added comment says this first call normalizes them. The one observable consequence today is the cannot-be-rendered-within-the-limit throw further down, which now interpolates the caller's raw gate and raw headSha where it previously interpolated the normalized gate and the sanitized headSha, so a gate passed with surrounding whitespace or mixed case is now reported unnormalized. The wider risk is the comment inviting a later reader to reuse these bindings as if normalized, which is the marker-disagreement trap the marker comment in postGateFindings warns about. — judge: act

scripts/github/post-gate-findings.mjs
```diff
@@ -315,7 +315,13 @@ export function sanitizeInline(value) {
 // number rather than a second hand-copied literal.
 export const GITHUB_COMMENT_MAX_CHARS = 65536;
 
+// This is the module's ONLY render entry point (renderBoundedFindingsCommentBody
+// below composes it rather than duplicating a render path), and it runs the
+// SAME validateAndSanitizeRenderInputs seam every caller — bounded or direct —
+// goes through: a value newly added to what gets rendered can never bypass
+// validation by calling this export directly instead of the bounded one.
 export function renderFindingsCommentBody({ gate, headSha, findings, omittedCounts = [], maxChars = GITHUB_COMMENT_MAX_CHARS }) {
```
mfittko, line 323: **medium** (`coverage`): omittedCounts is a caller-supplied parameter of the now-guarded export that is rendered but never validated, and no test pins it. The new doc comment claims a value newly added to what gets rendered can never bypass validation via a direct call, and acceptance criterion 1 claims no exported render entry point can bypass input validation, but validateAndSanitizeRenderInputs takes only gate, headSha, findings and maxChars. A direct caller passing an omittedCounts entry with an unknown severity renders the literal text 'undefined' in the omission note at line 338, and a non-array truthy omittedCounts reaches the reduce at line 337 as an unnamed TypeError. This is the same bypass-by-omission class the issue's item one raises, left on the very export the PR just guarded, with no pin either way. — judge: act

scripts/github/post-gate-findings.mjs
```diff
@@ -577,7 +599,10 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars })
 // kept still cannot fit, so a round that truly cannot be posted is
 // never reported as a success.
 export function renderBoundedFindingsCommentBody({ gate, headSha, findings, maxChars = GITHUB_COMMENT_MAX_CHARS }) {
-  ({ gate, headSha } = validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars }));
+  // renderFindingsCommentBody now runs validateAndSanitizeRenderInputs itself
+  // (see its own doc comment), so this first call both validates and
+  // normalizes gate/headSha; every later call in this function is idempotent
+  // against the same already-valid input.
   const body = renderFindingsCommentBody({ gate, headSha, findings, maxChars });
```
mfittko, line 674: **low** (`determinism`): Dropping the reassignment from validateAndSanitizeRenderInputs in renderBoundedFindingsCommentBody leaves that function's own gate and headSha bindings RAW, so its fail-closed refusal text at line 643 (Gate findings comment for gate ... at head ...) now echoes caller-shaped input instead of the normalized gate and sanitized headSha that were actually rendered. Same logical input with different casing or padding (draft_gate vs Draft_Gate vs padded) now yields different refusal text, and a whitespace-bearing headSha reaches the message unsanitized (harmless today only because the CLI emits it through JSON.stringify at line 890). No test pins the interpolated values, so the change is uncovered. Rendered bodies are unaffected: every render call re-normalizes idempotently, and no production caller reaches the direct export, so previously-valid production output stays byte-identical. — judge: act

packages/core/test/gate-fanin.test.mjs
```diff
@@ -23,8 +23,20 @@ import {
...
 
+// SEVERITY_ORDER and VALID_SEVERITIES must be frozen like this file's other
+// exported vocabulary constants (GATE_CONFIG_KEY, LEGACY_SEVERITY_ALIASES,
+// FANIN_SYNTHETIC_ANGLES, JUDGE_DISPOSITIONS) — nothing in the repo mutates
+// them today, but an unfrozen shared array/set is a footgun a later change
+// could silently corrupt for every consumer.
+test("SEVERITY_ORDER and VALID_SEVERITIES are frozen, matching their frozen sibling exports", () => {
+  assert.equal(Object.isFrozen(SEVERITY_ORDER), true, "SEVERITY_ORDER must be frozen");
+  assert.equal(Object.isFrozen(VALID_SEVERITIES), true, "VALID_SEVERITIES must be frozen");
+  assert.throws(() => { SEVERITY_ORDER.push("bogus"); }, TypeError);
```
mfittko: **low** (`determinism`): The new frozen-export pin attempts a real mutation of the shared module-level SEVERITY_ORDER export and relies solely on the throw to prevent it. If the freeze ever regresses (exactly the case this pin exists for), the push SUCCEEDS before assert.throws fails, leaving the bogus entry in a module constant that later tests in the same file read: the zeroSeverityCounts key pin at line 86 and the tallySeverities pins at lines 120 and 122 would then fail too, and the line 119 case (which asserts a bogus severity is excluded from the tally) flips meaning. One regression becomes a cascade of unrelated order-dependent failures. — judge: act

scripts/github/post-gate-findings.mjs
```diff
@@ -547,11 +555,25 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars })
...
+        // (renderFindingsCommentBody's own files.filter) — so a raw ref that
+        // is non-blank but sanitizes to nothing (e.g. a bare "`") must be
+        // rejected here, matching angle/summary/disposition above, rather
+        // than vanishing from the rendered comment with no trace.
+        if (sanitizeCodeSpan(file).length === 0) {
+          throw new Error(`renderBoundedFindingsCommentBody: findings[${i}].files[${j}] sanitizes to an empty code span, got ${describeInvalidValue(file)}`);
+        }
+        j += 1;
+      }
     }
```
mfittko, line 611: **medium** (`input-validation`): judgeDisposition is rendered into the comment but is a rendered finding field the seam never validates. The render uses a bare truthiness check, so a truthy non-string, for example an object, a number, or a boolean, is String-coerced by sanitizeInline and posted as junk, and a whitespace-only string collapses to an empty italic run. That is the exact failure mode the sibling disposition guards a few lines above exist to prevent, and it falsifies the absolute claim this PR adds to the render entry point's doc comment that no rendered value can bypass validation by omission. — judge: act

scripts/github/post-gate-findings.mjs
mfittko, line 323: **low** (`input-validation`): omittedCounts is a parameter of the newly guarded export only, so the seam never receives or validates it, yet it is rendered into the omission note. A direct caller passing an unknown severity renders the breakdown as a count followed by the word undefined, a non-numeric count renders a concatenated total, and a non-array value crashes with an unnamed TypeError instead of the seam's named error. The production path builds it from summarizeDroppedBySeverity so it is always well formed there, but the doc comment added here states the no-bypass property without exception. — judge: act

scripts/github/post-gate-findings.mjs
```diff
@@ -577,7 +599,10 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars })
 // kept still cannot fit, so a round that truly cannot be posted is
 // never reported as a success.
 export function renderBoundedFindingsCommentBody({ gate, headSha, findings, maxChars = GITHUB_COMMENT_MAX_CHARS }) {
-  ({ gate, headSha } = validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars }));
+  // renderFindingsCommentBody now runs validateAndSanitizeRenderInputs itself
+  // (see its own doc comment), so this first call both validates and
```
mfittko: **low** (`pr-description`): The comment replacing the removed validation call claims this first call both validates and normalizes gate/headSha, but the reassignment now happens inside renderFindingsCommentBody's own scope, so renderBoundedFindingsCommentBody's locals stay raw; the throw at line 643 consequently interpolates the un-normalized, un-sanitized gate and headSha where it previously interpolated the normalized ones. — judge: act

scripts/github/post-gate-findings.mjs
mfittko: **low** (`scope`): Item one is only half-done for diagnostics: the direct export now runs the shared validation seam, but all 16 thrown messages inside validateAndSanitizeRenderInputs still hardcode the prefix renderBoundedFindingsCommentBody, so a caller of renderFindingsCommentBody is told a function it never called rejected its input. — judge: act

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `1e88ef5a760474f2f0efaed0c1d29911aa2d9731`
**Verdict:** clean

**Findings summary:** 8 angles reviewed; 6 findings (see per-angle breakdown below).

- `coverage` → `clean`
- `determinism` → `clean`
- `gate-evidence` → `clean`
- `scope` → `clean`

> **low** (`pr-description`): The body's Scope and context names only SEVERITY_ORDER and VALID_SEVERITIES as newly frozen, but the diff also freezes NON_DEFECT_SEVERITIES in the same hunk, and the new gate-fanin pin asserts all three. The body under-describes the change by one export. — judge: act


> **low** (`pr-description`): The justification clause for keeping the export states the test file calls it 40 times. That is the pre-change count; at the reviewed head the same test file has 48 invocation sites, since this PR adds eight more. The load-bearing part of the claim, no caller outside the module and its test file, is verified accurate. — judge: act

**Next action:** Apply the 6 judge-act fixes (headSha sanitize-to-empty guard foremost), push, rerun the draft chain

scripts/github/post-gate-findings.mjs
```diff
@@ -577,7 +650,19 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars })
...
 export function renderBoundedFindingsCommentBody({ gate, headSha, findings, maxChars = GITHUB_COMMENT_MAX_CHARS }) {
-  ({ gate, headSha } = validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars }));
+  // renderFindingsCommentBody runs validateAndSanitizeRenderInputs itself
+  // (see its own doc comment) on every call below, so that alone would be
+  // enough to keep the RENDERED body safe. But this function's OWN gate/headSha
+  // bindings must also be normalized/sanitized here, in this scope, before
+  // anything else reads them — the fail-closed throw further down interpolates
+  // them directly, and dropOrder/renderWithDropped below close over them too —
+  // so a caller-shaped (unnormalized/unsanitized) value never leaks into this
+  // function's own output. This call is idempotent against renderFindingsCommentBody's
```
mfittko, line 668: **low** (`contract-surface`): The new bounded-entry docblock asserts the seam call is idempotent so the inner re-validation "can never reject them", but headSha is the one field checked raw-only and never for sanitize-to-empty: a headSha of three backticks passes the raw non-empty check, is returned as the empty string by sanitizeInline, and the composed renderFindingsCommentBody call then throws headSha must be a non-empty string. That also misattributes the refusal to renderFindingsCommentBody even though the caller invoked the bounded export, contradicting the same round's entryPoint attribution claim in the seam docblock. Direct-export callers get the mirror gap: no second validation, so the comment renders the bare line Reviewed head: with the value gone, the silent-loss class this PR closed for files and disposition. — judge: act

scripts/github/post-gate-findings.mjs
```diff
@@ -577,7 +650,19 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars })
...
+  // own re-validation below: gate/headSha are already normalized/sanitized by
+  // the time they reach it, so that second call can never reject them.
+  ({ gate, headSha } = validateAndSanitizeRenderInputs({
```
mfittko, line 671: **low** (`correctness`): The new comment above this call states the second validation "can never reject" the already-normalized values. That holds for gate and for every headSha whose sanitized form is non-empty, but not for a headSha that sanitizes to empty: the wrapper's own call accepts the raw value and rebinds headSha to the empty string, and the composed renderFindingsCommentBody call then throws its own non-empty headSha refusal. The caller invoked the bounded export but the error names renderFindingsCommentBody, which is the exact misattribution the entryPoint parameter was added to eliminate, and the reported value is the empty string rather than the caller's input. Everything else checks out: the drop path's internally built omittedCounts always satisfies the new validation because summarizeDroppedBySeverity keys on normalizeSeverity output and emits positive integer counts, sanitizeInline is idempotent because it never re-encodes an ampersand and its outputs contain no raw backtick, angle bracket or square bracket left to re-encode, and all twenty-two throws in the seam interpolate entryPoint with no hardcoded name left behind. — judge: act

scripts/github/post-gate-findings.mjs
```diff
@@ -470,17 +486,17 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars })
...
   if (typeof gate !== "string") {
-    throw new Error(`renderBoundedFindingsCommentBody: gate must be a non-empty string, got ${describeInvalidValue(gate)}`);
+    throw new Error(`${entryPoint}: gate must be a non-empty string, got ${describeInvalidValue(gate)}`);
   }
   const normalizedGate = gate.trim().toLowerCase();
   if (!KNOWN_GATES.has(normalizedGate)) {
-    throw new Error(`renderBoundedFindingsCommentBody: gate must be one of: ${[...KNOWN_GATES].join(", ")}, got ${describeInvalidValue(gate)}`);
+    throw new Error(`${entryPoint}: gate must be one of: ${[...KNOWN_GATES].join(", ")}, got ${describeInvalidValue(gate)}`);
   }
   if (typeof headSha !== "string" || headSha.trim().length === 0) {
```
mfittko, line 495: **medium** (`input-validation`): headSha is the only rendered value in the seam with no sanitize-to-empty guard. The round-2 batch added that guard for angle, summary, disposition, judgeDisposition and files entries, but headSha is still checked only on the RAW value, then returned as sanitizeInline of that value at line 634. A raw headSha of three backticks, or one backtick, passes the non-empty check and sanitizes to the empty string, so the direct export renders the identity line as "Reviewed head: " with nothing after it and still writes the gate marker. That is the same silently-vanishing-value class the acceptance criterion on file references closes, applied to the comment's head-identity field. Not reachable from the CLI, which constrains --head-sha to hex via normalizeHeadSha, so this is a programmatic-caller hole only. — judge: act

scripts/github/post-gate-findings.mjs
```diff
@@ -538,25 +554,82 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars })
...
+  }
+  let m = 0;
+  for (const entry of omittedCounts) {
+    if (!entry || typeof entry !== "object" || Array.isArray(entry)) {
+      throw new Error(`${entryPoint}: omittedCounts[${m}] must be an object, got ${describeInvalidValue(entry)}`);
+    }
+    if (!SEVERITY_ORDER.includes(entry.severity)) {
+      throw new Error(`${entryPoint}: omittedCounts[${m}].severity must be one of: ${SEVERITY_ORDER.join(", ")}, got ${describeInvalidValue(entry.severity)}`);
+    }
+    if (!Number.isFinite(entry.count) || entry.count < 0) {
```
mfittko: **nit** (`input-validation`): The new omittedCounts count check accepts zero and non-integers, while maxChars twelve lines below uses Number.isInteger with a positive bound. A direct caller passing a count of zero renders the note "0 finding(s) omitted from this comment (0 Nit)", and a count of 2.5 renders "2.5 Nit". The internally built omittedCounts from summarizeDroppedBySeverity only ever carries positive integers, so no production path is affected. — judge: act

scripts/github/post-gate-findings.mjs
```diff
@@ -577,7 +658,19 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars })
...
 export function renderBoundedFindingsCommentBody({ gate, headSha, findings, maxChars = GITHUB_COMMENT_MAX_CHARS }) {
-  ({ gate, headSha } = validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars }));
+  // renderFindingsCommentBody runs validateAndSanitizeRenderInputs itself
+  // (see its own doc comment) on every call below, so that alone would be
+  // enough to keep the RENDERED body safe. But this function's OWN gate/headSha
+  // bindings must also be normalized/sanitized here, in this scope, before
+  // anything else reads them — the fail-closed throw further down interpolates
+  // them directly, and dropOrder/renderWithDropped below close over them too —
+  // so a caller-shaped (unnormalized/unsanitized) value never leaks into this
+  // function's own output. This call is idempotent against renderFindingsCommentBody's
```
mfittko, line 668: **low** (`determinism`): The composed double-validation is safe only because sanitizeInline is idempotent, and this added comment states that as the reason the inner re-validation can never reject. Idempotency holds today (after one pass no backtick, less-than or opening-bracket character survives, whitespace is already collapsed and trimmed, and the inserted entity text reintroduces no trigger), but no test pins it. A later hardening rule that entity-encodes the ampersand would break it silently: the bounded export would sanitize headSha once, the composed inner call would sanitize the already-encoded value again, and the rendered identity line would then carry a different value than the one the bounded export validated and interpolates into its own fail-closed refusal. — judge: defer

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `f795390765031f8909088b4bd4d6065243e01cc0`
**Verdict:** clean

**Findings summary:** 8 angles reviewed; 1 finding (see per-angle breakdown below).

- `correctness` → `clean`
- `coverage` → `clean`
- `gate-evidence` → `clean`
- `pr-description` → `clean`
- `scope` → `clean`

**Next action:** Mark the PR ready for review (round-3 residual judge-deferred; follow-up filed at close-out)

scripts/github/post-gate-findings.mjs
mfittko: Deferred at gate close (round 3, fingerprint 1a8e4102d07e7381, severity low, angle contract-surface): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/loop/gate-fanin.mjs
mfittko: Deferred at gate close (round 3, fingerprint f11eb04827c36586, severity low, angle contract-surface): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/post-gate-findings.mjs
mfittko: Deferred at gate close (round 3, fingerprint 26e8360e3d6215f5, severity nit, angle contract-surface): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/post-gate-findings.mjs
mfittko: Deferred at gate close (round 3, fingerprint 39bb8fe116944028, severity low, angle correctness): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/post-gate-findings.mjs
mfittko: Deferred at gate close (round 3, fingerprint 156948d41e667b55, severity low, angle determinism): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/test/gate-fanin.test.mjs
mfittko: Deferred at gate close (round 3, fingerprint 9e4d13db05066449, severity low, angle determinism): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/post-gate-findings.mjs
mfittko: Deferred at gate close (round 3, fingerprint edf17dc6de1ddab3, severity low, angle input-validation): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/post-gate-findings.mjs
mfittko: Deferred at gate close (round 3, fingerprint fd20b90bc1d540d3, severity low, angle pr-description): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/post-gate-findings.mjs
mfittko: Deferred at gate close (round 3, fingerprint a4f76b7d023cb2b4, severity low, angle scope): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/post-gate-findings.mjs
mfittko: Deferred at gate close (round 3, fingerprint b9ef0a16ac54b16c, severity low, angle contract-surface): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/post-gate-findings.mjs
mfittko: Deferred at gate close (round 3, fingerprint 51167e5af6db5014, severity low, angle correctness): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/post-gate-findings.mjs
mfittko: Deferred at gate close (round 3, fingerprint 7ebed0b343984f23, severity nit, angle input-validation): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/post-gate-findings.mjs
mfittko: Deferred at gate close (round 3, fingerprint 59460bb598d62e99, severity low, angle determinism): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/post-gate-findings.mjs
mfittko: Fixed in commit 1e88ef5a760474f2f0efaed0c1d29911aa2d9731: each finding's judgeDisposition is validated in the seam under the same non-empty-string and sanitizes-non-empty rule as disposition, with pins on both the refusal and the accepted shapes. Mutation-checked: removing the block fails two pins.

scripts/github/post-gate-findings.mjs
mfittko: Fixed in commit 1e88ef5a760474f2f0efaed0c1d29911aa2d9731: the validation seam's docblock was rewritten to state truthfully what it guards on both entry points, and the seam now actually validates every rendered caller-supplied value including omittedCounts and each finding's judgeDisposition, so the ONE-place claim holds again. Round-two contract-surface review verified the docblock claims true.

scripts/github/post-gate-findings.mjs
mfittko: Fixed in commit 1e88ef5a760474f2f0efaed0c1d29911aa2d9731: the seam validates omittedCounts (each entry must carry a known severity and, after the follow-up commit f795390765031f8909088b4bd4d6065243e01cc0, a positive integer count matching the maxChars strictness), with pins for the unknown-severity and non-array shapes that previously rendered the literal text undefined or crashed in a reduce. Mutation-checked: removing the validation block fails four pins.

scripts/github/post-gate-findings.mjs
mfittko: Fixed in commit f795390765031f8909088b4bd4d6065243e01cc0: headSha gained the same sanitize-to-empty guard as every other rendered field, so a raw value of backticks now refuses instead of rendering a broken identity line, and the refusal names the entry point actually invoked. Mutation-checked: reverting the guard fails the new pin and reproduces the misattributed error the finding described.

packages/core/src/loop/gate-fanin.mjs
```diff
@@ -185,13 +185,18 @@ export const SEVERITY_ORDER = ["high", "question", "medium", "low", "nit"];
 // partition so a consumer (e.g. config.mjs's BLOCKING_SEVERITY_SPELLINGS
 // vocabulary contract test) derives "defect severities" as
 // SEVERITY_ORDER minus this set, rather than re-hand-listing "question"/"nit".
-export const NON_DEFECT_SEVERITIES = new Set(["question", "nit"]);
+export const NON_DEFECT_SEVERITIES = Object.freeze(new Set(["question", "nit"]));
```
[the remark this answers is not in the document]
mfittko: Fixed in 6b778659. NON_DEFECT_SEVERITIES now carries the same clarifying note as VALID_SEVERITIES: freezing a Set locks only its own properties, not Set.prototype.add/delete, so the freeze here is for export consistency and to stop a stray own-property assignment on the Set object, not to block content mutation.

packages/core/test/gate-fanin.test.mjs
```diff
@@ -23,8 +23,32 @@ import {
...
   SEVERITY_ORDER,
+  VALID_SEVERITIES,
+  NON_DEFECT_SEVERITIES,
 } from "../src/loop/gate-fanin.mjs";
 
+// SEVERITY_ORDER, VALID_SEVERITIES, and NON_DEFECT_SEVERITIES must be frozen
+// like this file's other exported vocabulary constants (GATE_CONFIG_KEY,
+// LEGACY_SEVERITY_ALIASES, FANIN_SYNTHETIC_ANGLES, JUDGE_DISPOSITIONS) —
+// nothing in the repo mutates them today, but an unfrozen shared array/set is
+// a footgun a later change could silently corrupt for every consumer.
```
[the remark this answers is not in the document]
mfittko: Fixed in 6b778659. Reworded the test comment to separate the two cases: SEVERITY_ORDER is an Array, so Object.freeze does block Array.prototype.push (verified by the assert.throws below it); VALID_SEVERITIES and NON_DEFECT_SEVERITIES are Sets, where freeze only locks own properties and does not block Set.prototype.add/delete.

scripts/github/post-gate-findings.mjs
```diff
@@ -470,17 +486,25 @@ function validateAndSanitizeRenderInputs({ gate, headSha, findings, maxChars })
...
+    throw new Error(`${entryPoint}: gate must be one of: ${[...KNOWN_GATES].join(", ")}, got ${describeInvalidValue(gate)}`);
   }
   if (typeof headSha !== "string" || headSha.trim().length === 0) {
-    throw new Error(`renderBoundedFindingsCommentBody: headSha must be a non-empty string, got ${describeInvalidValue(headSha)}`);
+    throw new Error(`${entryPoint}: headSha must be a non-empty string, got ${describeInvalidValue(headSha)}`);
+  }
+  // The render uses the SANITIZED headSha (sanitizeInline), never the raw
+  // one, so a non-blank string that sanitizes to nothing (e.g. a bare
+  // "```") must be rejected here too — matching angle/summary/disposition/
+  // judgeDisposition/files above — otherwise the identity line renders as
```
mfittko, line 501: **low** (`contradiction-lens`): The new headSha sanitize-to-empty comment says it matches the angle, summary, disposition, judgeDisposition and files guards above, but every one of those guards is BELOW this point in the same function: angle at lines 532 and 535, judgeDisposition at 583, the files sanitize guard after it. Two of them (judgeDisposition, files sanitize-to-empty) are also new in this same change, so above cannot mean pre-existing precedent either. The PR body repeats the same claim with the words already applied, which reads as if those two guards predated this change. — judge: defer

scripts/github/post-gate-findings.mjs
```diff
@@ -446,17 +454,25 @@ function describeInvalidValue(value) {
...
+// validation by omission on EITHER render entry point — four consecutive
+// review rounds each added one more one-off guard here before this
+// consolidation. `entryPoint` names the export that actually ran this call
+// (never hand-copied per throw site), so a rejection is always attributed to
+// the function the caller actually invoked, not a hardcoded stand-in.
+// Returns the comment's identity-key fields ready to render: gate NORMALIZED
+// (trim + lowercase, constrained to KNOWN_GATES — never sanitized, since it
+// is a closed two-value vocabulary, not free text) and headSha SANITIZED
+// (see sanitizeInline above; it is not a closed set); findings are validated
+// only here, since their own free-text fields (summary/angle/files/disposition)
```
mfittko, line 473: **low** (`contradiction-lens`): The seam docblock, rewritten by this change, still enumerates the per-finding free-text fields sanitized at render time as summary, angle, files, disposition, omitting judgeDisposition. This same change added a judgeDisposition guard here and judgeDisposition is sanitized at render time by the judgeSuffix line, and the comment at line 501 does list it. The docblock is the one a future field-adder reads to learn which fields the seam defers to render-time sanitization, so the omission is the exact bypass-by-omission trap the seam exists to prevent. — judge: defer

packages/core/test/gate-fanin.test.mjs
```diff
@@ -23,8 +23,36 @@ import {
...
+// (both Sets) only locks their own properties, not Set.prototype.add/delete,
+// so it doesn't block content mutation the way it blocks Array.prototype.push
+// on SEVERITY_ORDER below — it's applied for export consistency and to stop
+// a stray own-property assignment on the Set object itself.
+test("SEVERITY_ORDER, VALID_SEVERITIES, and NON_DEFECT_SEVERITIES are frozen, matching their frozen sibling exports", () => {
+  assert.equal(Object.isFrozen(SEVERITY_ORDER), true, "SEVERITY_ORDER must be frozen");
+  assert.equal(Object.isFrozen(VALID_SEVERITIES), true, "VALID_SEVERITIES must be frozen");
+  assert.equal(Object.isFrozen(NON_DEFECT_SEVERITIES), true, "NON_DEFECT_SEVERITIES must be frozen");
+  try {
+    assert.throws(() => { SEVERITY_ORDER.push("bogus"); }, TypeError);
```
mfittko, line 44: **nit** (`yagni`): The push-throws assertion is redundant with the Object.isFrozen assertion three lines above it. This module is an ESM file, so it is always strict mode, where Array.prototype.push on a frozen array always throws TypeError. The redundant assertion is the sole reason the test needs the surrounding try/finally plus the indexOf/splice cleanup at lines 50 to 52, so eleven lines of defensive test machinery exist to re-check a property already asserted. — judge: reject

packages/core/src/loop/gate-fanin.mjs
```diff
@@ -185,13 +185,23 @@ export const SEVERITY_ORDER = ["high", "question", "medium", "low", "nit"];
...
+// exports (GATE_CONFIG_KEY, LEGACY_SEVERITY_ALIASES, etc.), and it does stop
+// a caller from attaching a stray own property to the Set object itself.
+export const NON_DEFECT_SEVERITIES = Object.freeze(new Set(["question", "nit"]));
 
 // Marker gate name → gates.<key> config key. Owned here so every caller of
 // resolveFanoutGroups maps the same way; passing the marker name verbatim
 // resolves no groups and silently downgrades pairing enforcement.
 export const GATE_CONFIG_KEY = Object.freeze({ draft_gate: "draft", pre_approval_gate: "preApproval" });
-export const VALID_SEVERITIES = new Set(SEVERITY_ORDER);
+// Object.freeze on a Set only locks its OWN properties, not the add/delete
```
mfittko, line 199: **nit** (`yagni`): The five-line Set-freeze caveat is duplicated verbatim at lines 188 to 192 and again at lines 199 to 203, twelve lines apart in the same file, and a third paraphrase of it lives in the test comment. The two source copies say nothing different from each other. — judge: reject

mfittko (commented):


**Reviewed head SHA:** `6b7786596bd4536158de33635b0c04607257af56`
**Verdict:** clean

**Findings summary:** 5 angles reviewed; 8 findings (see per-angle breakdown below).

- `contradiction-lens` → `findings_present` (2 findings)
- `pr-checklist-matrix` → `findings_present` (4 findings)
- `yagni` → `findings_present` (2 findings)

> **low** (`pr-checklist-matrix`): The AC / DoD / Non-goals matrix has exactly five rows for the five issue-1615 acceptance criteria, but the diff carries four further behavior changes with no row anywhere: omittedCounts validation (array, known severity, positive integer count), per-finding judgeDisposition validation, the headSha sanitizes-to-empty refusal, and the entryPoint-parameterized refusal attribution that replaces the hardcoded function name at every throw site. These are disclosed only in the prose Scope section, so a reader auditing the matrix against the diff cannot anchor them to any acceptance criterion, any definition-of-done evidence item, or any declared non-goal boundary. — judge: act


> **low** (`pr-checklist-matrix`): The Scope and context section enumerates the added tests by name and the enumeration is incomplete. It names six pins (direct-export bypass, files sanitizes-to-empty, headSha sanitizes-to-empty on both entry points, sparse-files hole, angle bracketed-path legibility, frozen exports) while the diff adds fourteen tests. Eight go unnamed: four omittedCounts pins (unknown severity, non-array, non-object element, non-positive-integer count), three judgeDisposition pins (non-string or blank, sanitizes-to-empty, null accepted unchanged), and the bounded-path pin asserting the fail-closed throw reports the normalized gate and sanitized head. — judge: act


> **low** (`pr-checklist-matrix`): The body states two mutually inconsistent mutation-check scopes. Scope and context claims every change was mutation-checked (revert fails its pin, restore verified byte-identical); definition-of-done item two narrows the same claim to three items (guard routing, the sanitize-empty refusal, the hole-handling conversion). A reader cannot tell whether the omittedCounts, judgeDisposition, and entry-point-attribution changes were mutation-checked. The matrix compounds this with the only quantified evidence claim in the body, twenty tests failing on revert of the direct-export pin, which is not reproducible from the recorded validation artifact (it records only verify and schema:check at exit code zero). — judge: act


> **nit** (`pr-checklist-matrix`): Scope and context describes the new headSha guard as matching the same guard already applied to angle, summary, disposition, judgeDisposition, and files. The judgeDisposition and files sanitize-to-empty guards are introduced by this same diff, so already applied misdescribes two of the five as pre-existing. — judge: act

**Next action:** Apply the four judge-act PR-body edits, then merge under the recorded standing authorization once gate-evidence status is green

scripts/github/post-gate-findings.mjs
mfittko: Deferred at gate close (round 1, fingerprint 7770293e2d2a2292, severity low, angle contradiction-lens): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

scripts/github/post-gate-findings.mjs
mfittko: Deferred at gate close (round 1, fingerprint e185f33a72298c37, severity low, angle contradiction-lens): low findings are deferred at gate close after the fixer triaged them (fix-if-cheap-in-the-same-commit, else defer); the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/test/gate-fanin.test.mjs
mfittko: Deferred at gate close (round 1, fingerprint 5182256e707320c4, severity nit, angle yagni): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

packages/core/src/loop/gate-fanin.mjs
mfittko: Deferred at gate close (round 1, fingerprint 5cc7420f94bdfb69, severity nit, angle yagni): nit findings are deferred immediately at gate close, with no fixer cycle; the deferral is recorded on this thread's marker and in the durable findings ledger.

PR #1794, opened 2026-08-20
fix(gate): judge polish — qualified read-only wording, followUpDraft shape validation, pure-seam coverage check

### LINKED ISSUE #1656: Polish: judge agent description wording + validateFindingsArray followUpDraft shape + applyJudgeDispositions coverage gap (deferred from #1525)
## Summary

Three non-blocking findings from the pre-approval gate review of #1525 (judge agent). The gate returned clean (0 high); these are polish/nit-level hardening suggestions deferred to keep #1525 converging.

## Findings

1. **`agents/judge.agent.md` "read-only" wording** — the body already distinguishes "read-only over the repository" from the verdict-artifact write (lines 22-23), but two later mentions use bare "read-only" without the qualifier: "Your read-only boundary" (line 24) and the reviewer sentence at line 72. The frontmatter `tools:` list includes `write` and `bash`, so an unqualified "read-only" is imprecise. Qualify every remaining bare "read-only" that describes the judge, and keep the verdict artifact under `tmp/` stated as the sole permitted write. Nit-level wording clarity.

2. **`scripts/github/write-gate-findings-log.mjs` `validateFindingsArray` followUpDraft shape** — the validator accepts a `followUpDraft` object from the `--findings` / `--findings-file` input without validating its inner shape (lines 161-163). `validateJudgeVerdict` (`packages/core/src/loop/gate-fanin.mjs:910-913`) requires `title` non-empty string and `body` string. Mirror that rule at this trust boundary: a malformed draft throws via `parseError` with the `[i].followUpDraft` path, like every other field check in `validateFindingsArray`.

3. **`packages/core/src/loop/gate-fanin.mjs` `applyJudgeDispositions` coverage gap** — a finding with no corresponding disposition entry is silently left unenriched by the pure function (lines 947-966). Since #1525 merged, issue #1658 landed a fail-closed coverage check, but one layer up in `runJudgePass` (`scripts/loop/judge-pass.mjs:239-244`), so any other caller of the pure seam still gets silent omission. Two defects remain: (a) the check belongs in `applyJudgeDispositions` itself, next to the duplicate-index detection already in `validateJudgeVerdict` (gate-fanin.mjs:894-897); (b) the `runJudgePass` message maps `uncovered.map((f) => f.index)`, but enriched findings carry no `index` field, so it prints `undefined` instead of the uncovered positions. Move the check into the pure function, report 0-based array positions, and let `runJudgePass` inherit the behavior through the seam.

## Acceptance criteria

- [x] In `agents/judge.agent.md`, every mention of "read-only" that describes the judge is qualified as read-only over the repository (or equivalent), and the verdict artifact under `tmp/` remains stated as the sole permitted write. Lines 24 and 72 are the current bare mentions.
- [x] `validateFindingsArray` in `scripts/github/write-gate-findings-log.mjs` rejects a `followUpDraft` whose `title` is not a non-empty string or whose `body` is not a string; the rejection throws via `parseError` and names the `[i].followUpDraft` path.
- [x] A well-formed `followUpDraft` (`title` non-empty string, `body` string) passes through `validateFindingsArray` unchanged, as today.
- [x] `applyJudgeDispositions` throws when the dispositions do not cover every finding (any finding left without `judgeDisposition`); an empty findings array with an empty dispositions array still returns without error.
- [x] The coverage error message lists the uncovered findings by 0-based array position, not by a nonexistent `index` field on the finding objects.
- [x] `runJudgePass` keeps failing closed on an uncovered verdict through the shared seam; its local duplicate check in `scripts/loop/judge-pass.mjs` is removed or delegates, and the existing coverage test (`test/loop/judge-pass.test.mjs` line 108) still passes, updated only if its asserted message changes.
- [x] New tests exist: a coverage-gap throw plus the vacuous empty-array pass in `packages/core/test/gate-fanin.test.mjs`, and a malformed-`followUpDraft` rejection plus a valid pass-through in `test/github/write-gate-findings-log.test.mjs`.

## Definition of done

- [x] `node --test packages/core/test/gate-fanin.test.mjs` passes.
- [x] `node --test test/github/write-gate-findings-log.test.mjs` passes.
- [x] `node --test test/loop/judge-pass.test.mjs` passes.
- [x] Root `npm test` (run-p aggregate: assets, extension, scripts, core, docs, pack) is green.

## Non-goals

- Any change to the judge's authority split or the gate verdict semantics.
- Re-opening the #1525 gate pipeline.
- Changing the judge's frontmatter `tools:` list; this issue only fixes prose.
- Changing the judge verdict artifact schema or `validateJudgeVerdict` itself (its followUpDraft and duplicate-index rules are the reference, not the target).
- Re-doing the `runJudgePass`-layer coverage check from issue #1658; this issue relocates it into the pure seam and fixes its message.

Closes nothing at merge beyond itself — hardening only.

## Refined (2026-08-17)

Auto-grill against main 81e86581: all three sites verified in current code; finding 3 narrowed to reflect the coverage check issue #1658 already landed in `runJudgePass`.

mfittko:
## Summary

Lands the three polish items deferred from the judge-agent gate review: the judge doc's remaining bare "read-only" mentions are qualified as read-only over the repository (the tmp/ verdict artifact stays the sole permitted write); `validateFindingsArray` now validates a `followUpDraft`'s inner shape at the trust boundary (title non-empty string, body string) instead of accepting any object; and the judge-disposition coverage check moves from `runJudgePass` into the pure `applyJudgeDispositions` seam, failing closed for every caller and reporting uncovered findings by 0-based array position instead of a nonexistent `index` field.

## Scope and context

Twelve files (nine sources plus three regenerated .claude mirrors). `agents/judge.agent.md` (+ regenerated `.claude/agents/judge.md` mirror): prose-only qualification of two bare "read-only" mentions; frontmatter untouched. `scripts/github/write-gate-findings-log.mjs`: `validateFindingsArray` rejects a malformed or non-object `followUpDraft` via `parseError` naming the `[i].followUpDraft` path (presence-gated with a nullish-only carve-out, so a present-but-falsy draft is rejected rather than silently dropped), mirroring `validateJudgeVerdict`'s rule (which itself is unchanged); the `--judge-verdict` USAGE entry states the fail-closed dispose-every-finding precondition. `packages/core/src/loop/gate-fanin.mjs`: `applyJudgeDispositions` derives coverage from the verdict's own disposed-index set — an already-enriched ledger cannot satisfy the check by carried fields — throws when any finding is left undisposed (0-based positions), resets judge-owned fields (`followUpDraft`/`judgeCriterion`) before re-merge so a re-disposition drops stale prior-round state, and still passes empty findings with empty dispositions. `scripts/loop/judge-pass.mjs`: the local coverage check is removed; `runJudgePass` inherits the fail-closed behavior through the seam with the same error message. `skills/docs/gate-review-sub-loop-contract.md` and `skills/dev-loop/SKILL.md` (+ regenerated mirrors): the documented fail-closed enumerations gain the undisposed-finding trigger. Tests across `packages/core/test/gate-fanin.test.mjs`, `test/github/write-gate-findings-log.test.mjs`, and `test/loop/judge-pass.test.mjs`: coverage-gap throws (single and plural), vacuous empty-array pass, pre-enriched-ledger regression, stale-draft-reset regression, inherited `--judge-verdict` coverage throw with no ledger written, malformed/non-object/whitespace-title `followUpDraft` rejections, raw-store pass-through pin, and the judge-pass position-reporting regex extended through its indexes tail with non-coincident fixture indexes.

## Acceptance criteria

- [x] Every "read-only" describing the judge in `agents/judge.agent.md` is qualified as read-only over the repository; the tmp/ verdict artifact remains the sole permitted write.
- [x] `validateFindingsArray` rejects a `followUpDraft` whose `title` is not a non-empty string or whose `body` is not a string, throwing via `parseError` and naming the `[i].followUpDraft` path.
- [x] A well-formed `followUpDraft` passes through `validateFindingsArray` unchanged.
- [x] `applyJudgeDispositions` throws when dispositions do not cover every finding; empty findings + empty dispositions return without error.
- [x] The coverage error lists uncovered findings by 0-based array position.
- [x] `runJudgePass` keeps failing closed on an uncovered verdict through the shared seam; its local duplicate check is removed; the existing coverage test still passes.
- [x] New tests exist for the changed behavior: coverage-gap throws, empty-array pass, pre-enriched and stale-draft regressions (gate-fanin); followUpDraft rejections, pass-through, and inherited coverage-throw pins (write-gate-findings-log).

## Definition of done

- [x] `node --test packages/core/test/gate-fanin.test.mjs` passes (119 tests).
- [x] `node --test test/github/write-gate-findings-log.test.mjs` passes (96 tests).
- [x] `node --test test/loop/judge-pass.test.mjs` passes (11 tests).
- [x] Root `npm test` aggregate green (`npm run verify` covers the same suites; CI runs it on this PR).

|---|---|---|
| AC1: judge doc wording qualified | DoD 4 (verify aggregate covers test:assets, incl. mirror reproducibility) | no frontmatter change |
| AC2: malformed followUpDraft rejected | DoD 2 | validateJudgeVerdict unchanged |
| AC3: well-formed draft passes through | DoD 2 | no schema change |
| AC4: pure-seam coverage throw + vacuous pass | DoD 1 | no verdict-semantics change |
| AC5: 0-based positions in the error | DoD 1 | — |
| AC6: runJudgePass inherits fail-closed | DoD 3 | non-goal 4: no re-doing the runJudgePass-layer check |
| AC7: new tests exist for the changed behavior | DoD 1 + DoD 2 | — |

## Non-goals

- Any change to the judge's authority split or the gate verdict semantics.
- Changing the judge's frontmatter `tools:` list; prose only.
- Changing the judge verdict artifact schema or `validateJudgeVerdict` itself.
- Re-doing the `runJudgePass`-layer coverage check; this relocates it into the pure seam and fixes its message.

## Validation

```sh
node --test packages/core/test/gate-fanin.test.mjs
node --test test/github/write-gate-findings-log.test.mjs
node --test test/loop/judge-pass.test.mjs
npm run verify
```

Closes #1656

### DIFF
--- a/agents/judge.agent.md
+++ b/agents/judge.agent.md
@@ -21,7 +21,7 @@ You are the dedicated judge agent for the gate fan-out/fan-in chain. You hold th
 
 - You are **read-only over the repository**: you inspect code, the diff, the issue, and prior ledgers, but you never edit a tracked file. You have no `edit` tool.
 - The **only** thing you write is your own verdict artifact (to the deterministic path the conductor hands you, under `tmp/`). You do not write code, docs, comments, or any other file.
-- An actor that can fix will fix, and relevance judgment collapses into fixing. Your read-only boundary is what keeps relevance judgment independent of the fixing role.
+- An actor that can fix will fix, and relevance judgment collapses into fixing. Your read-only-over-the-repository boundary is what keeps relevance judgment independent of the fixing role.
 
 ## Inputs
 
@@ -69,7 +69,7 @@ Write a single JSON object to the deterministic path the conductor names (under 
 - **You do not soften `must-fix` on correctness grounds.** A real defect stays a real defect. You decide *where* it is fixed (this PR or a follow-up), not *whether* it is real. The fixer retains reproduction-based rejection; you do not override a finding's severity.
 - **You do not replace the severity-based disposition.** The severity-derived `disposition` (accepted-for-fix/deferred/needs-answer) stays intact; your `judgeDisposition` (act/defer/reject) is the relevance axis on top of it, not a replacement.
 - **You do not fix.** You have no `edit` tool and you write only your verdict artifact.
-- **You do not change reviewer fresh-context isolation.** Reviewers stay single-angle, read-only, and fresh-context by design.
+- **You do not change reviewer fresh-context isolation.** Reviewers stay single-angle, read-only over the repository, and fresh-context by design.
 
 ## Authority split (relevance vs reproduction)
 
--- a/packages/core/src/loop/gate-fanin.mjs
+++ b/packages/core/src/loop/gate-fanin.mjs
@@ -1016,7 +1016,10 @@ export function validateJudgeVerdict(verdict) {
  *
  * Pure. Fails closed (throws) when a disposition references an out-of-range
  * index — a judge verdict that names a finding that is not in the ledger is a
- * mismatch, never a silent enrichment.
+ * mismatch, never a silent enrichment — and when the dispositions do not
+ * cover every finding: an undisposed finding must never be silently dropped
+ * from the fixer's act list. An empty findings array with an empty
+ * dispositions array is vacuously covered and returns without error.
  *
  * @param {Array<object>} findings — the flat consolidated findings array
  * @param {object} judgeVerdict — the validated judge verdict artifact
@@ -1039,6 +1042,15 @@ export function applyJudgeDispositions(findings, judgeVerdict) {
     if (d.disposition === "defer" && d.followUpDraft) {
       target.followUpDraft = d.followUpDraft;
     }
+  }
+  const uncovered = enriched.reduce((positions, f, i) => {
+    if (!f.judgeDisposition) positions.push(i);
+    return positions;
+  }, /** @type {number[]} */ ([]));
+  if (uncovered.length > 0) {
+    throw new Error(
+      `judge verdict does not dispose ${uncovered.length} finding(s) (indexes: ${uncovered.join(", ")}) — fail closed; an undisposed finding must never be silently dropped from the fixer act list`
+    );
   }
   return { findings: enriched, scopeDrift: validated.scopeDrift };
 }
--- a/packages/core/src/loop/gate-fanin.mjs
+++ b/packages/core/src/loop/gate-fanin.mjs
@@ -1034,6 +1034,13 @@ export function applyJudgeDispositions(findings, judgeVerdict) {
       throw new Error(`judge disposition index ${d.index} is out of range (findings has ${enriched.length} entries)`);
     }
     const target = enriched[d.index];
+    // Reset judge-owned fields before the re-merge: a pre-enriched finding
+    // (already-enriched from a prior round, re-disposed by THIS verdict)
+    // must not let stale judgeCriterion/followUpDraft survive a
+    // defer -> act/reject re-disposition — the merged copy carries only
+    // what the current disposition provides, never prior-round residue.
+    delete target.judgeCriterion;
+    delete target.followUpDraft;
     target.judgeDisposition = d.disposition;
     target.judgeRationale = d.rationale;
     if (typeof d.criterion === "string" && d.criterion.trim().length > 0) {
--- a/packages/core/src/loop/gate-fanin.mjs
+++ b/packages/core/src/loop/gate-fanin.mjs
@@ -1043,8 +1043,14 @@ export function applyJudgeDispositions(findings, judgeVerdict) {
       target.followUpDraft = d.followUpDraft;
     }
   }
-  const uncovered = enriched.reduce((positions, f, i) => {
-    if (!f.judgeDisposition) positions.push(i);
+  // Coverage is judged against THIS verdict's disposed-index set, not field
+  // presence on the merged copy — an already-enriched ledger (a finding that
+  // already carries judgeDisposition from a prior round) must not let a
+  // verdict that disposes nothing pass silently. validateJudgeVerdict already
+  // rejects duplicate indexes, so the Set is exact.
+  const disposed = new Set(validated.dispositions.map((d) => d.index));
+  const uncovered = enriched.reduce((positions, _f, i) => {
+    if (!disposed.has(i)) positions.push(i);
     return positions;
   }, /** @type {number[]} */ ([]));
   if (uncovered.length > 0) {
--- a/packages/core/test/gate-fanin.test.mjs
+++ b/packages/core/test/gate-fanin.test.mjs
@@ -1255,6 +1255,19 @@ describe("applyJudgeDispositions (#1525)", () => {
     );
   });
 
+  test("re-disposing a pre-enriched defer-with-draft finding to act drops the stale followUpDraft", () => {
+    const alreadyEnriched = baseFindings.map((f, i) => (i === 1
+      ? { ...f, judgeDisposition: "defer", judgeRationale: "prior round", judgeCriterion: "NG-2", followUpDraft: { title: "stale", body: "stale body" } }
+      : { ...f }));
+    const verdict = judgeVerdict([
+      { index: 0, disposition: "act", rationale: "in-scope", criterion: "AC-1" },
+      { index: 1, disposition: "act", rationale: "now in-scope, no longer deferred", criterion: "AC-2" },
+    ]);
+    const { findings } = applyJudgeDispositions(alreadyEnriched, verdict);
+    assert.equal(findings[1].judgeDisposition, "act");
+    assert.equal(findings[1].followUpDraft, undefined);
+  });
+
   test("an empty findings array with an empty dispositions array is vacuously covered", () => {
     const verdict = judgeVerdict([]);
     const { findings } = applyJudgeDispositions([], verdict);
--- a/packages/core/test/gate-fanin.test.mjs
+++ b/packages/core/test/gate-fanin.test.mjs
@@ -1238,6 +1238,23 @@ describe("applyJudgeDispositions (#1525)", () => {
     );
   });
 
+  test("fails closed when a disposition leaves two findings uncovered, listing both positions", () => {
+    const verdict = judgeVerdict([]);
+    assert.throws(
+      () => applyJudgeDispositions(baseFindings, verdict),
+      /does not dispose 2 finding\(s\) \(indexes: 0, 1\)/,
+    );
+  });
+
+  test("fails closed on a pre-enriched ledger when the current verdict disposes nothing (coverage is judged against THIS verdict, not field presence)", () => {
+    const alreadyEnriched = baseFindings.map((f) => ({ ...f, judgeDisposition: "act", judgeRationale: "prior round" }));
+    const verdict = judgeVerdict([]);
+    assert.throws(
+      () => applyJudgeDispositions(alreadyEnriched, verdict),
+      /does not dispose 2 finding\(s\) \(indexes: 0, 1\)/,
+    );
+  });
+
   test("an empty findings array with an empty dispositions array is vacuously covered", () => {
     const verdict = judgeVerdict([]);
     const { findings } = applyJudgeDispositions([], verdict);
--- a/packages/core/test/gate-fanin.test.mjs
+++ b/packages/core/test/gate-fanin.test.mjs
@@ -1129,6 +1129,7 @@ describe("applyJudgeDispositions (#1525)", () => {
   test("a finding acted on against a named criterion gets judgeDisposition act", () => {
     const verdict = judgeVerdict([
       { index: 0, disposition: "act", rationale: "fixes the null-deref named in AC criterion 1", criterion: "AC-1: parser must not crash on null input" },
+      { index: 1, disposition: "reject", rationale: "style-only rename, out of scope", criterion: "Non-goal 3" },
     ]);
     const { findings, scopeDrift } = applyJudgeDispositions(baseFindings, verdict);
     assert.equal(findings[0].judgeDisposition, "act");
@@ -1164,7 +1165,10 @@ describe("applyJudgeDispositions (#1525)", () => {
 
   test("scope-drift verdict is raised when the diff exceeds the declared scope", () => {
     const verdict = judgeVerdict(
-      [{ index: 0, disposition: "act", rationale: "in-scope", criterion: "AC-1" }],
+      [
+        { index: 0, disposition: "act", rationale: "in-scope", criterion: "AC-1" },
+        { index: 1, disposition: "reject", rationale: "style-only rename, out of scope", criterion: "Non-goal 3" },
+      ],
       { verdict: "drift_detected", rationale: "the diff adds a new CLI flag not in any acceptance criterion; criterion 4 limits scope to the parser", driftedAreas: ["cli surface"] },
     );
     const { scopeDrift } = applyJudgeDispositions(baseFindings, verdict);
@@ -1222,5 +1226,21 @@ describe("applyJudgeDispositions (#1525)", () => {
     const logShape = toFindingsLogShape(baseFindings);
     assert.equal(logShape[0].judgeDisposition, undefined);
     assert.equal(logShape[0].followUpDraft, undefined);
+  });
+
+  test("fails closed when a disposition leaves a finding uncovered, naming its 0-based array position", () => {
+    const verdict = judgeVerdict([
+      { index: 0, disposition: "act", rationale: "in-scope", criterion: "AC-1" },
+    ]);
+    assert.throws(
+      () => applyJudgeDispositions(baseFindings, verdict),
+      /does not dispose 1 finding\(s\) \(indexes: 1\)/,
+    );
+  });
+
+  test("an empty findings array with an empty dispositions array is vacuously covered", () => {
+    const verdict = judgeVerdict([]);
+    const { findings } = applyJudgeDispositions([], verdict);
+    assert.deepEqual(findings, []);
   });
 });
--- a/scripts/github/write-gate-findings-log.mjs
+++ b/scripts/github/write-gate-findings-log.mjs
@@ -159,6 +159,14 @@ function validateFindingsArray(parsed, flagLabel) {
       entry.judgeCriterion = f.judgeCriterion.trim();
     }
     if (f.followUpDraft && typeof f.followUpDraft === "object" && !Array.isArray(f.followUpDraft)) {
+      // Mirrors validateJudgeVerdict's followUpDraft shape rule
+      // (packages/core/src/loop/gate-fanin.mjs): a non-empty title string and
+      // a body string, so this trust boundary never passes through a
+      // malformed draft that would surface later as a broken follow-up file.
+      const draft = f.followUpDraft;
+      if (typeof draft.title !== "string" || draft.title.trim().length === 0 || typeof draft.body !== "string") {
+        throw parseError(`${flagLabel}[${i}].followUpDraft must have a non-empty title and a body string`);
+      }
       entry.followUpDraft = f.followUpDraft;
     }
     return entry;
--- a/scripts/github/write-gate-findings-log.mjs
+++ b/scripts/github/write-gate-findings-log.mjs
@@ -35,8 +35,9 @@ Optional:
                                  enriched with the judge's relevance-based dispositions (judgeDisposition /
                                  judgeRationale / judgeCriterion / followUpDraft) via applyJudgeDispositions before
                                  the ledger is written, so the durable ledger and posted findings comment carry what
-                                 was consciously not acted on and why (#1525). Optional; when absent the ledger
-                                 writes byte-identically to before.
+                                 was consciously not acted on and why (#1525). The verdict must dispose every finding
+                                 (one disposition per 0-based ledger position) or the run FAILS CLOSED and writes no
+                                 ledger. Optional; when absent the ledger writes byte-identically to before.
   --tmp-root <path>              Root tmp directory (default: tmp/)
 
 ${JQ_OUTPUT_USAGE}
@@ -158,12 +159,13 @@ function validateFindingsArray(parsed, flagLabel) {
     if (typeof f.judgeCriterion === "string" && f.judgeCriterion.trim().length > 0) {
       entry.judgeCriterion = f.judgeCriterion.trim();
     }
-    if (f.followUpDraft) {
+    if (f.followUpDraft !== undefined && f.followUpDraft !== null) {
       // Mirrors validateJudgeVerdict's followUpDraft shape rule
-      // (packages/core/src/loop/gate-fanin.mjs): gate on presence first — a
-      // truthy non-object draft (string/array/number/boolean) is rejected
-      // here rather than silently dropped, then require a non-empty title
-      // string and a body string, so this trust boundary never passes
+      // (packages/core/src/loop/gate-fanin.mjs): gate on presence, carving
+      // out only nullish (undefined/null) as absent, so a present-but-falsy
+      // draft ("", 0, false) is rejected as a non-object rather than silently
+      // dropped like every other malformed draft, then require a non-empty
+      // title string and a body string, so this trust boundary never passes
       // through, or loses without a diagnostic, a malformed draft that would
       // surface later as a broken follow-up file.
       if (typeof f.followUpDraft !== "object" || Array.isArray(f.followUpDraft)) {
--- a/scripts/github/write-gate-findings-log.mjs
+++ b/scripts/github/write-gate-findings-log.mjs
@@ -158,15 +158,25 @@ function validateFindingsArray(parsed, flagLabel) {
     if (typeof f.judgeCriterion === "string" && f.judgeCriterion.trim().length > 0) {
       entry.judgeCriterion = f.judgeCriterion.trim();
     }
-    if (f.followUpDraft && typeof f.followUpDraft === "object" && !Array.isArray(f.followUpDraft)) {
+    if (f.followUpDraft) {
       // Mirrors validateJudgeVerdict's followUpDraft shape rule
-      // (packages/core/src/loop/gate-fanin.mjs): a non-empty title string and
-      // a body string, so this trust boundary never passes through a
-      // malformed draft that would surface later as a broken follow-up file.
+      // (packages/core/src/loop/gate-fanin.mjs): gate on presence first — a
+      // truthy non-object draft (string/array/number/boolean) is rejected
+      // here rather than silently dropped, then require a non-empty title
+      // string and a body string, so this trust boundary never passes
+      // through, or loses without a diagnostic, a malformed draft that would
+      // surface later as a broken follow-up file.
+      if (typeof f.followUpDraft !== "object" || Array.isArray(f.followUpDraft)) {
+        throw parseError(`${flagLabel}[${i}].followUpDraft must be an object`);
+      }
       const draft = f.followUpDraft;
       if (typeof draft.title !== "string" || draft.title.trim().length === 0 || typeof draft.body !== "string") {
         throw parseError(`${flagLabel}[${i}].followUpDraft must have a non-empty title and a body string`);
       }
+      // Validated against the trimmed title but stored raw (untrimmed): this
+      // mirrors validateJudgeVerdict's own raw pass-through, so a
+      // well-formed draft passes through unchanged rather than being
+      // silently renormalized.
       entry.followUpDraft = f.followUpDraft;
     }
     return entry;
@@ -431,7 +441,8 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
   // judge's relevance-based dispositions (act/defer/reject + rationale +
   // follow-up drafts) before writing the ledger (#1525). The judge runs after
   // fan-in and before the fix pass; applyJudgeDispositions is the pure merge
-  // seam that fails closed on a malformed verdict or an out-of-range index.
+  // seam that fails closed on a malformed verdict, an out-of-range index, or
+  // dispositions that do not cover every finding.
   let findings = rawFindings;
   let scopeDrift;
   if (options.judgeVerdict) {
@@ -497,8 +508,9 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
     // changes based on the judge artifact's content (#1745). A --judge-verdict
     // run can still surface a different, earlier error first: the judge
     // artifact is read, parsed, and applied before this point runs, so an
-    // unreadable, malformed, or out-of-range-index judge artifact throws its
-    // own parse error ahead of this contradiction check, not instead of it.
+    // unreadable, malformed, out-of-range-index, or incomplete-coverage judge
+    // artifact throws its own parse error ahead of this contradiction check,
+    // not instead of it.
     if (callerVerdict !== normalizedOverallVerdict) {
       throw parseError(
         `--verdict ${JSON.stringify(callerVerdict)} contradicts the wrapper's "overallVerdict" ${JSON.stringify(normalizedOverallVerdict)} (GATE-COMMENT-VERDICT-VALUES; skills/docs/gate-review-comment-contract.md) — the consolidator's computed round verdict, which judge dispositions from --judge-verdict never alter`,
--- a/scripts/loop/judge-pass.mjs
+++ b/scripts/loop/judge-pass.mjs
@@ -209,7 +209,7 @@ export function validateCliArgs(options) {
  *     (trim+lowercase) or the pass fails closed — a stale verdict staged from an
  *     earlier head must never feed the fixer's act list
  *   - applies the judge's relevance dispositions via `applyJudgeDispositions`
- *     (fail-closed on an out-of-range index)
+ *     (fail-closed on an out-of-range index and on undisposed findings)
  *
  * The fix pass consumes ONLY the returned `act` list (`GATE-EXEC-JUDGE-AUTHORITY-SPLIT`);
  * the returned `enriched` ledger carries `judgeDisposition` / `judgeRationale` /
@@ -231,17 +231,11 @@ export function runJudgePass(findings, judgeVerdict, headSha) {
       `judge verdict headSha ${JSON.stringify(validated.headSha)} does not match current head ${JSON.stringify(headSha)} — refuse a stale verdict; re-run the judge at the current head`
     );
   }
+  // applyJudgeDispositions fails closed on both an out-of-range index and an
+  // undisposed finding (the coverage check lives in that shared pure seam),
+  // so runJudgePass inherits fail-closed coverage without restating it here.
   const applied = applyJudgeDispositions(findings, judgeVerdict);
   const enriched = applied.findings;
-  // Fail closed when the judge verdict does not dispose every finding: an
-  // undisposed finding would otherwise be silently dropped from the fixer's act
-  // list (it only lingers in the enriched ledger without a disposition).
-  const uncovered = enriched.filter((f) => !f.judgeDisposition);
-  if (uncovered.length > 0) {
-    throw new Error(
-      `judge verdict does not dispose ${uncovered.length} finding(s) (indexes: ${uncovered.map((f) => f.index).join(", ")}) — fail closed; an undisposed finding must never be silently dropped from the fixer act list`
-    );
-  }
   const act = enriched.filter((f) => f.judgeDisposition === "act");
   const counts = { act: 0, defer: 0, reject: 0 };
   for (const f of enriched) {
--- a/skills/docs/gate-review-sub-loop-contract.md
+++ b/skills/docs/gate-review-sub-loop-contract.md
@@ -1128,7 +1128,8 @@ The shape is validated by `validateJudgeVerdict` (`@dev-loops/core/loop/gate-fan
 
 **Merge seam.** The conductor enriches the consolidated findings with the judge's
 dispositions via `applyJudgeDispositions(findings, judgeVerdict)` (`@dev-loops/core/loop/gate-fanin`,
-pure, fail-closed on a malformed verdict or an out-of-range index), then writes the durable
+pure, fail-closed on a malformed verdict, an out-of-range index, or dispositions that do not
+cover every finding), then writes the durable
 ledger via `write-gate-findings-log.mjs --judge-verdict <path>`. The enriched findings carry
 `judgeDisposition` / `judgeRationale` / `judgeCriterion` / `followUpDraft` so the disposition
 ledger and the posted findings comment show what was consciously not acted on and why
@@ -1144,7 +1145,7 @@ never feed the fixer), applies the dispositions via `applyJudgeDispositions`, an
 exactly the findings the judge marked `act` (`--out`) plus the enriched ledger
 (`--ledger-out`). The conductor hands that act list — never the full unfiltered ledger —
 to the fixer pass (`GATE-EXEC-JUDGE-AUTHORITY-SPLIT`). If `judge-pass` fails closed (stale
-head, malformed verdict, out-of-range index), the conductor re-runs the judge at the
+head, malformed verdict, out-of-range index, undisposed finding), the conductor re-runs the judge at the
 current head rather than degrading to severity-only disposition for a wired gate.
 
 <!-- rule: GATE-EXEC-JUDGE-AUTHORITY-SPLIT -->
--- a/test/github/write-gate-findings-log.test.mjs
+++ b/test/github/write-gate-findings-log.test.mjs
@@ -493,6 +493,51 @@ test("writeGateFindingsLog rejects a followUpDraft with a non-string body", asyn
   }, /\[0\]\.followUpDraft must have a non-empty title and a body string/);
 });
 
+test("writeGateFindingsLog rejects a followUpDraft with a missing title", async () => {
+  await assert.rejects(async () => {
+    await writeGateFindingsLog({
+      repo: "a/b",
+      pr: 1,
+      gate: "draft_gate",
+      headSha: "abc1234500000000000000000000000000000000",
+      verdict: "clean",
+      findings: JSON.stringify([
+        { severity: "low", angle: "docs", summary: "x", followUpDraft: { body: "b" } },
+      ]),
+    });
+  }, /\[0\]\.followUpDraft must have a non-empty title and a body string/);
+});
+
+test("writeGateFindingsLog rejects a non-object (string) followUpDraft instead of silently dropping it", async () => {
+  await assert.rejects(async () => {
+    await writeGateFindingsLog({
+      repo: "a/b",
+      pr: 1,
+      gate: "draft_gate",
+      headSha: "abc1234500000000000000000000000000000000",
+      verdict: "clean",
+      findings: JSON.stringify([
+        { severity: "low", angle: "docs", summary: "x", followUpDraft: "see the follow-up" },
+      ]),
+    });
+  }, /\[0\]\.followUpDraft must be an object/);
+});
+
+test("writeGateFindingsLog rejects an array-wrapped followUpDraft instead of silently dropping it", async () => {
+  await assert.rejects(async () => {
+    await writeGateFindingsLog({
+      repo: "a/b",
+      pr: 1,
+      gate: "draft_gate",
+      headSha: "abc1234500000000000000000000000000000000",
+      verdict: "clean",
+      findings: JSON.stringify([
+        { severity: "low", angle: "docs", summary: "x", followUpDraft: ["t", "b"] },
+      ]),
+    });
+  }, /\[0\]\.followUpDraft must be an object/);
+});
+
 test("writeGateFindingsLog passes a well-formed followUpDraft through unchanged", async () => {
   const tmpDir = await mkdtemp(path.join(os.tmpdir(), "gate-findings-followup-"));
   try {
@@ -1816,6 +1861,49 @@ test("writeGateFindingsLog enriches findings from --judge-verdict and records sc
     assert.equal(parsed.findings[1].judgeDisposition, "reject");
     assert.equal(parsed.scopeDrift.verdict, "drift_detected");
     assert.deepEqual(parsed.scopeDrift.driftedAreas, ["cli surface"]);
+  } finally {
+    await rm(tmpDir, { recursive: true, force: true });
+  }
+});
+
+// The --judge-verdict path inherits applyJudgeDispositions's coverage
+// fail-closed check the same as the pure seam and runJudgePass: a verdict
+// that leaves a finding undisposed must abort the write rather than persist
+// a ledger with a silently-dropped finding.
+test("writeGateFindingsLog --judge-verdict fails closed when the verdict does not dispose every finding and writes no ledger", async () => {
+  const tmpDir = await mkdtemp(path.join(os.tmpdir(), "gate-findings-judge-coverage-"));
+  try {
+    const headSha = "d4".repeat(20);
+    const judgeVerdictPath = path.join(tmpDir, "judge-verdict.json");
+    await writeFile(judgeVerdictPath, JSON.stringify({
+      headSha,
+      scopeDrift: { verdict: "within_scope", rationale: "matches the briefed AC set", driftedAreas: [] },
+      dispositions: [
+        { index: 0, disposition: "act", rationale: "fixes the defect named in AC-1", criterion: "AC-1" },
+      ],
+    }), "utf8");
+
+    await assert.rejects(
+      () => writeGateFindingsLog({
+        repo: "o/n",
+        pr: 23,
+        gate: "draft_gate",
+        headSha,
+        verdict: "findings_present",
+        findings: JSON.stringify([
+          { severity: "must-fix", angle: "correctness", summary: "null deref", disposition: "accepted-for-fix" },
+          { severity: "low", angle: "docs", summary: "rename variable", disposition: "deferred" },
+        ]),
+        judgeVerdict: judgeVerdictPath,
+        tmpRoot: tmpDir,
+      }),
+      /does not dispose 1 finding\(s\) \(indexes: 1\)/,
+    );
+    await assert.rejects(
+      () => readFile(buildLogPath({ repo: "o/n", pr: 23, gate: "draft_gate", headSha, tmpRoot: tmpDir }), "utf8"),
+      /ENOENT/,
+      "an undisposed-finding verdict must not write a ledger",
+    );
   } finally {
     await rm(tmpDir, { recursive: true, force: true });
   }
--- a/test/github/write-gate-findings-log.test.mjs
+++ b/test/github/write-gate-findings-log.test.mjs
@@ -463,6 +463,57 @@ test("writeGateFindingsLog rejects empty-string resolvedIn", async () => {
   }, /resolvedIn must be a non-empty string/);
 });
 
+test("writeGateFindingsLog rejects a followUpDraft with an empty title", async () => {
+  await assert.rejects(async () => {
+    await writeGateFindingsLog({
+      repo: "a/b",
+      pr: 1,
+      gate: "draft_gate",
+      headSha: "abc1234500000000000000000000000000000000",
+      verdict: "clean",
+      findings: JSON.stringify([
+        { severity: "low", angle: "docs", summary: "x", followUpDraft: { title: "", body: "b" } },
+      ]),
+    });
+  }, /\[0\]\.followUpDraft must have a non-empty title and a body string/);
+});
+
+test("writeGateFindingsLog rejects a followUpDraft with a non-string body", async () => {
+  await assert.rejects(async () => {
+    await writeGateFindingsLog({
+      repo: "a/b",
+      pr: 1,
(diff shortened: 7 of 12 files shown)


mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `612f7d26382c409ad93b86c7bc48a6c946ab6c18`
**Verdict:** clean

**Findings summary:** 9 angles reviewed; 13 findings (see per-angle breakdown below).

- `coverage` → `findings_present` (4 findings)
- `gate-evidence` → `clean`
- `link-check` → `clean`
- `pr-description` → `clean`
- `scope` → `clean`

> **low** (`contract-surface`): Documented fail-closed surface is now stale for the function this PR changes. Line 1131 describes applyJudgeDispositions as 'fail-closed on a malformed verdict or an out-of-range index', and line 1147 enumerates judge-pass's fail-closed triggers as '(stale head, malformed verdict, out-of-range index)'. Undisposed findings are now a third/fourth trigger, so a conductor reading the contract has no documented handling for a coverage-gap failure even though the doc already requires 'One disposition per finding' (line 1120). — judge: act
> Location: `skills/docs/gate-review-sub-loop-contract.md:1131`

**Next action:** Fix pass on the 11-item act list (coverage-check Set fix, followUpDraft guard, doc/comment syncs, test pins), then round 2.

packages/core/src/loop/gate-fanin.mjs
```diff
@@ -1040,6 +1043,15 @@ export function applyJudgeDispositions(findings, judgeVerdict) {
       target.followUpDraft = d.followUpDraft;
     }
   }
+  const uncovered = enriched.reduce((positions, f, i) => {
+    if (!f.judgeDisposition) positions.push(i);
```
mfittko: **medium** (`contract-surface`): The new coverage check tests !f.judgeDisposition on the enriched copy, not whether THIS verdict disposed the position. enriched is {...f} of the caller's findings, and a judge-enriched ledger already carries judgeDisposition (judge-pass --ledger-out emits that shape and --findings-file documents the same wrapper as input; write-gate-findings-log also carries the field through). Re-running judge-pass over an already-enriched ledger with a verdict that omits dispositions passes the check silently, and stale prior-round dispositions/rationales flow into the fixer act list — the exact silent-drop the check exists to prevent. The doc comment says it fails closed 'when the dispositions do not cover every finding', which disagrees with what the code checks. — judge: act

scripts/github/write-gate-findings-log.mjs
```diff
@@ -159,6 +159,14 @@ function validateFindingsArray(parsed, flagLabel) {
       entry.judgeCriterion = f.judgeCriterion.trim();
     }
     if (f.followUpDraft && typeof f.followUpDraft === "object" && !Array.isArray(f.followUpDraft)) {
+      // Mirrors validateJudgeVerdict's followUpDraft shape rule
+      // (packages/core/src/loop/gate-fanin.mjs): a non-empty title string and
+      // a body string, so this trust boundary never passes through a
+      // malformed draft that would surface later as a broken follow-up file.
+      const draft = f.followUpDraft;
+      if (typeof draft.title !== "string" || draft.title.trim().length === 0 || typeof draft.body !== "string") {
```
mfittko, line 175: **low** (`contract-surface`): The mirror of validateJudgeVerdict's followUpDraft rule is one-sided. The inner shape now rejects, but the outer guard on line 161 still means a truthy non-object draft (followUpDraft: "see the follow-up" or &#91;"t", "b"]) falls through and is silently dropped from the written entry, while validateJudgeVerdict (gate-fanin.mjs:985) rejects exactly that shape. A trust boundary that drops malformed input silently loses the deferred follow-up with no error. — judge: act

scripts/github/write-gate-findings-log.mjs
```diff
@@ -159,6 +159,14 @@ function validateFindingsArray(parsed, flagLabel) {
...
+        throw parseError(`${flagLabel}[${i}].followUpDraft must have a non-empty title and a body string`);
+      }
       entry.followUpDraft = f.followUpDraft;
```
mfittko, line 182: **low** (`contract-surface`): The durable-ledger validator still accepts a judgeDisposition: "defer" entry with no followUpDraft, while both validateJudgeVerdict (gate-fanin.mjs:984-987) and the contract doc (skills/docs/gate-review-sub-loop-contract.md:1122, 'defer ... MUST carry a followUpDraft') make the draft mandatory on defer. The documented soft-cap contract is enforced on the verdict artifact but not on the ledger that renders it, so a deferred finding can be written with nothing fileable attached. — judge: defer

packages/core/src/loop/gate-fanin.mjs
mfittko: **medium** (`correctness`): The relocated coverage check tests field PRESENCE on the merged copy (!f.judgeDisposition) rather than coverage by the CURRENT verdict, so an already-enriched ledger plus a verdict that disposes nothing passes silently. Verified by direct import: applyJudgeDispositions(&#91;{...judgeDisposition:'act'},{...judgeDisposition:'reject'}], {dispositions: &#91;]}) returns without throwing, and runJudgePass on the same input reports counts {act:1,defer:0,reject:1} and hands the fixer a 1-entry act list built entirely from a PRIOR round's dispositions. The shape is reachable through both callers: write-gate-findings-log's validateFindingsArray deliberately carries an input-supplied judgeDisposition through (scripts/github/write-gate-findings-log.mjs:148-155), judge-pass's validateFindingsArray returns finding objects unchanged (scripts/loop/judge-pass.mjs:266-276), and resolveFindingsInput unwraps ANY object carrying a findings array (scripts/github/_findings-input.mjs:22-27), so the durable enriched tmp/gate-findings ledger is an accepted --findings-file for judge-pass. AC4 states the seam throws when dispositions do not cover every finding; in this case they cover none and it does not throw, so the stale-judgment leak the check exists to stop still reaches the act list. — judge: act

packages/core/src/loop/gate-fanin.mjs
```diff
@@ -1016,7 +1016,10 @@ export function validateJudgeVerdict(verdict) {
  *
  * Pure. Fails closed (throws) when a disposition references an out-of-range
  * index — a judge verdict that names a finding that is not in the ledger is a
- * mismatch, never a silent enrichment.
+ * mismatch, never a silent enrichment — and when the dispositions do not
```
mfittko, line 1019: **low** (`correctness`): The seam's JSDoc gains the new fail-closed condition, but the three sibling sites that enumerate the same condition set were not updated and now understate the contract: skills/docs/gate-review-sub-loop-contract.md:1131 ("pure, fail-closed on a malformed verdict or an out-of-range index") and :1147 ("stale head, malformed verdict, out-of-range index"), plus scripts/github/write-gate-findings-log.mjs:434 and :500. The write-gate-findings-log sites matter most: that caller had NO coverage check before this PR, so its documented --judge-verdict behavior actually changed (a partial verdict now aborts the ledger write) while its own comments still list the old two conditions. — judge: act

packages/core/src/loop/gate-fanin.mjs
```diff
@@ -1040,6 +1043,15 @@ export function applyJudgeDispositions(findings, judgeVerdict) {
       target.followUpDraft = d.followUpDraft;
     }
   }
+  const uncovered = enriched.reduce((positions, f, i) => {
+    if (!f.judgeDisposition) positions.push(i);
+    return positions;
+  }, /** @type {number[]} */ ([]));
+  if (uncovered.length > 0) {
```
mfittko, line 1063: **medium** (`coverage`): The new coverage throw changes behavior for a second production caller, writeGateFindingsLog's --judge-verdict path (scripts/github/write-gate-findings-log.mjs:446), and no test pins it. The PR's central claim is that the check now fails closed for every caller, but the suites only cover the pure seam (packages/core/test/gate-fanin.test.mjs:1231) and runJudgePass (test/loop/judge-pass.test.mjs:108). The existing --judge-verdict tests (test/github/write-gate-findings-log.test.mjs:1784, 1833) all dispose every finding, so a regression that skips the coverage check on that path stays green. That caller's own doc comment (write-gate-findings-log.mjs:434) also still says the seam fails closed only on a malformed verdict or an out-of-range index. — judge: act

scripts/loop/judge-pass.mjs
```diff
@@ -231,17 +231,11 @@ export function runJudgePass(findings, judgeVerdict, headSha) {
       `judge verdict headSha ${JSON.stringify(validated.headSha)} does not match current head ${JSON.stringify(headSha)} — refuse a stale verdict; re-run the judge at the current head`
     );
   }
+  // applyJudgeDispositions fails closed on both an out-of-range index and an
+  // undisposed finding (the coverage check lives in that shared pure seam),
+  // so runJudgePass inherits fail-closed coverage without restating it here.
```
mfittko, line 236: **low** (`coverage`): The runJudgePass regression test for the fixed message cannot fail. test/loop/judge-pass.test.mjs:113 asserts only /does not dispose 1 finding\(s\)/, stopping before the "(indexes: ...)" segment that AC5 is about, and its fixture (line 111) sets index: 0 / index: 1 on the findings, so the buggy f.index mapping and the correct array position produce identical output. The vestigial index fields are now misleading: at this layer nothing distinguishes the pre-fix message from the post-fix one. — judge: act

packages/core/test/gate-fanin.test.mjs
```diff
@@ -1223,4 +1227,20 @@ describe("applyJudgeDispositions (#1525)", () => {
...
     assert.equal(logShape[0].followUpDraft, undefined);
   });
+
+  test("fails closed when a disposition leaves a finding uncovered, naming its 0-based array position", () => {
+    const verdict = judgeVerdict([
+      { index: 0, disposition: "act", rationale: "in-scope", criterion: "AC-1" },
+    ]);
+    assert.throws(
+      () => applyJudgeDispositions(baseFindings, verdict),
+      /does not dispose 1 finding\(s\) \(indexes: 1\)/,
```
mfittko, line 1237: **low** (`coverage`): Only the single-uncovered-finding case is exercised, so the message's plural count and the ", " position join are unpinned. Both existing coverage tests (this one and test/loop/judge-pass.test.mjs:108) leave exactly one finding undisposed, so a regression that reports only the first uncovered position, or joins without a separator, still passes. — judge: act

scripts/github/write-gate-findings-log.mjs
mfittko, line 175: **low** (`coverage`): The new followUpDraft guard's negative coverage is partial. The two added tests hit the empty-title branch and the non-string-body branch, leaving the typeof draft.title !== "string" branch (a missing or numeric title, the most likely real malformation) untested. More importantly, a non-object followUpDraft (string, number, array) never reaches the guard at all because of the enclosing typeof/isArray condition at line 161 and is silently dropped from the entry, with no test pinning either the drop or a rejection. The result is asymmetric strictness at the trust boundary: a malformed object draft now throws, while a malformed non-object draft still disappears without error. — judge: act

scripts/github/write-gate-findings-log.mjs
mfittko, line 182: **low** (`determinism`): The new followUpDraft check validates the TRIMMED title (draft.title.trim().length === 0, line 167) but line 170 persists f.followUpDraft verbatim, so a title like " Rename parser var " is accepted and stored untrimmed. Every sibling optional string in this same validator normalizes on the way in (judgeRationale line 156, judgeCriterion line 159, disposition, resolvedIn, files line 102), and the comment above entry.files names exactly this failure class: validation accepts the loose form while a consumer keying on the trimmed form silently disagrees at a different layer. Two follow-up drafts differing only in surrounding whitespace persist as distinct ledger values that the validator treats as identical, so any later title-keyed dedup/identity step (follow-up issue filing) is order/whitespace dependent rather than deterministic. — judge: reject

scripts/github/write-gate-findings-log.mjs
```diff
@@ -159,6 +159,14 @@ function validateFindingsArray(parsed, flagLabel) {
       entry.judgeCriterion = f.judgeCriterion.trim();
     }
     if (f.followUpDraft && typeof f.followUpDraft === "object" && !Array.isArray(f.followUpDraft)) {
+      // Mirrors validateJudgeVerdict's followUpDraft shape rule
+      // (packages/core/src/loop/gate-fanin.mjs): a non-empty title string and
+      // a body string, so this trust boundary never passes through a
+      // malformed draft that would surface later as a broken follow-up file.
+      const draft = f.followUpDraft;
```
mfittko, line 174: **low** (`input-validation`): The new inner-shape check runs only INSIDE the outer object guard at line 161, so the malformed shapes most likely to come from a producer bug never reach it: a non-object followUpDraft is silently dropped from the ledger instead of rejected. Verified against writeGateFindingsLog with followUpDraft set to "see issue 42", &#91;{title:'t',body:'b'}], 42, and true — all four wrote a ledger with followUpDraft: undefined and no error. The array case is the realistic one (a producer wrapping the draft), and the effect is a deferred finding losing its follow-up draft with no diagnostic, which is exactly what the added comment says this trust boundary must prevent. The reference rule diverges here: validateJudgeVerdict throws followUpDraft is required on a defer disposition for a non-object draft (packages/core/src/loop/gate-fanin.mjs:983-986). — judge: act

scripts/github/write-gate-findings-log.mjs
mfittko, line 175: **nit** (`input-validation`): Emptiness is judged on the TRIMMED title (draft.title.trim().length === 0) but the raw object is stored untrimmed at line 170, so title: " t " is accepted and persisted with its padding. Every neighboring field in this same validator normalizes to the trimmed value (judgeRationale, judgeCriterion, disposition, and files, whose comment explicitly warns about validating one form while consumers key on another). — judge: act

scripts/github/write-gate-findings-log.mjs
mfittko: Deferred (judge: defer): requiring a followUpDraft on a defer judgeDisposition in the durable-ledger validator is a new cross-field rule beyond this PR's shape mirror — tracked in follow-up issue 1795.

packages/core/src/loop/gate-fanin.mjs
mfittko: Fixed in 7c23e22b: same root fix as the contract-surface finding — the execution-confirmed bypass now throws; the fixer act list can no longer be built from prior-round dispositions.

packages/core/src/loop/gate-fanin.mjs
mfittko: Fixed in 7c23e22b: the fail-closed enumerations in gate-review-sub-loop-contract.md (both listed lines) and the stale comments in write-gate-findings-log.mjs now name the undisposed-finding trigger; mirror regenerated.

packages/core/src/loop/gate-fanin.mjs
mfittko: Fixed in 7c23e22b: applyJudgeDispositions now derives coverage from a Set of the verdict's disposed indexes instead of probing judgeDisposition on the merged copy, so an already-enriched ledger with an empty-disposition verdict throws instead of passing stale dispositions through. Regression test pins the pre-enriched + empty-dispositions case.

scripts/github/write-gate-findings-log.mjs
mfittko: Fixed in 7c23e22b: validateFindingsArray now rejects any truthy non-object/array followUpDraft via parseError naming the [i].followUpDraft path, closing the silent-drop asymmetry with validateJudgeVerdict; absent stays a no-op.

packages/core/src/loop/gate-fanin.mjs
mfittko: Fixed in 7c23e22b: a test now proves writeGateFindingsLog --judge-verdict inherits the coverage throw (undisposed finding fails the write, no ledger file created), covering the second production caller the PR's every-caller claim names.

scripts/loop/judge-pass.mjs
mfittko: Fixed in 7c23e22b: the judge-pass regression regex extends through the (indexes: ...) tail and the fixture's index fields (7/9) no longer coincide with array positions, so a regression to f.index is distinguishable.

packages/core/test/gate-fanin.test.mjs
mfittko: Fixed in 7c23e22b: a two-finding dispose-nothing case pins the plural count and the comma-space join.

scripts/github/write-gate-findings-log.mjs
mfittko: Fixed in 7c23e22b: the missing/non-string-title branch and the non-object rejection (string and array) are pinned by new tests.

scripts/github/write-gate-findings-log.mjs
mfittko: Rejected (judge: reject): storing the trimmed title would violate the pass-through-unchanged acceptance criterion; validation trims for the check while storing the raw object deliberately, and a comment now states the split.

scripts/github/write-gate-findings-log.mjs
mfittko: Fixed in 7c23e22b: same guard completion — non-object drafts (string/array/number/boolean) are rejected instead of silently dropped.

scripts/github/write-gate-findings-log.mjs
mfittko: Fixed in 7c23e22b: a one-line comment states the validate-trimmed/store-raw split is deliberate parity with validateJudgeVerdict.

scripts/github/write-gate-findings-log.mjs
```diff
@@ -423,7 +441,8 @@ export async function writeGateFindingsLog(options, { repoRoot = process.cwd() }
   // judge's relevance-based dispositions (act/defer/reject + rationale +
   // follow-up drafts) before writing the ledger (#1525). The judge runs after
   // fan-in and before the fix pass; applyJudgeDispositions is the pure merge
-  // seam that fails closed on a malformed verdict or an out-of-range index.
+  // seam that fails closed on a malformed verdict, an out-of-range index, or
+  // dispositions that do not cover every finding.
```
mfittko, line 447: **low** (`contract-surface`): The new --judge-verdict fail-closed precondition (the verdict must dispose EVERY finding or the ledger write aborts) is recorded only in this internal comment. The CLI USAGE entry for --judge-verdict (lines 34-39) still presents the flag as purely additive enrichment ("Optional; when absent the ledger writes byte-identically to before") and names no failure mode, so a caller reading --help sees no reason a --judge-verdict run can now abort with no ledger written. The sibling CLI judge-pass.mjs documents its own analogous fail-closed rule in USAGE ("must be current-head ... or the pass FAILS CLOSED"), so the two CLI surfaces now describe the same seam at different fidelity. — judge: act

scripts/github/write-gate-findings-log.mjs
```diff
@@ -158,7 +158,25 @@ function validateFindingsArray(parsed, flagLabel) {
     if (typeof f.judgeCriterion === "string" && f.judgeCriterion.trim().length > 0) {
       entry.judgeCriterion = f.judgeCriterion.trim();
     }
-    if (f.followUpDraft && typeof f.followUpDraft === "object" && !Array.isArray(f.followUpDraft)) {
+    if (f.followUpDraft) {
```
mfittko: **low** (`contract-surface`): The presence gate is truthiness-based, so a falsy non-object draft (followUpDraft: "", 0, or false) still falls through both new checks and is dropped with no diagnostic, while the comment two lines below claims this trust boundary "never passes through, or loses without a diagnostic, a malformed draft". The claim is broader than the code: only truthy malformed drafts are rejected. Sibling optional fields in the same function use presence gating ("line" in f, "disposition" in f, "resolvedIn" in f), which does reject those values. — judge: act

skills/docs/gate-review-sub-loop-contract.md
```diff
@@ -1128,7 +1128,8 @@ The shape is validated by `validateJudgeVerdict` (`@dev-loops/core/loop/gate-fan
 
 **Merge seam.** The conductor enriches the consolidated findings with the judge's
 dispositions via `applyJudgeDispositions(findings, judgeVerdict)` (`@dev-loops/core/loop/gate-fanin`,
-pure, fail-closed on a malformed verdict or an out-of-range index), then writes the durable
+pure, fail-closed on a malformed verdict, an out-of-range index, or dispositions that do not
```
mfittko, line 1131: **low** (`correctness`): PR description scope statement understates the change: it says "Seven files" and enumerates per-file changes, but the diff touches 10. The contract-doc updates (skills/docs/gate-review-sub-loop-contract.md + its .claude mirror) and the tightened assertion in test/loop/judge-pass.test.mjs are not mentioned in Scope and context, so the doc edits under review have no declared scope entry. — judge: act

scripts/github/write-gate-findings-log.mjs
```diff
@@ -158,7 +158,25 @@ function validateFindingsArray(parsed, flagLabel) {
...
+      // truthy non-object draft (string/array/number/boolean) is rejected
+      // here rather than silently dropped, then require a non-empty title
+      // string and a body string, so this trust boundary never passes
+      // through, or loses without a diagnostic, a malformed draft that would
+      // surface later as a broken follow-up file.
+      if (typeof f.followUpDraft !== "object" || Array.isArray(f.followUpDraft)) {
+        throw parseError(`${flagLabel}[${i}].followUpDraft must be an object`);
+      }
+      const draft = f.followUpDraft;
+      if (typeof draft.title !== "string" || draft.title.trim().length === 0 || typeof draft.body !== "string") {
```
mfittko, line 175: **low** (`coverage`): The two followUpDraft contracts the new comment declares are unpinned: (a) the title is validated TRIMMED, so a whitespace-only title must be rejected, and (b) the draft is stored RAW/untrimmed. Every new test uses title "t" with no surrounding whitespace, so dropping .trim() from the line-173 guard, or trimming the title on the line-180 store, leaves the whole suite green. — judge: act

packages/core/src/loop/gate-fanin.mjs
```diff
@@ -1040,6 +1043,21 @@ export function applyJudgeDispositions(findings, judgeVerdict) {
       target.followUpDraft = d.followUpDraft;
```
mfittko, line 1050: **low** (`determinism`): On the pre-enriched-ledger path the new coverage check now defends, prior-round judge state survives the re-merge: judgeDisposition/judgeRationale are always overwritten, but followUpDraft is only assigned when the new disposition is defer, and judgeCriterion only when the new disposition carries a non-empty criterion. A finding that was defer+followUpDraft last round and is act/reject this round keeps the stale followUpDraft (and possibly a stale judgeCriterion), so the written ledger pairs this round's disposition with last round's follow-up draft. The path is reachable: validateFindingsArray in scripts/github/write-gate-findings-log.mjs (lines 148-180) accepts judgeDisposition/judgeRationale/judgeCriterion/followUpDraft on input, and the same call can pass --judge-verdict. — judge: act

packages/core/src/loop/gate-fanin.mjs
```diff
@@ -1040,6 +1043,21 @@ export function applyJudgeDispositions(findings, judgeVerdict) {
...
     }
   }
+  // Coverage is judged against THIS verdict's disposed-index set, not field
+  // presence on the merged copy — an already-enriched ledger (a finding that
+  // already carries judgeDisposition from a prior round) must not let a
+  // verdict that disposes nothing pass silently. validateJudgeVerdict already
+  // rejects duplicate indexes, so the Set is exact.
+  const disposed = new Set(validated.dispositions.map((d) => d.index));
+  const uncovered = enriched.reduce((positions, _f, i) => {
```
mfittko, line 1059: **low** (`input-validation`): The new coverage check is computed over the coerced list, so a non-array findings argument satisfies it vacuously. Line 1030 still does const list = Array.isArray(findings) ? findings : &#91;], so passing the ledger WRAPPER ({overallVerdict, findings}) or any non-array with an empty dispositions array yields enriched = &#91;], an empty uncovered set, and a silent clean return instead of the fail-closed error this PR adds. Both in-repo callers validate the array first (write-gate-findings-log via _findings-input.mjs, judge-pass via its own validateFindingsArray), so it is not reachable today, but the seam's own fail-closed-for-every-caller guarantee has this hole. — judge: defer

scripts/github/write-gate-findings-log.mjs
mfittko: **low** (`input-validation`): The new followUpDraft check gates on truthiness, so a present-but-falsy malformed draft still slips through silently: followUpDraft: "", 0, or false is dropped with no diagnostic, while followUpDraft: "see the follow-up" now throws. This is also inconsistent with this same validator's convention for other optional fields ("line" in f, "disposition" in f, "resolvedIn" in f), which gate on key presence and reject any malformed value. — judge: act

packages/core/test/gate-fanin.test.mjs
```diff
@@ -1223,4 +1227,37 @@ describe("applyJudgeDispositions (#1525)", () => {
...
+    );
+  });
+
+  test("fails closed when a disposition leaves two findings uncovered, listing both positions", () => {
```
mfittko, line 1241: **low** (`pr-description`): Definition-of-done items 1 and 2 carry stale test counts written before the round-2 commit 7c23e22b. DoD 1 claims packages/core/test/gate-fanin.test.mjs passes with 116 tests, but round 2 added two tests (two-findings-uncovered at line 1241 and the pre-enriched-ledger case), so the file now declares 118. DoD 2 claims test/github/write-gate-findings-log.test.mjs passes with 90 tests, but round 2 added four (missing title, non-object string, array-wrapped, --judge-verdict coverage fail-closed), so it now declares 94. A verifier checking the DoD literally against a run at this head sees a mismatch on both rows. — judge: act

test/github/write-gate-findings-log.test.mjs
```diff
@@ -463,6 +463,102 @@ test("writeGateFindingsLog rejects empty-string resolvedIn", async () => {
...
+      headSha: "abc1234500000000000000000000000000000000",
+      verdict: "clean",
+      findings: JSON.stringify([
+        { severity: "low", angle: "docs", summary: "x", followUpDraft: { title: "t", body: 42 } },
+      ]),
+    });
+  }, /\[0\]\.followUpDraft must have a non-empty title and a body string/);
+});
+
+test("writeGateFindingsLog rejects a followUpDraft with a missing title", async () => {
```
mfittko, line 496: **nit** (`pr-description`): AC7 and the matrix row "AC7: the four new tests exist" still describe four new tests. The head adds eleven (four in gate-fanin.test.mjs, seven in write-gate-findings-log.test.mjs including the non-object/array-wrapped rejections at line 496 onward and the --judge-verdict coverage case). The AC is satisfied (it states a minimum), but the count wording is stale. — judge: act

skills/docs/gate-review-sub-loop-contract.md
```diff
@@ -1128,7 +1128,8 @@ The shape is validated by `validateJudgeVerdict` (`@dev-loops/core/loop/gate-fan
...
+cover every finding), then writes the durable
```
mfittko, line 1132: **low** (`scope`): The PR body's Scope section is stale against the current head: it says "Seven files" while the diff touches 10. The contract-doc sync (skills/docs/gate-review-sub-loop-contract.md and its .claude/skills mirror) appears nowhere in the body, and test/loop/judge-pass.test.mjs appears only in the Definition of done and Validation blocks, never in the Scope enumeration. Every file is justified (the doc enumerates applyJudgeDispositions's fail-closed conditions and had to gain the coverage case; AC6 explicitly allows the judge-pass test update because the asserted message changed), so this is a description-contract gap, not out-of-scope code. — judge: act

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `7c23e22baaa181064a79690cbab4a2f2f00770cc`
**Verdict:** clean

**Findings summary:** 9 angles reviewed; 11 findings (see per-angle breakdown below).

- `gate-evidence` → `clean`
- `link-check` → `clean`

> **low** (`correctness`): Stale twin enumeration of judge-pass fail-closed cases. This diff updated skills/docs/gate-review-sub-loop-contract.md:1148 to add "undisposed finding" to the fail-closed list, but the operator-facing copy of the same list in skills/dev-loop/SKILL.md:197 still reads "(stale verdict head, malformed verdict, out-of-range index)". A conductor that hits the new coverage failure reads the inline dispatch guidance and finds no sanctioned recovery for it. — judge: act
> Location: `skills/dev-loop/SKILL.md:197`

**Next action:** Final in-window fix pass (USAGE line, presence gate, SKILL twin sync, trim pins, stale-draft reset) + body refresh, then round 3 and draft-to-ready.

packages/core/test/gate-fanin.test.mjs
mfittko: Fixed via PR-body edit: the DoD counts are refreshed from real runs (119 / 96 / 11).

test/github/write-gate-findings-log.test.mjs
mfittko: Fixed via PR-body edit: the fixed-number wording is replaced by an enumerating description of the added pins.

skills/docs/gate-review-sub-loop-contract.md
mfittko: Fixed via PR-body edit: the Scope section now says ten files and declares the contract-doc pair, the dev-loop SKILL twin, and the judge-pass test.

scripts/github/write-gate-findings-log.mjs
mfittko: Fixed in 0497776f: the --judge-verdict USAGE entry now states the fail-closed precondition — the verdict must dispose every finding or the run fails closed and writes no ledger — mirroring judge-pass's wording.

scripts/github/write-gate-findings-log.mjs
mfittko: Fixed in 0497776f: the followUpDraft gate is presence-based with a nullish-only carve-out, so a present-but-falsy draft (empty string, 0, false) is rejected as a non-object instead of silently dropped.

skills/docs/gate-review-sub-loop-contract.md
mfittko: Fixed in 0497776f: the skills/dev-loop/SKILL.md twin enumeration now names the undisposed-finding trigger, matching the contract doc; mirrors regenerated.

scripts/github/write-gate-findings-log.mjs
mfittko: Fixed in 0497776f: two pins added — a whitespace-only title is rejected (validate-trimmed) and a padded title is stored raw in the written ledger (store-raw).

packages/core/src/loop/gate-fanin.mjs
mfittko: Fixed in 0497776f: applyJudgeDispositions resets judge-owned fields (followUpDraft/judgeCriterion) before the re-merge, so a defer-to-act re-disposition drops the stale draft; a regression test pins it.

packages/core/src/loop/gate-fanin.mjs
mfittko: Deferred (judge: defer): rejecting a non-array findings argument instead of coercing to [] is an input-contract change beyond this PR's scope — appended to follow-up issue 1795.

scripts/github/write-gate-findings-log.mjs
mfittko: Fixed in 0497776f: same presence-gate fix — falsy drafts are rejected, never silently dropped.

packages/core/test/gate-fanin.test.mjs
```diff
@@ -1223,4 +1227,50 @@ describe("applyJudgeDispositions (#1525)", () => {
...
+    const alreadyEnriched = baseFindings.map((f, i) => (i === 1
+      ? { ...f, judgeDisposition: "defer", judgeRationale: "prior round", judgeCriterion: "NG-2", followUpDraft: { title: "stale", body: "stale body" } }
+      : { ...f }));
+    const verdict = judgeVerdict([
+      { index: 0, disposition: "act", rationale: "in-scope", criterion: "AC-1" },
+      { index: 1, disposition: "act", rationale: "now in-scope, no longer deferred", criterion: "AC-2" },
+    ]);
+    const { findings } = applyJudgeDispositions(alreadyEnriched, verdict);
+    assert.equal(findings[1].judgeDisposition, "act");
+    assert.equal(findings[1].followUpDraft, undefined);
```
mfittko, line 1268: **low** (`correctness`): The round-2 stale-reset test pins only the followUpDraft delete. delete target.judgeCriterion (packages/core/src/loop/gate-fanin.mjs:1042) is effectively untested: this test's re-disposition supplies criterion: "AC-2", which would overwrite the stale "NG-2" even without the delete. Since criterion is OPTIONAL in validateJudgeVerdict (only rationale is required), the reachable case a prior-round finding carrying judgeCriterion re-disposed by a criterion-less disposition has no pin, so removing the judgeCriterion delete would not fail any suite. — judge: defer

scripts/github/write-gate-findings-log.mjs
```diff
@@ -158,7 +159,26 @@ function validateFindingsArray(parsed, flagLabel) {
     if (typeof f.judgeCriterion === "string" && f.judgeCriterion.trim().length > 0) {
       entry.judgeCriterion = f.judgeCriterion.trim();
     }
-    if (f.followUpDraft && typeof f.followUpDraft === "object" && !Array.isArray(f.followUpDraft)) {
+    if (f.followUpDraft !== undefined && f.followUpDraft !== null) {
+      // Mirrors validateJudgeVerdict's followUpDraft shape rule
```
mfittko, line 163: **nit** (`correctness`): The added comment says the block "Mirrors validateJudgeVerdict's followUpDraft shape rule ... gate on presence, carving out only nullish (undefined/null) as absent". Only the inner title/body predicate mirrors validateJudgeVerdict; the presence gate does not. validateJudgeVerdict (gate-fanin.mjs:984-992) gates on disposition === "defer", REQUIRES a draft there (nullish is rejected, not treated as absent), and applies no shape check at all on act/reject. The new rule here is strictly stricter than the cited reference, not a mirror of its presence gate. — judge: defer

scripts/github/write-gate-findings-log.mjs
```diff
@@ -158,7 +159,26 @@ function validateFindingsArray(parsed, flagLabel) {
     if (typeof f.judgeCriterion === "string" && f.judgeCriterion.trim().length > 0) {
       entry.judgeCriterion = f.judgeCriterion.trim();
     }
-    if (f.followUpDraft && typeof f.followUpDraft === "object" && !Array.isArray(f.followUpDraft)) {
+    if (f.followUpDraft !== undefined && f.followUpDraft !== null) {
```
mfittko, line 162: **low** (`coverage`): The nullish-only carve-out is unpinned on both sides. The presence gate deliberately replaces the old truthiness gate (f.followUpDraft && ...) so that a present-but-falsy draft ("", 0, false) is REJECTED rather than silently dropped, while null/undefined stay treated as absent — the comment at lines 163-169 calls this out as the design decision. No test supplies a falsy-present draft, and no test supplies followUpDraft: null. The landed tests only cover a truthy non-object string and an array, both of which a regression back to if (f.followUpDraft) { if (typeof ... !== "object") throw } would still reject. That regression silently drops ""/0/false and passes the entire current suite. — judge: defer

packages/core/src/loop/gate-fanin.mjs
```diff
@@ -1031,6 +1034,13 @@ export function applyJudgeDispositions(findings, judgeVerdict) {
       throw new Error(`judge disposition index ${d.index} is out of range (findings has ${enriched.length} entries)`);
     }
     const target = enriched[d.index];
+    // Reset judge-owned fields before the re-merge: a pre-enriched finding
+    // (already-enriched from a prior round, re-disposed by THIS verdict)
+    // must not let stale judgeCriterion/followUpDraft survive a
+    // defer -> act/reject re-disposition — the merged copy carries only
+    // what the current disposition provides, never prior-round residue.
+    delete target.judgeCriterion;
+    delete target.followUpDraft;
```
mfittko, line 1043: **low** (`coverage`): The added delete statements make applyJudgeDispositions's documented purity ("Pure", JSDoc line 1017; shallow copy at line 1031) newly destructive to the caller's array if the copy ever regresses, yet no test pins input non-mutation. Before this PR the function only ADDED fields, so an in-place regression (const enriched = list;) was benign; now the same regression would strip judgeCriterion/followUpDraft off the CALLER's prior-round ledger objects. The stale-draft regression test asserts only findings&#91;1].followUpDraft === undefined, which holds identically when findings and the input array are the same object, so the whole suite passes through that regression. — judge: defer

packages/core/test/gate-fanin.test.mjs
mfittko, line 1268: **low** (`determinism`): The stale-reset regression pins only half the reset: delete target.judgeCriterion (packages/core/src/loop/gate-fanin.mjs:1042) is unpinned. In this test the new disposition for index 1 supplies criterion: "AC-2", so judgeCriterion is overwritten by the assignment regardless of the delete; every other applyJudgeDispositions test also supplies a criterion on every disposition. No test covers a pre-enriched judgeCriterion re-disposed by a disposition WITHOUT criterion, so removing the delete keeps the whole suite green while the enriched ledger silently carries prior-round criterion text — output that depends on hidden prior-round state rather than on the current verdict alone. — judge: defer

scripts/github/write-gate-findings-log.mjs
mfittko, line 162: **low** (`input-validation`): Validation drift inside one trust boundary: followUpDraft now fails closed on any present-but-malformed value, while its three sibling judge fields immediately above (judgeDisposition line 149, judgeRationale line 156, judgeCriterion line 159) still gate on typeof === "string" and SILENTLY DROP a non-string value. So { judgeDisposition: 42 } or { judgeRationale: {...} } writes a ledger missing that field with no diagnostic, which is exactly the harm class the new followUpDraft comment (lines 167-174) cites as the reason to reject rather than drop. Same reachability class as the newly-guarded field (hand-authored/foreign --findings input; the judge path is pre-validated by validateJudgeVerdict). — judge: defer

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `0497776fe9de6ede7027676789c23d6871a9ab3d`
**Verdict:** clean

**Findings summary:** 9 angles reviewed; 8 findings (see per-angle breakdown below).

- `gate-evidence` → `clean`
- `link-check` → `clean`
- `pr-description` → `clean`

> **low** (`contract-surface`): The judge-pass CLI's own USAGE surface never gained the coverage precondition this PR adds, while every sibling surface did. Round 2 fixed exactly this class for the twin CLI (write-gate-findings-log.mjs:38-40 now states "The verdict must dispose every finding ... or the run FAILS CLOSED and writes no ledger"), and the doc/SKILL/JSDoc enumerations gained "undisposed finding" (skills/docs/gate-review-sub-loop-contract.md:1148, skills/dev-loop/SKILL.md:197, scripts/loop/judge-pass.mjs:212). judge-pass is the command whose sole job is applying the dispositions, yet its --judge-verdict entry (lines 30-37) still names only shape validation and current-head freshness as fail-closed causes, and the exit-code list at line 56 reads "stale verdict head, malformed verdict, out-of-range disposition, etc." — the open "etc." keeps it from being false, but an operator reading --help for the command that now aborts on a partial verdict sees no statement of the dispose-every-finding requirement. — judge: defer
> Location: `scripts/loop/judge-pass.mjs:56`


> **nit** (`scope`): The scope section opens with "Ten files." but the diff at this head touches twelve (nine source files plus three regenerated .claude mirrors). Ten is the count of files named explicitly in the prose; the two skill-doc mirrors are only referenced generically as "(+ regenerated mirrors)", so the declared count matches neither the 12-file diff nor the 9 logical (non-mirror) units. Every changed file does belong to the declared scope; only the numeric prefix is inconsistent. — judge: act

**Next action:** Window closed: body numeric fix, deferrals filed (pins, validator convergence, judge-pass USAGE), then draft-to-ready.

packages/core/src/loop/gate-fanin.mjs
mfittko: Deferred (judge: defer, window closed): tracked in follow-up issue 1796 (test pins for the judge-disposition seam and followUpDraft carve-out).

packages/core/test/gate-fanin.test.mjs
mfittko: Deferred (judge: defer, window closed): the unpinned judgeCriterion reset is tracked in follow-up issue 1796.

scripts/github/write-gate-findings-log.mjs
mfittko: Deferred (judge: defer, window closed): converging the sibling judge-field validation is tracked in follow-up issue 1795 (second append recorded there).

packages/core/test/gate-fanin.test.mjs
mfittko: Deferred (judge: defer, window closed): folded into follow-up issue 1795's validator-convergence scope (second append recorded there).

scripts/github/write-gate-findings-log.mjs
mfittko: Deferred (judge: defer, window closed): the comment's mirror overclaim travels with follow-up issue 1795's validator-convergence work (second append recorded there).

scripts/github/write-gate-findings-log.mjs
mfittko: Deferred (judge: defer, window closed): tracked in follow-up issue 1796 (input non-mutation pin now that the deletes make purity load-bearing).

scripts/github/write-gate-findings-log.mjs
```diff
@@ -158,7 +159,26 @@ function validateFindingsArray(parsed, flagLabel) {
...
+      // Mirrors validateJudgeVerdict's followUpDraft shape rule
+      // (packages/core/src/loop/gate-fanin.mjs): gate on presence, carving
+      // out only nullish (undefined/null) as absent, so a present-but-falsy
+      // draft ("", 0, false) is rejected as a non-object rather than silently
+      // dropped like every other malformed draft, then require a non-empty
```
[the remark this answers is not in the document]
mfittko: Agreed, and already triaged: the round-3 gate judge deferred exactly this comment correction to follow-up issue 1795 (validator-convergence work, second append), because a doc-comment-only commit at the merge boundary would stale the posted gate verdicts. The comment will state the stricter nullish-only carve-out there.

agents/judge.agent.md
```diff
@@ -69,7 +69,7 @@ Write a single JSON object to the deterministic path the conductor names (under
 - **You do not soften `must-fix` on correctness grounds.** A real defect stays a real defect. You decide *where* it is fixed (this PR or a follow-up), not *whether* it is real. The fixer retains reproduction-based rejection; you do not override a finding's severity.
 - **You do not replace the severity-based disposition.** The severity-derived `disposition` (accepted-for-fix/deferred/needs-answer) stays intact; your `judgeDisposition` (act/defer/reject) is the relevance axis on top of it, not a replacement.
 - **You do not fix.** You have no `edit` tool and you write only your verdict artifact.
-- **You do not change reviewer fresh-context isolation.** Reviewers stay single-angle, read-only, and fresh-context by design.
+- **You do not change reviewer fresh-context isolation.** Reviewers stay single-angle, read-only over the repository, and fresh-context by design.
```
mfittko, line 72: **low** (`contradiction-lens`): The diff edits this line to qualify "read-only" but keeps the stale claim that reviewers "stay single-angle", which contradicts agents/review.agent.md:32 ("a declared GROUP of angles (grouped dispatch, the shipped default - see resolveFanoutGroups)") and this very round's grouped 3-angle dispatch. review.agent.md is the correct side; judge.agent.md:72 (mirror .claude/agents/judge.md:69) is wrong. — judge: act

packages/core/src/loop/gate-fanin.mjs
```diff
@@ -1031,6 +1034,13 @@ export function applyJudgeDispositions(findings, judgeVerdict) {
       throw new Error(`judge disposition index ${d.index} is out of range (findings has ${enriched.length} entries)`);
     }
     const target = enriched[d.index];
+    // Reset judge-owned fields before the re-merge: a pre-enriched finding
+    // (already-enriched from a prior round, re-disposed by THIS verdict)
+    // must not let stale judgeCriterion/followUpDraft survive a
+    // defer -> act/reject re-disposition — the merged copy carries only
+    // what the current disposition provides, never prior-round residue.
+    delete target.judgeCriterion;
```
mfittko, line 1042: **nit** (`yagni`): The judge-owned-field reset guards a re-disposition of an already-enriched ledger, but no in-repo caller feeds enriched findings back through the seam: judge-pass takes the round's fan-in ledger (contract Phase 3.5) and carry-forward only carries clean-verdict angle provenance, never findings. No AC states the behavior (only AC7's test list references it), so it is defensive hardening for a hypothetical caller. — judge: reject

mfittko (commented):


**Reviewed head SHA:** `0497776fe9de6ede7027676789c23d6871a9ab3d`
**Verdict:** clean

**Findings summary:** 7 angles reviewed; 6 findings (see per-angle breakdown below).

- `docs` → `findings_present` (1 finding)

> **low** (`docs`): The --findings USAGE entry still lists only "severity, disposition, angle, summary, and optional positive-integer line", but this PR makes followUpDraft a hard-rejected field on that same flag. followUpDraft is documented nowhere as an accepted --findings input (only as a judge-verdict field in skills/docs/gate-review-sub-loop-contract.md:1115 and agents/judge.agent.md:53), so a caller who trips the new --findings&#91;i].followUpDraft must have a non-empty title and a body string error gets a parseError-attached USAGE that never names the field or its shape. — judge: reject
> Location: `scripts/github/write-gate-findings-log.mjs:24`


> **low** (`dry`): The PR replaced the lenient truthy-object followUpDraft check in validateFindingsArray (write-gate-findings-log.mjs:162-183) with a fail-closed shape check, but toFindingsLogShape still carries the ORIGINAL lenient copy of the same rule (f.followUpDraft && typeof f.followUpDraft === "object" && !Array.isArray(...)), which silently drops a malformed or non-object draft. That is the exact behavior the new validator's own comment says must never happen ("never passes through, or loses without a diagnostic, a malformed draft"), and toFindingsLogShape sits UPSTREAM of the new guard on the sanctioned pipeline path (scripts/loop/consolidate-fanin.mjs:1266 shapes the findings that feed write-gate-findings-log --findings). Net effect: the followUpDraft shape rule now exists in three places with two behaviors (validateJudgeVerdict:985-991 throws on defer, validateFindingsArray:171-177 throws on presence, toFindingsLogShape:1114 drops silently), so the new fail-closed guard fires only on drafts that survive the upstream silent drop. — judge: defer
> Location: `packages/core/src/loop/gate-fanin.mjs:1114`


> **low** (`pr-checklist-matrix`): Matrix row AC1 cites "assets:check green (mirror regenerated)" in the "Verified by DoD item(s)" column, but assets:check is not one of the four enumerated DoD items and is not part of npm test or npm run verify (package.json: verify runs test:assets, a different suite; assets:check is a standalone script). Consequently DoD 4 (root npm test aggregate) is cited by no matrix row, and AC1 has no DoD-item citation at all. The underlying coverage does exist (test/contracts/claude-assets-reproducible.test.mjs asserts the committed .claude tree is byte-reproducible, and it runs inside test:assets => npm test), so this is a traceability label gap, not a real verification hole. — judge: act


> **low** (`pr-checklist-matrix`): Non-goal mapping mismatch in the matrix: AC6's "Non-goal boundary" cell reads "no gate-pipeline reopen", which is a non-goal of linked issue #1656 but is NOT one of the four non-goals listed in this PR's own Non-goals section. Meanwhile the PR's non-goal 4 ("Re-doing the runJudgePass-layer coverage check; this relocates it into the pure seam and fixes its message") is represented in no matrix row, even though AC6 is exactly the criterion it bounds. — judge: act

**Next action:** Matrix body acts applied; single-angle wording correction folded into the imminent judge-policy PR (recorded on issue 1795); tick verified PR boxes, resolve threads, merge.

agents/judge.agent.md
mfittko: Acted with an orchestrator adjustment (judge: act): the judge asked for a merge-boundary commit for this one descriptive sentence; re-gating a twelve-file PR over prose is disproportionate, so the wording correction (grouped dispatch, not single-angle) is folded into the imminent judge-policy contract PR and recorded on issue 1795's third append. The docs-only lows in this round (the --findings USAGE shape gap) were rejected under the operator-adopted disposition policy.

packages/core/src/loop/gate-fanin.mjs
mfittko: Rejected (judge: reject, operator-adopted policy: a nit never produces an issue): the reset also protects future callers of the pure seam; removing it would reopen the stale-state hazard the prior round fixed.

Issue #1920, opened 2026-09-03
Follow up docs-reference validator latent false positives from disposition audit


mfittko:
## Objective

Harden the docs-reference contract test after the disposition audit for issue #1882.

## In scope

Replace the brittle scan-size proxy with explicit required-root assertions, and decide/pin support for setext headings and non-Markdown anchor targets so valid references are not false-failed.

## Acceptance criteria

- [x] Required root surfaces (`README.md`, `PLAN.md`, `AGENTS.md`, and at least one `docs/` file) are asserted directly.
- [x] Setext-heading behavior is explicit and covered by a focused test, or the contract documents that setext headings are unsupported.
- [x] Non-Markdown anchor behavior is explicit and covered by a focused test, without broadening filesystem reads.
- [x] `npm run verify` passes.

## Definition of done

- [x] Tests cover each accepted/rejected anchor behavior and remove the brittle size threshold.
- [x] Relevant docs explain the supported link surface.
- [x] `npm run verify` passes with no failures.

## Non-goals

- No redesign of the docs-reference validator.
- No external URL validation.
- No unrelated docs cleanup or review-disposition changes.

PR #1876, opened 2026-08-29
feat(loop): commit-msg guard hook enforcing the commit-message contract

mfittko:
Closes #1869

## Objective

Enforce the commit-message contract at commit time via a `commit-msg` git hook installed through the same worktree-provision path as the secret-scan/default-branch guard. Closes the commit-seam analogue of the PR-description hole: attribution trailers, no-bare-`#N`, and conventional subject form were rule-but-unenforced, so a non-compliant commit landed silently (the retro found a real rc.7 merge commit missing both trailers).

## In scope

A `commit-msg` hook installed through the existing worktree-provision path that fail-closed validates the attribution trailers, the conventional subject form (full conventional type set: `feat|fix|chore|docs|test|refactor|revert|perf|style|ci|build`), and bare-`#N` usage, with a per-commit waiver and git/tooling-authored exemptions. Out of scope: rewriting history, body-prose quality, and new dependencies.

## What

- `packages/core/src/loop/commit-msg-guard.mjs` (new): `renderCommitMsgGuardHook()` renders a self-contained ESM `commit-msg` hook (marker `dev-loops:commit-msg-guard`); `installCommitMsgGuard()` mirrors the default-branch-guard's atomic-write + foreign-hook-preservation pattern.
- `scripts/loop/ensure-worktree.mjs`: shares `resolveHooksInstallTarget` and installs the commit-msg guard alongside the branch guard at every `ensureWorktree()` return site (reported as `commitMsgGuard`).
- The hook validates fail-closed: requires the `Co-Authored-By: Claude …` + `Claude-Session:` trailers on an agent-authored commit (`CLAUDECODE=1`, the existing `isClaudeHarness` signal); rejects a bare non-issue `#N` (allows `Closes/Fixes/Refs #N`); requires conventional subject form; merge commits exempt; a per-commit waiver marker bypasses.
- Docs: `skills/docs/worktree-guidance.md` (+ mirror) rule `WORKTREE-COMMIT-MSG-GUARD`; registered in `required-rules.json`.

## Acceptance criteria

- [x] A commit missing the required attribution trailers is rejected by the `commit-msg` hook, pinned by a test.
- [x] A commit with a bare non-issue `#N` is rejected; a genuine `Closes/Fixes/Refs #N` is allowed. Pinned.
- [x] A commit subject not in `type(scope): summary` form is rejected, pinned.
- [x] The hook installs through the existing worktree/guard provision path, in every worktree.
[...]

PR #1875, opened 2026-08-29
test(docs): deterministic docs-reference validator (npm scripts, CLI subcommands, anchors)

mfittko:
Closes #1865

## Objective

Add a deterministic docs-reference validator so renamed/removed `npm run` scripts, CLI subcommands, and dangling markdown anchors in shipped docs fail closed instead of relying on soft agent judgment. Same defect shape as the PR-description hole: a mechanically-checkable requirement that was left un-enforced.

## In scope

A new contract test that resolves `npm run <name>` tokens against `package.json`, CLI subcommand citations against the route tables, and markdown anchors against real headings — across the `docs/`/`README`/`PLAN`/`AGENTS` surfaces, with an allowlist for historical/spec surfaces — plus the one live stale-reference fix. Out of scope: the soft semantic-drift docs angle, external-URL validation, and prose/style linting.

## What

- `test/contracts/referenced-docs-commands-shipped.test.mjs` (new): resolves `npm run <name>` tokens against `package.json` scripts, `dev-loops <category> <subcommand>` / `node cli/index.mjs …` citations against `cli/index.mjs`'s route tables, and markdown intra-/cross-file anchors (`(#section)`, `(other.md#section)`) against real headings (github-slugger-accurate). Surfaces: `README`/`PLAN`/`AGENTS`, `docs/` (recursive), `skills/`, `agents/`, `.claude/`. Runs under `npm run verify` via the existing `test/contracts/*.test.mjs` glob.
- `cli/index.mjs`: exported the canonical `SUBCOMMAND_ROUTES` + `SUBCOMMAND_ALIASES` so the check reuses the single source of truth (no re-declared route list).
- `scripts/docs/validate-links.mjs`: factored a shared fence-aware link iterator; added exported `extractAnchorLinks` + `collectMarkdownFiles`; existing link behavior unchanged.
- `docs/repo-wiki-manual-first.md`: live-casualty fix (line 144 cited `npm run build`, which resolves against the *cloned* repo-wiki checkout, not this repo — reworded so it no longer reads as a dangling same-repo script reference).

## Acceptance criteria

- [x] A doc citing an `npm run <name>` absent from `package.json` scripts FAILS the check, pinned by a test.
- [x] A doc citing a CLI subcommand that no longer resolves FAILS (with `--help`/alias passing correctly), pinned by a test.
- [x] A markdown anchor link resolving to no heading FAILS, pinned by a test.
[...]

PR #1872, opened 2026-08-29
feat(gate): enforce PR-description contract on tracker-backed PRs, fail-closed at draft-exit

mfittko:
Closes #1863

## Objective

Enforce the PR-description contract (Acceptance criteria + Definition of done as real checkboxes, explicit Non-goals, and `Closes #N`) on tracker-backed PRs, fail-closed at the draft-exit boundary. Closes the enforcement hole that let 9 rc.7 PRs ship stripped-down bodies through a green gate: the deterministic validator existed but only ran on the lightweight/issue-less path; tracker-backed PRs got only a soft `pr-description` review angle.

## In scope

Wire the deterministic `validateTrackerBackedPrBodySpec` into the draft-exit gate (`pre-pr-ready-gate.mjs` + `ready-for-review.mjs`) so a tracker-backed PR body missing the required invariants cannot leave draft. Reuse the existing validator; add tests; document the enforcement. Out of scope: the soft `pr-description` review angle, the lightweight/issue-less path, and re-merging rc.7 code.

## What

- `packages/core/src/loop/issue-refinement-artifact.mjs`: new `validateTrackerBackedPrBodySpec({ body, closingIssues })` — a thin reuse of the existing `validatePrBodySpec` (with `requireOpenQuestions:false`, since the tracker-backed contract in `copilot-loop-operations.md` never names an Open-questions section, and single-issue `expectedIssue` selection so umbrella PRs aren't forced to name one). No second divergent checker.
- `scripts/loop/pre-pr-ready-gate.mjs` + `scripts/github/ready-for-review.mjs`: both now fetch the PR body and, for any draft PR closing ≥1 issue, run `validateTrackerBackedPrBodySpec` and fail closed (no waiver surface) BEFORE `gh pr ready` runs. This is a hard code-path block at the same draft-exit layer the `gates.size` work uses — it cannot be soft-judged away. Issue-less PRs keep the lightweight path unchanged.
- `skills/docs/copilot-loop-operations.md` (+ regenerated `.claude/` mirror): documents the mechanical enforcement.

## Acceptance criteria

- [x] A tracker-backed PR whose body lacks AC checkboxes, DoD checkboxes, Non-goals, or `Closes #N` FAILS closed at draft-exit, pinned by a test (`pr-body-spec-tracker-backed.test.mjs` pins each missing invariant individually; the gate/wrapper integration tests pin the fail-closed path).
- [x] The check reuses `validatePrBodySpec` (extended), not a second divergent implementation.
[...]

PR #1921, opened 2026-09-03
fix(gate): require merit rationale for deferred findings

### LINKED ISSUE #1882: Audit resolved review-comment dispositions for soundness: low/medium findings examined on merits, not reflexively deferred
## Objective

Ensure every resolved gate/review finding was dismissed on its MERITS, not reflexively deferred because its severity made it a "closure candidate." Operator principle (2026-08-30): dismissing a low or medium finding without even looking at it is not the right call — candidacy for closure is not itself a reason to close.

## In scope

A retroactive audit of resolved review-comment dispositions plus a going-forward guardrail on the disposition step. Out of scope: reversing the avoid-head-bump/net-reduction churn policy, or forcing a fix for every low/medium.

## Context

The gate disposition policy (net-reduction; nit-disposition-avoids-head-bump) resolves nits/lows — and in-window mediums — with rationale to avoid a head-bump + full re-gate cascade on a converged PR. That is sound for avoiding churn. The risk: a finding can get closed BECAUSE it is a deferral candidate (its severity/timing), without a genuine per-finding examination of whether it should actually be acted on. A low/medium finding must be examined on its merits before closure; "nit -> resolve" / "low -> defer" applied by reflex is the failure mode. (This is the disposition-side analogue of the enforcement gaps this milestone closes, and relates to the fresh-context retro #1870.)

## Fix

1. **Retroactive audit (one-time, fresh-context):** re-examine every resolved resolve-with-rationale / deferral on the recent enforcement PRs (#1872, #1875, #1876) and a sampled set of prior merged PRs, on the merits: was the finding actually looked at, and is the rationale sound, or would acting on it (fix rather than defer) have been the better call? List any disposition that should be revisited; file a follow-up fix issue where warranted.
2. **Going-forward guardrail:** the disposition step must record a per-finding, merit-based rationale — not a severity-driven auto-close. Every low/medium deferral carries an explicit "examined on merits: <why deferring is correct>" note; a bare severity/candidacy-only dismissal without engagement is defined non-conforming. Where mechanically feasible, pin it (e.g. the disposition marker must carry a non-empty merit rationale distinct from the severity label).

## Acceptance criteria

- [x] A retroactive audit re-examines every resolved low/medium (and question) disposition on #1872 / #1875 / #1876 plus a sampled set of prior merged PRs on the merits; findings that should have been acted on are listed, with follow-up fix issues filed where warranted.
- [x] The disposition contract requires a per-finding merit-based rationale; a severity/candidacy-only dismissal ("nit -> resolve" / "low -> defer") with no examined-on-merits engagement is defined as non-conforming.
- [x] Low/medium deferrals carry an explicit examined-on-merits rationale, pinned by a check or contract test where mechanically feasible.
- [x] The relationship to the net-reduction / nit-disposition churn policy is reconciled in the contract docs: churn-avoidance stays; dismissing-without-looking is disallowed.

## Definition of done

- [x] Audit run and results recorded; any warranted follow-up fixes filed.
- [x] Disposition contract docs updated; `npm run verify` + `assets:check` green if code/test is added.

## Non-goals

- Not reversing the avoid-head-bump / net-reduction policy — churn avoidance stays; the change is that closure requires genuine examination, not that every finding must be fixed.
- Not forcing a fix for every low/medium: a sound, examined rationale to defer remains valid.

mfittko:
Closes #1882

## Summary

Audit resolved low/medium/question review dispositions on the recent enforcement PRs and a prior merged sample, then enforce per-finding merit rationale at the gate disposition step.

## Scope and context

The existing net-reduction policy remains: sound low/medium findings may be deferred and nits may resolve without a fixer cycle. This change prevents severity or round eligibility from being used as the reason itself. `close-gate-findings.mjs` now extracts the rendered finding summary, fails closed when it cannot, and includes an `Examined on merits:` rationale naming the finding and the applicable scope, acceptance-criteria, fix-window, or filing-bar basis.

## Acceptance criteria

- [x] Audit results cover every resolved low/medium/question disposition on PRs #1872, #1875, and #1876 plus sampled prior merged PRs, with concrete paths and severities.
- [x] Warranted live follow-up was filed as issue #1920; no sampled merged PR requires reopening.
- [x] The disposition contract defines severity-only closure as non-conforming while preserving churn avoidance.
- [x] Low/medium resolve-without-fix replies carry an explicit `Examined on merits:` rationale and fail closed when finding detail is unavailable.
- [x] Focused tests pin the positive and fail-closed rationale behavior.

## Definition of done

- [x] `npm run verify` passes.
- [x] `npm run assets:check` passes.
- [x] Audit is recorded on issue #1882 and contract docs are updated, including the generated Claude mirror.

## Non-goals

- No reversal of net-reduction or avoid-head-bump policy.
- No requirement to fix every low/medium finding.
- No reopening of merged PRs whose substantive dispositions were sound or already fixed/tracked.
- No broad redesign of gate-finding markers or reviewer severity classification.

## Risks and open questions

The rationale is generated from the posted finding summary plus the deterministic disposition basis; a human judge/reviewer still owns substantive merits. Follow-up issue #1920 owns latent docs-validator false-positive cleanup.


scripts/github/close-gate-findings.mjs
```diff
@@ -381,7 +411,7 @@ async function runDispositionPass({ repo, pr, round, threads, snapshot, login, m
     // null/mismatched.
     for (const target of fileableTargets) {
```
mfittko, line 411: **high** (`correctness`): In runDispositionPass's fileable-target loop, stampDeferredDisposition (the PATCH that stamps disposition=deferred issue=N onto the thread's marker, line 413) runs BEFORE dispositionMessage/buildMeritRationale (line 414), and buildMeritRationale now throws (via extractFindingSummary, line 201-210) whenever the thread's second line does not match the strict ^**severity** (angle): summary$ shape. parseFindingMarker (used inside stampDeferredDisposition to validate/parse the marker) and extractFindingSummary's regex (used to build the merit rationale) are two independently-validated structures with no shared invariant: a thread can carry a perfectly parseable marker (so the PATCH succeeds and commits) while its human-readable second line is malformed, edited, truncated, or otherwise off-shape, so extractFindingSummary throws right after the PATCH has already landed. The result is a thread stamped disposition=deferred issue=N that is never replied-to or resolved, and the exception aborts the whole for (const target of fileableTargets) loop, so any remaining fileable targets in the batch are also left unprocessed for this round and the entire unfiledTargets loop (nit/non-visible-low) never runs either — one malformed comment body blocks disposition for every other, unrelated, well-formed thread in the same round. Because the thread's own second line never changes on retry (only the marker's line-1 stamp does, and that thread short-circuits to return early once already stamped in stampDeferredDisposition, skipping the PATCH but still hitting the same throw in dispositionMessage), every subsequent close-gate-findings run reproduces the identical throw forever: the round can never report ok: true, unresolvedGateThreadCount is never computed, and the gate-close assertion can never observe the round as complete — a single malformed thread permanently deadlocks that PR's gate pipeline until a human manually edits or resolves the offending comment out-of-band. A … — judge: act

mfittko, line 411: **medium** (`coverage`): In runDispositionPass's fileable-target loop, stampDeferredDisposition (which PATCHes the GitHub comment to disposition=deferred) runs BEFORE dispositionMessage/buildMeritRationale, which now throws when the thread body has no parseable finding summary (line 187). If a fileable target's body fails extractFindingSummary, the PATCH has already landed but the reply+resolve never happens, and the whole pass aborts. On retry, stampDeferredDisposition's own idempotency guard (lines 311-320) treats the already-stamped, issue-number-matching marker as a no-op and returns normally, so dispositionMessage is invoked again on the SAME malformed body and throws again — this thread can never complete disposition, and no test exercises this stamp-then-throw interaction with the pre-existing retry-idempotency contract (test/github/close-gate-findings.test.mjs documents and tests idempotent retries elsewhere, e.g. issue 1807 comments, but not combined with the new fail-closed extraction). — judge: reject

mfittko, line 411: **medium** (`determinism`): In the fileable-deferral loop, stampDeferredDisposition (a PATCH that mutates the thread's marker to disposition=deferred issue=N) runs BEFORE dispositionMessage builds the reply text, and dispositionMessage now calls buildMeritRationale -> extractFindingSummary, which THROWS when the finding line does not match its regex. If that throw ever fires (e.g. a legacy-format or otherwise malformed self-authored finding body), the marker is already stamped disposition=deferred but the reply+resolve never runs, and the uncaught exception aborts the whole for (const target of fileableTargets) loop, so every remaining target queued behind it that round is silently never processed either. On the next run, selectDispositionTargets re-selects the same unresolved, already-stamped thread; stampDeferredDisposition now short-circuits (no re-PATCH) and dispositionMessage throws the identical error again, forever — the thread is stuck open with a stamped-but-unresolved marker, and it permanently blocks every other fileable target queued after it in the batch on every subsequent retry, with no automatic recovery path (a human must edit/repair the GitHub comment body). Before this PR, message-building was pure string interpolation and could not throw, so this stamp-then-message ordering was safe; adding a throwing step after the mutation is a new determinism/idempotency regression in the retry path. — judge: reject

scripts/github/close-gate-findings.mjs
```diff
@@ -381,7 +411,7 @@ async function runDispositionPass({ repo, pr, round, threads, snapshot, login, m
...
       await stampDeferredDisposition({ repo, commentId: target.commentId, round, mediumFixWindow, issueNumber }, { env, ghCommand });
```
mfittko: **low** (`gate-evidence`): For fileable targets, stampDeferredDisposition (GitHub-side mutation: marks the thread marker disposition=deferred with a real issue number) runs on line 413 BEFORE dispositionMessage on line 414, which now calls buildMeritRationale -> extractFindingSummary and can throw when the thread body does not match the expected **severity** (angle): summary shape. If that throw happens, the marker is already stamped and the follow-up issue may already reference the target, but no reply is posted and the thread stays unresolved. Every retry repeats the same order and re-throws identically (the underlying body format does not change), so the target is stuck in a stamped-but-unexplained state requiring manual intervention rather than a clean, side-effect-free refusal. Under the CURRENT single renderer (renderFindingLine in scripts/github/_gate-finding-surface.mjs) that feeds review-thread bodies this is unreachable, but the guard exists specifically to fail closed on unexpected input, so its own failure path should not leave partial GitHub-side state. — judge: reject

scripts/github/close-gate-findings.mjs
```diff
@@ -193,8 +194,36 @@ function unfiledResolutionReason(severity) {
   return "this low finding carries no operator-visibility signal, so it is resolved with rationale at gate close instead of filed to a tracked follow-up issue (net-reduction disposition policy)";
 }
 
-function unfiledResolutionMessage({ fp, severity, angle, round }) {
-  return `Resolved at gate close (round ${round}, fingerprint ${fp}, severity ${severity}, angle ${angle}): ${unfiledResolutionReason(severity)}.`;
+// A severity/round rule only selects an eligible disposition boundary; it is
+// not a per-finding reason to close. Require the rendered finding summary so
+// every resolve-without-fix reply records what was examined, rather than a
+// bare "low -> defer"/"nit -> resolve" label (#1882).
+function extractFindingSummary(body) {
```
mfittko, line 202: **high** (`input-validation`): extractFindingSummary's regex (^\*\*&#91;^*\n]+\*\*\s+\(&#91;^\n]+\):\s+(.+)$, line 203) is the sole input-validation gate for the untrusted GitHub comment body (thread.body, itself round-tripped through the GitHub API and editable by any collaborator/bot after posting) that buildMeritRationale requires. It validates the format independently of FINDING_MARKER_RE (which only covers line 1, the HTML marker comment) with no shared contract, so a thread whose marker is valid but whose second line was edited, truncated, or never matched this exact shape throws inside runDispositionPass's per-target loop (see the correctness finding at close-gate-findings.mjs:412 for the resulting stuck-thread/duplicate-append behavior). This is a legitimate fail-closed choice per the PR's stated intent, but the failure is unrecoverable from input alone (the malformed body never changes) and unbounded in blast radius (it aborts the whole batch, not just the one bad target) rather than being isolated/reported per-thread. — judge: act

scripts/github/close-gate-findings.mjs
```diff
@@ -193,8 +194,36 @@ function unfiledResolutionReason(severity) {
...
+export function buildMeritRationale({ body, severity, operatorVisible = false, round, mediumFixWindow }) {
+  const summary = sanitizeInline(extractFindingSummary(body));
+  const reason = severity === "medium"
+    ? `it remained open past the round-${mediumFixWindow} fix window and no in-scope fix was selected`
+    : severity === "low" && operatorVisible
+      ? "fixer triage found no same-commit fix warranted, while tracking preserves the operator-visible concern"
+      : severity === "low"
+        ? "fixer triage found no same-commit fix warranted and the finding does not clear the operator-visibility filing bar"
+        : "it is cosmetic and does not warrant a fixer cycle or tracked follow-up";
+  return `Examined on merits: \"${summary}\" was reviewed against current scope and acceptance criteria; ${reason}.`;
```
mfittko, line 225: **nit** (`input-validation`): buildMeritRationale wraps the extracted summary in literal double quotes (\"${summary}\") but sanitizeInline (imported from post-gate-findings.mjs) never escapes ". A finding summary containing a literal double quote renders as nested unescaped quotes in the posted comment (e.g. Examined on merits: "the "foo" bug" was reviewed...), which is cosmetically confusing but not a markdown/HTML injection risk (no active markup construct is broken). — judge: act

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `330c8d4654e0abc0043e14f4c1bdf8eb5844db20`
**Verdict:** findings_present

**Findings summary:** 9 angles reviewed; 10 findings (see per-angle breakdown below).

- `gate-evidence` → `findings_present` (2 findings)
- `link-check` → `clean`
- `pr-description` → `clean`
- `scope` → `clean`

> **medium** (`contract-surface`): The new 'Examined on merits:' requirement is documented only in the fan-in Consolidation bullet (lines 1344-1347), not in the authoritative GATE-EXEC-THREAD-DISPOSITION rule block (1814-1880) that both the new ADR 0055's References section and close-gate-findings.mjs's own error text (line 207: 'GATE-EXEC-THREAD-DISPOSITION violation: selected finding has no parseable finding summary') cite as this behavior's home. The rule block already narrates the exact reply-content contract for nit/low/medium closures ('its resolving reply names the rationale in-thread and nothing more', line 1858-1859) without ever mentioning the mandatory 'Examined on merits:' phrase or per-finding-summary requirement introduced by this PR. A maintainer or reviewer who follows the cited anchor to find this rule's content will not find it there. — judge: act
> Location: `skills/docs/gate-review-sub-loop-contract.md:1814`


> **medium** (`coverage`): buildMeritRationale has 4 severity/visibility reason branches (medium; low+operatorVisible; low+not-visible; nit/default), but only the low+not-visible branch's actual generated text is asserted anywhere (test/github/close-gate-findings.test.mjs:475-483, a standalone unit call). Every integration test that exercises the medium disposition path (test/github/close-gate-findings.test.mjs:505-524, 612-655), the operator-visible-low path (test/github/close-gate-findings.test.mjs:558-579), and the nit path (test/github/close-gate-findings.test.mjs:584-605) only asserts unrelated pre-existing substrings via postReplyEntry/assertStdinIncludes, never the new 'Examined on merits:' text or the reason string for that branch. A regression that swaps/breaks the medium, low-visible, or nit reason strings (e.g. wrong mediumFixWindow interpolation, wrong branch selection) would ship silently. — judge: act
> Location: `scripts/github/close-gate-findings.mjs:192`


> **low** (`coverage`): extractFindingSummary strips a trailing judge disposition suffix via body.match(...).replace(/\s+— judge:.*$/u, ""), matching renderFindingLine's judgeSuffix format (' — judge: ...') in scripts/github/_gate-finding-surface.mjs:438-441. No test fixture in test/github/close-gate-findings.test.mjs includes a judge-suffix-bearing finding body reaching buildMeritRationale/extractFindingSummary (the unit test at line 476 uses a plain summary with no judge suffix). A future divergence in the em-dash/format used by the two sides (render vs strip) would silently leak the judge disposition text into the merit rationale, undetected. — judge: reject
> Location: `scripts/github/close-gate-findings.mjs:185`


> **low** (`gate-evidence`): PR DoD checkbox claims npm run assets:check passes, but the recorded validation record only captures npm run verify (run-p ... test:assets test:extension test:scripts test:core test:docs test:pack test:dev-loop test:workflows), which does not invoke assets:check (a separate script: node scripts/claude/generate-claude-assets.mjs --check). No suite entry evidences that specific DoD item at this head. Manual diff of .claude/skills/docs/gate-review-sub-loop-contract.md against skills/docs/gate-review-sub-loop-contract.md at HEAD shows the mirror is in sync, so the underlying claim appears true, but the recorded validation artifact does not substantiate it. — judge: reject
> Location: `tmp/gate-context/mfittko-dev-loops/pr-1921/draft_gate-330c8d4654e0abc0043e14f4c1bdf8eb5844db20.validation.json`

scripts/github/close-gate-findings.mjs
mfittko: Fixed in f76ba45d. buildMeritRationale now collapses embedded double quotes in the summary before wrapping it, so a summary containing a quote no longer renders as nested unescaped quotes. Pinned by a focused test in test/github/close-gate-findings.test.mjs.

scripts/github/close-gate-findings.mjs
mfittko: Fixed in f76ba45d (shared root cause). Building the merit rationale before stampDeferredDisposition means a parse failure now refuses cleanly with no partial GitHub-side state (no stamp, no follow-up-issue append) rather than leaving a stamped-but-unresolved marker. Covered by the isolation regression test.

scripts/github/close-gate-findings.mjs
mfittko: Fixed in f76ba45d. extractFindingSummary's throw is now isolated per target: a malformed thread body is recorded in dispositionFailures and skipped, so the fail-closed refusal no longer aborts the whole batch (bounded blast radius) and the thread stays unresolved to keep the gate blocked. Regression test added in test/github/close-gate-findings.test.mjs.

scripts/github/close-gate-findings.mjs
mfittko: Fixed in f76ba45d. The disposition pass now builds and validates each resolve-without-fix reply (buildMeritRationale) BEFORE any mutating GitHub call, and processes each target in its own try/catch. A target whose finding summary cannot be parsed is recorded in dispositionFailures and left unresolved, so it never lands a marker stamp with no reply (the permanent-retry deadlock) and never aborts disposition for the other well-formed threads in the batch. An unbuildable target is also partitioned off before the follow-up-issue append, so it is never recorded on an issue it can never be replied against. Regression test added in test/github/close-gate-findings.test.mjs.

mfittko: Fixed in f76ba45d (shared root cause). The disposition pass now builds/validates each reply before any mutation and isolates each target in its own try/catch, so a malformed body can no longer leave a stamped-but-unresolved thread or abort the rest of the batch. See test/github/close-gate-findings.test.mjs for the isolation regression test.

mfittko: Fixed in f76ba45d (shared root cause). The disposition pass now builds/validates each reply before any mutation and isolates each target in its own try/catch, so a malformed body can no longer leave a stamped-but-unresolved thread or abort the rest of the batch. Pinned by test/github/close-gate-findings.test.mjs.

scripts/github/close-gate-findings.mjs
```diff
@@ -348,58 +382,108 @@ async function runDispositionPass({ repo, pr, round, threads, snapshot, login, m
...
+        } catch (err) {
+          recordFailure(target, err);
+        }
+      }
     }
   }
   // #1846: a nit or a non-operator-visible low is resolved-with-rationale
   // in-thread — no GET/PATCH round-trip (there is nothing to stamp: the
   // marker's disposition field stays absent), no follow-up issue.
   for (const target of unfiledTargets) {
```
mfittko, line 469: **low** (`coverage`): The malformed-summary fail-closed isolation (build-before-mutate, recordFailure, no partial state) is only pinned for the FILEABLE loop (lines 406-455, test 'a malformed fileable target is isolated...' at test/github/close-gate-findings.test.mjs:623). The unfiled loop (nit / non-operator-visible-low, lines 469-480) has its own separate try/catch around unfiledResolutionMessage -> buildMeritRationale and is a distinct, independently-regressable code path (no shared stamp/follow-up-issue plumbing with the fileable branch), but no test exercises a malformed body through it. A regression that drops the try/catch, mis-wires recordFailure, or breaks unfiledResolved counting in this specific loop would not be caught by any existing test — every nit/invisible-low fixture in the suite (nitBody, invisibleLowBody, etc.) already carries a parseable summary line. — judge: reject

scripts/github/close-gate-findings.mjs
```diff
@@ -193,8 +195,39 @@ function unfiledResolutionReason(severity) {
...
-  return `Resolved at gate close (round ${round}, fingerprint ${fp}, severity ${severity}, angle ${angle}): ${unfiledResolutionReason(severity)}.`;
+// A severity/round rule only selects an eligible disposition boundary; it is
+// not a per-finding reason to close. Require the rendered finding summary so
+// every resolve-without-fix reply records what was examined, rather than a
+// bare "low -> defer"/"nit -> resolve" label (#1882).
+function extractFindingSummary(body) {
+  const match = typeof body === "string"
+    ? body.match(/^\*\*[^*\n]+\*\*\s+\(`[^`\n]+`\):\s+(.+)$/mu)
+    : null;
+  const summary = match?.[1]?.replace(/\s+— judge:.*$/u, "").trim();
```
mfittko, line 206: **low** (`coverage`): extractFindingSummary strips a trailing judge-disposition annotation via .replace(/\s+— judge:.*$/u, ""). This suffix is a real, reachable production shape: renderFindingLine (scripts/github/_gate-finding-surface.mjs:436-440) and equivalent renderers in post-gate-findings.mjs/upsert-checkpoint-verdict.mjs append — judge: &lt;disposition> to the same rendered finding line whenever a finding carries a non-empty judgeDisposition (any severity, including low/medium, per write-gate-findings-log.mjs's JUDGE_DISPOSITIONS). No test in this PR (or elsewhere) constructs a thread body with that suffix present and asserts buildMeritRationale/extractFindingSummary strips it correctly rather than embedding the raw '— judge: ...' text inside the quoted 'Examined on merits' rationale. — judge: reject

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `f76ba45d248d0c2bc1a0adc02e752ce3819a9f63`
**Verdict:** clean

**Findings summary:** 9 angles reviewed; 2 findings (see per-angle breakdown below).

- `correctness` → `clean`
- `determinism` → `clean`
- `gate-evidence` → `clean`
- `link-check` → `clean`
- `pr-description` → `clean`
- `scope` → `clean`

**Next action:** x

scripts/github/close-gate-findings.mjs
mfittko: Resolved at gate close (round 8, fingerprint f6bf86aa4aa9c31b, severity low, angle coverage): this low finding carries no operator-visibility signal, so it is resolved with rationale at gate close instead of filed to a tracked follow-up issue (net-reduction disposition policy).

scripts/github/close-gate-findings.mjs
mfittko: Resolved at gate close (round 8, fingerprint 540dc7ff9ef0e789, severity low, angle coverage): this low finding carries no operator-visibility signal, so it is resolved with rationale at gate close instead of filed to a tracked follow-up issue (net-reduction disposition policy).

mfittko (commented):


**Reviewed head SHA:** `f76ba45d248d0c2bc1a0adc02e752ce3819a9f63`
**Verdict:** clean

**Findings summary:** 7 angles reviewed; 1 finding (see per-angle breakdown below).

- `docs` → `clean`
- `dry` → `clean`
- `yagni` → `clean`

> **low** (`pr-checklist-matrix`): PR DoD box '&#91;x] npm run assets:check passes' has no corresponding suite in this round's recorded validation evidence — jq '&#91;.suites&#91;].name]' on the validation record returns only &#91;"verify"], and assets:check (package.json: node scripts/claude/generate-claude-assets.mjs --check) is a separate npm script not invoked by verify or by any CI workflow. Manually diffing .claude/skills/docs/gate-review-sub-loop-contract.md against skills/docs/gate-review-sub-loop-contract.md at this head shows the mirror hunks are identical to the source hunks, so the checkbox is very likely truthful, but it is not backed by machine evidence in the gate's own validation capture. — judge: reject
> Location: `tmp/gate-context/mfittko-dev-loops/pr-1921/pre_approval_gate-f76ba45d248d0c2bc1a0adc02e752ce3819a9f63.validation.json`

**Next action:** Pre-approval gate clean; resolve the 1 judge-rejected low, then merge-ready checks

PR #2349, opened 2026-09-21
fix(gate-review): apply the inline severity floor to review-gate verdicts

mfittko:
## Summary

The `review` gate's hardcoded `activeGateConfig` (upsert-checkpoint-verdict.mjs) carried only `blockCleanOnFindingSeverities`, so `activeGateConfig.inlineSeverityFloor` was `undefined` and the fold guard was a no-op: every locatable finding posted inline on a review-gate round. That contradicted the documented `medium` default of the GATE-COMMENT-INLINE-SEVERITY-FLOOR contract (Refs #2263) — the draft and pre-approval gates fold low/nit findings into the verdict body's collapsed `<details>` block, while review-gate rounds posted them all as inline comments.

## Objective

Make `--gate review` verdicts honor the same inline severity floor as the other gates, so a review round folds sub-floor findings into the verdict body instead of posting one inline comment per finding.

## In scope

- The review-gate branch of `activeGateConfig` in `scripts/github/upsert-checkpoint-verdict.mjs`.
- One regression test pinning the review-only branch (locatable low/nit fold, zero inline comments).
- One changelog fragment under `changes/`.

## Explicit non-goals

- No change to the `review` gate's VERDICT semantics (its verdict stays informational). Folding does remove review-round low/nit findings from `unresolvedGateThreadCount` — a folded finding creates no gate-authored thread — which is the documented `GATE-COMMENT-INLINE-SEVERITY-FLOOR` fold property, a comment-surface effect rather than a blocking-severity change.
- No new `gates.review` config key, and no change to the draft/pre-approval floor resolution.
- No change to `close-gate-findings.mjs` (it already defaults to `medium`).

## Acceptance criteria

- [ ] A `--gate review` round with locatable low/nit findings posts ZERO inline comments and folds them into the verdict body's suppressed-findings block.
- [ ] Medium/high findings still post inline; a `question` never folds.
- [ ] Draft and pre-approval gate behavior is unchanged.

## Definition of done

- [ ] `inlineSeverityFloor: "medium"` is set on the review-gate `activeGateConfig`.
- [ ] A regression test fails if the floor is removed again.
- [ ] `bun run verify` passes.
- [ ] Changelog fragment added.

## Open questions / risks

- The review gate has no per-gate config key, so the documented `gates.<gate>.inlineSeverityFloor` lowering escape hatch is unreachable for review rounds; the floor is fixed at the documented `medium` default. Documenting that is follow-up territory, not this change.

## Change

[...]

Issue #1907, opened 2026-09-01
Deterministic dispatch discipline: bounded-parallel fan-out, retry-on-transient, blocking joins, end-of-run contract


mfittko:
## Summary

One session of dev-loop operation (2026-08-31) burned 8+ dispatch trees across 3 model tiers and surfaced 4 distinct failure classes in the chained-dispatch pattern. Every failure was recoverable because gate verdicts and findings artifacts are durable, but the recovery knowledge lives only in operator steering messages — exactly where durable rules must never live (#1871). This issue pins the validated dispatch discipline as contract prose + anti-patterns, deterministic enough to survive session loss.

Operator decision (2026-09-01): scope extended from dispatch-discipline prose to the mechanism change — this repo flips from the sequential fan-out workaround to bounded-parallel dispatch, absorbing the bounded-parallel work previously sketched as a #1785-adjacent follow-up.

## Failure classes observed (evidence from live runs)

1. **Ended async parents are terminal.** A dev-loop parent that ends its turn after dispatching a nested child is never woken by the child's completion — the chain silently dead-ends (observed: unit-1 reviewer completed, wrote its artifact, parent stayed terminal). The sanctioned patterns are blocking joins (parent stays alive) or contact_supervisor before ending a turn; "end turn and await wake" is structurally unsound.
2. **Provider 4xx kills trees.** A parallel dispatch of two children tripped `429 Concurrent requests` and the run died — no backoff, no retry, tree lost mid-gate. Later a `402 Insufficient Balance` killed another tree outright. Reviewer units write idempotent single-artifact findings, so retrying the SAME unit is always safe.
3. **Sticky provider fallback.** One intermittent 429 caused reviewers to be pinned to a fallback provider for the rest of the session even after the cap window passed — provider choice must be a per-dispatch decision.
4. **Post-merge drift.** After merging, a child re-entered consolidation machinery whose artifacts already existed, typo'd an absolute path with the error swallowed by `2>/dev/null`, wandered through wrong-directory `ls` probes, and burned its remaining budget without producing a final report.
[...]

Issue #1086, opened 2026-07-02
feat(docs): cross-harness regression contract — promote harness-agnostic non-regression to AGENTS.md + skills/docs (pre-0.7)


mfittko:
## Summary

dev-loops is harness-agnostic (Pi + Claude Code). The principle "any harness-specific change MUST NOT regress other harnesses" is currently captured ad-hoc inside #1084's "Non-regression constraint" section. It should be promoted to a repo-wide rule in AGENTS.md (the canonical main-agent instruction surface) with the detailed testing procedure in a `skills/docs/` contract doc, so it applies to ALL work — not just #1084 — and is enforced before the 0.7.0 release.

This surfaced during the 0.7 drive when Seam 2 (child fan-out) was identified as a pi-harness gap with a working Claude Code precedent: the fix is on pi's side, but it must not regress the Claude Code path that already works.

## Why AGENTS.md + skills/docs (not one or the other)

AGENTS.md already states: "Keep workflow procedure out of AGENTS; put shared contracts under `skills/docs/`. Brief, high-signal main-agent dispatch contracts … stay in AGENTS.md." So:
- **AGENTS.md**: the one-line hard rule (a universal constraint on all repo work, not workflow procedure).
- **`skills/docs/cross-harness-regression-contract.md`**: the detailed procedure (what counts as a harness-specific change, required Claude-Code-path test suites, the "additive on Pi / no-op-or-validated on Claude Code" bar, how to add coverage).

## Proposed change

1. Add a Working-rules bullet to the **repo** AGENTS.md (`/Users/mfittko/github/dev-loops/AGENTS.md`), not the global `~/.pi/agent/AGENTS.md`:
   > - dev-loops is harness-agnostic (Pi + Claude Code). Any harness-specific change MUST NOT regress other harnesses — add/keep cross-harness regression coverage. Canonical procedure: `skills/docs/cross-harness-regression-contract.md`.
2. Author `skills/docs/cross-harness-regression-contract.md` covering:
   - **What is a harness-specific change**: anything that touches a path conditional on the runtime harness (pi vs Claude Code) — runner-coordination, run-id injection, subagent/Task fan-out tool provisioning, env vars (`PI_SUBAGENT_RUN_ID`, `DEVLOOPS_RUN_ID`), session/intercom, TUI, or any `process.env` / capability detection that diverges per harness.
[...]

PR #2352, opened 2026-09-21
feat(gate-fanout): deterministic one-call parallel wave dispatch for Pi gate rounds

### LINKED ISSUE #2350: Pi gate fan-out has no deterministic parallel dispatch path — rounds silently serialize and time out
## What

Under Pi, the gate fan-out (`draft_gate` / `pre_approval_gate`) has no deterministic way to release a wave of reviewers concurrently. Every script that composes a round stops at per-unit prompt files (`emit-fanout-dispatch.mjs` → one `promptPath` per unit). Nothing emits the actual dispatch call, so the concurrency posture is left to the conductor agent to compose as prose — and in practice it does not hold.

## Why it matters

`gates.fanout.maxConcurrent: 3` is this repo's documented default posture. `GATE-EXEC-FANOUT-SEQUENTIAL-FALLBACK` states bounded parallelism is the DEFAULT and that `gates.fanout.sequential: true` is a load fallback which MUST be justified and recorded. Under Pi, a conductor reaches for N separate blocking `subagent` calls, and pi-subagents rejects the second onward:

```js
// pi-subagents/src/runs/foreground/subagent-executor.js:6965
const runsForeground = dispatchParams.clarify === true || (dispatchParams.async ?? asyncByDefault) !== true;
if (!runsForeground) return execute(...);                 // async:true bypasses the guard
if (deps.state.subagentInProgress === true) return duplicateSubagentCallResult(requestParams);
// → "Rejected: a subagent call is already in progress. Issue exactly ONE subagent call per turn."
```

That guard is foreground-only. Parallelism is available inside ONE call, and the only working shape in this pi-subagents version is a workflow script:

```js
// subagent({ workflowScriptPath: "<wave>.js", cwd: <worktree>, async: false, timeoutMs: 900000 })
return runs.all([
  { key: "draft-gate-holistic", agent: "review", context: "fresh", task: unitTask("holistic", "draft-gate-holistic") },
  { key: "draft-gate-coverage",  agent: "review", context: "fresh", task: unitTask("coverage", "draft-gate-coverage") }
]);
```

`tasks: [...]` is NOT an available shape in this version — the public API rejects it outright with `Legacy top-level chain and parallel inputs were removed; use workflowScript.` Only `workflowScript` (inline) or `workflowScriptPath` reach `runs.all`. That is the shape the contract already names — "Code-driven Pi `runs.all`" in the per-harness delivery table, with `GATE-EXEC-FANOUT-DISPATCH-KEY` requiring a unique non-empty `key` per item. But no code path emits it, and the SKILL's own phrasing ("up to `gates.fanout.maxConcurrent` dispatch units per wave via blocking joins", #1907) reads as N separate blocking calls, which is exactly the shape Pi rejects.

Observed cost: PR #2349's `draft_gate` round resolves to 6 dispatch units. Sequentially, at a 14-minute per-unit ceiling, that is ~84 minutes; two parallel waves is ~28 minutes. Two consecutive dev-loop runs died on their parent deadline (1800000ms, then 3600000ms) mid-fan-out with no verdict posted. The second run fell back to sequential joins after treating the guard rejection as "the harness allows only one subagent call per turn", which is a misdiagnosis — and the documented sequential fallback is supposed to be a justified, recorded load fallback, not a silent degrade.

## Verified working shape (observed on PR #2349)

A conductor that was steered onto the workflow-script shape completed a 6-unit wave concurrently:

| Time (UTC) | Angle artifacts written |
| --- | --- |
| 21:57:53 | determinism |
| 21:59:50 | correctness, input-validation |
| 22:00:57 | gate-evidence, pr-description, scope |

Three artifacts written in the same second, six within ~3 minutes — against ~72 minutes sequential at the per-unit ceiling. A follow-up two-unit wave (`holistic`, `coverage`) used the same `runs.all` shape with unique keys.

A second, independent finding from the same run: every reviewer prompt carried a bounded tool-call budget and a mandatory-write clause — "HARD BUDGET: finish within 25 tool calls and WRITE your artifact … never return without the file written." In the earlier run, reviewers hit their per-unit timeout mid-thought and wrote nothing, so the round produced no evidence at all. Bounded per-reviewer effort plus a mandatory artifact write is part of what makes a wave collectable.

## Evidence

- `pi-subagents/src/runs/foreground/subagent-executor.js:6965` — the foreground-only guard (`duplicateSubagentCallResult`).
- `tasks: [...]` rejection: `Legacy top-level chain and parallel inputs were removed; use workflowScript.`
- `emit-fanout-dispatch.mjs` emits per-unit `promptPath` files only. `emit-plan.json` carries `count: 6`, `maxConcurrent: 3` and per-unit `angles` — no dispatch call body.
- `rg 'runs\.all|workflowScript|tasks:' scripts packages .pi/skills` returns zero hits. The only references are prose in `skills/docs/gate-review-sub-loop-contract.md` and `test/contracts/gate-fanout-dispatch-key-contract.test.mjs`, which asserts the doc text contains `` `runs.all` `` and `` `key` `` — never that a runtime path issues a parallel call.
- Nothing detects the serialization. A sequential round still records per-angle sentinels and distinct reviewers, so the records-floor (#1868) and `requireFanoutProvenance` both pass. Only wall-clock regresses, which is invisible until a parent deadline fires.

## Acceptance criteria

- [ ] A deterministic seam emits the ready parallel dispatch body for a wave from the existing round artifacts (`emit-plan.json` units plus `promptPath` bytes), so the conductor never composes the call shape itself.
- [ ] The emitted body uses `workflowScript` / `workflowScriptPath` with `runs.all([...])` — never `tasks: [...]`, which this pi-subagents version rejects — and carries a unique non-empty `key` per unit (`GATE-EXEC-FANOUT-DISPATCH-KEY`).
- [ ] The emitted body releases up to `gates.fanout.maxConcurrent` units concurrently in ONE call.
- [ ] Emitted reviewer prompts carry a bounded tool-call budget plus a mandatory artifact-write clause, so a reviewer that runs long still lands its findings artifact.
- [ ] `gates.fanout.sequential: true` still resolves to a single-unit wave, and any other serialization fails closed instead of silently degrading a `gates.requireFanoutEvidence` gate.
- [ ] The Pi dispatch guidance in `skills/dev-loop/SKILL.md` (#1907) and the per-harness delivery table in `skills/docs/gate-review-sub-loop-contract.md` state the one-call workflow-script shape explicitly, so "blocking joins" can no longer be read as N separate calls.
- [ ] Regression coverage asserts the emitted wave is one `runs.all` call with unique keys, and fails when a unit lacks a key or when units are emitted as separate calls.

## Definition of done

- [ ] Contract test coverage for the emitted wave shape: unique keys, single `runs.all` call, concurrency bound honored.
- [ ] Cross-harness non-regression (#1086): the Claude Code / Codex agent-driven dispatch path is unchanged.
- [ ] `bun run verify` green.
- [ ] A live Pi gate round completes a multi-unit fan-out in parallel waves, preserving primer-then-release ordering (`GATE-EXEC-PRIME`) and consolidating normally at fan-in.

## Non-goals

- Changing pi-subagents' foreground guard, or upstreaming a `tasks: [...]` parallel input.
- Re-litigating the concurrency default, or the light-mode inline carve-out.
- Making Claude Code / Codex dispatch code-driven; their delivery stays agent-authored per the per-harness table.

## AC → DoD mapping

| Acceptance criterion | Required completion evidence |
| --- | --- |
| A deterministic seam emits the ready parallel dispatch body for a wave from the existing round artifacts (`emit-plan.json` units plus `promptPath` bytes), so the conductor never composes the call shape itself. | Contract test coverage for the emitted wave shape: unique keys, single `runs.all` call, concurrency bound honored. |
| The emitted body uses `workflowScript` / `workflowScriptPath` with `runs.all([...])` — never `tasks: [...]`, which this pi-subagents version rejects — and carries a unique non-empty `key` per unit (`GATE-EXEC-FANOUT-DISPATCH-KEY`). | Contract test coverage for the emitted wave shape: unique keys, single `runs.all` call, concurrency bound honored. |
| The emitted body releases up to `gates.fanout.maxConcurrent` units concurrently in ONE call. | Contract test coverage for the emitted wave shape: unique keys, single `runs.all` call, concurrency bound honored. |
| Emitted reviewer prompts carry a bounded tool-call budget plus a mandatory artifact-write clause, so a reviewer that runs long still lands its findings artifact. | A live Pi gate round completes a multi-unit fan-out in parallel waves, preserving primer-then-release ordering (`GATE-EXEC-PRIME`) and consolidating normally at fan-in. |
| `gates.fanout.sequential: true` still resolves to a single-unit wave, and any other serialization fails closed instead of silently degrading a `gates.requireFanoutEvidence` gate. | Contract test coverage for the emitted wave shape: unique keys, single `runs.all` call, concurrency bound honored. |
| The Pi dispatch guidance in `skills/dev-loop/SKILL.md` (#1907) and the per-harness delivery table in `skills/docs/gate-review-sub-loop-contract.md` state the one-call workflow-script shape explicitly, so "blocking joins" can no longer be read as N separate calls. | Cross-harness non-regression (#1086): the Claude Code / Codex agent-driven dispatch path is unchanged. |
| Regression coverage asserts the emitted wave is one `runs.all` call with unique keys, and fails when a unit lacks a key or when units are emitted as separate calls. | Contract test coverage for the emitted wave shape: unique keys, single `runs.all` call, concurrency bound honored. `bun run verify` green. |

## Scope boundary

This issue owns the deterministic Pi gate fan-out wave-dispatch seam (the emitter, its contract wording, and its regression tests). It does NOT own the pi-subagents foreground guard or any upstream `tasks: [...]` parallel input (#2342).

mfittko:
## Objective

Under Pi the gate fan-out had no deterministic way to release a wave of reviewers concurrently. `scripts/github/emit-fanout-dispatch.mjs` stopped at per-unit `promptPath` files, so the conductor agent composed the dispatch call itself — and reached for N separate blocking `subagent` calls, which pi-subagents rejects from the second onward (its `subagentInProgress` guard is foreground-only). The round silently serialized and died on the parent deadline: PR #2349's `draft_gate` resolves to 6 dispatch units, ~84 minutes sequentially against ~28 minutes in two parallel waves.

This PR adds the missing seam so the conductor never composes the call shape itself, and pins the one-call shape in the contracts so "blocking joins" can no longer be read as N separate calls.

## In scope

- `scripts/github/emit-wave-dispatch.mjs` (new): turns the round's emitted `emit-plan.json` units plus their `promptPath` bytes into ONE ready workflow script per wave — a single `runs.all([...])` call with a unique non-empty `key` per dispatch unit and each unit's composed prompt bytes inlined verbatim as its `task` — plus the ready `subagent({ workflowScriptPath, cwd, async: false, timeoutMs })` call body, persisted to the keyed `<gate>-<headSha>.wave-plan.json` sibling.
- `scripts/github/emit-fanout-dispatch.mjs`: mandatory artifact-write clause added to the emitted reviewer prompt suffix; usage text names the wave step.
- `skills/dev-loop/SKILL.md` and `skills/docs/gate-review-sub-loop-contract.md`: new rule `GATE-EXEC-FANOUT-WAVE-DISPATCH`, the one-call shape stated explicitly in the dispatch discipline and in the per-harness delivery table.
- `skills/docs/required-rules.json`: rule registration; regenerated `.claude/` mirror.
- `skills/copilot-pr-followup/SKILL.md`: Phase 2 step 4 wires the Pi wave seam (`emit-wave-dispatch.mjs`, ONE `subagent` call per emitted wave) into the operative runbook, alongside `GATE-EXEC-FANOUT-WAVE-DISPATCH`.
- `docs/decisions/0081-pi-gate-fanout-one-runs-all-wave.md` (new): the ADR record for the one-`runs.all`-call-per-wave decision.
- `test/github/emit-wave-dispatch.test.mjs` (new): wave-shape regression coverage.
- `test/contracts/gate-fanout-wave-dispatch-contract.test.mjs` (new) and `test/github/emit-fanout-dispatch.test.mjs`: contract and emitter regression coverage.
- `changes/2350-gate-fanout-wave-dispatch.md`: changelog fragment.

## Acceptance criteria

- [ ] A deterministic seam emits the ready parallel dispatch body for a wave from the existing round artifacts (`emit-plan.json` units plus `promptPath` bytes), so the conductor never composes the call shape itself.
- [ ] The emitted body uses `workflowScript` / `workflowScriptPath` with `runs.all([...])` — never `tasks: [...]`, which this pi-subagents version rejects — and carries a unique non-empty `key` per unit (`GATE-EXEC-FANOUT-DISPATCH-KEY`).
- [ ] The emitted body releases up to `gates.fanout.maxConcurrent` units concurrently in ONE call.
- [ ] Emitted reviewer prompts carry a bounded tool-call budget plus a mandatory artifact-write clause, so a reviewer that runs long still lands its findings artifact.
- [ ] `gates.fanout.sequential: true` still resolves to a single-unit wave, and any other serialization fails closed instead of silently degrading a `gates.requireFanoutEvidence` gate.
- [ ] The Pi dispatch guidance in `skills/dev-loop/SKILL.md` (#1907) and the per-harness delivery table in `skills/docs/gate-review-sub-loop-contract.md` state the one-call workflow-script shape explicitly.
- [ ] Regression coverage asserts the emitted wave is one `runs.all` call with unique keys, and fails when a unit lacks a key or when units are emitted as separate calls.

## Definition of done

- [ ] Contract test coverage for the emitted wave shape: unique keys, single `runs.all` call, concurrency bound honored, refusal on a unit without a key and on a separate-call partition.
- [ ] Cross-harness non-regression (#1086): the Claude Code / Codex agent-driven dispatch path is unchanged; the new suite is green under `CLAUDECODE=1`.
- [ ] `bun run verify` green.
- [ ] A live Pi gate round completes a multi-unit fan-out in parallel waves, preserving primer-then-release ordering (`GATE-EXEC-PRIME`) and consolidating normally at fan-in.

## Non-goals

- Detecting a conductor that bypasses the emitter entirely. The emitter refuses to emit a non-sanctioned serialization; a conductor that skips `emit-wave-dispatch.mjs` and issues N per-unit calls is still undetected at gate time, which remains the open gap this issue records as "nothing detects the serialization".
- Changing pi-subagents' foreground guard, or upstreaming a `tasks: [...]` parallel input.
- Re-litigating the concurrency default, or the light-mode inline carve-out.
- Making Claude Code / Codex dispatch code-driven; their delivery stays agent-authored per the per-harness table.

## Validation

- `bun run verify`: pass.
- `bun run test:doc-guard`: pass.
- `node scripts/loop/check-comment-discipline.mjs --base origin/main`: pass.
- `node scripts/loop/pre-flight-gate.mjs --expected-branch issue-2350 --check-subagents`: pass.
- `CLAUDECODE=1 bun test test/github/emit-wave-dispatch.test.mjs`: pass.

Closes #2350


scripts/github/emit-fanout-dispatch.mjs
```diff
@@ -80,7 +80,11 @@ Output (stdout, JSON):
   Wave the EMITTED units at most \`maxConcurrent\` at a time (1 when
   gates.fanout.sequential is set). Do NOT use the artifact's fanout.wavePlan to
   bound this step: that plan is computed over the UNSPLIT resolveFanoutGroups
-  units and no longer matches this step's split unit set.
+  units and no longer matches this step's split unit set. Run
+  \`scripts/github/emit-wave-dispatch.mjs\` at the same (repo, pr, gate, headSha)
```
mfittko, line 85: **medium** (`correctness`): The newly added wave-step guidance is unconditional ("Run scripts/github/emit-wave-dispatch.mjs at the same (repo, pr, gate, headSha) key next"), but the sibling's own ZERO-UNIT success path is reachable: with --pending and every resolved angle carried, emit-fanout-dispatch returns {ok:true, count:0, units:&#91;]} (emit-fanout-dispatch.mjs:577-595). The new wave emitter refuses exit 1 on exactly that plan — if (plan?.ok !== true || units.length === 0) at emit-wave-dispatch.mjs:452-453 with the message "carries no dispatch units ... re-run emit-fanout-dispatch.mjs". A conductor that follows the emitted guidance on a legitimately empty pending round therefore hits a fail-closed refusal that reads as a round failure and tells it to re-run the emitter (a loop), instead of simply dispatching nothing. — judge: act
Recommendation: Either gate the guidance on count > 0 ("when this step emitted one or more units, run emit-wave-dispatch.mjs next"), or make the wave emitter treat an ok:true count:0 plan as a zero-wave success (payload with empty waves/calls) so the zero-unit pending path is not a false refusal. Add a regression test for the zero-unit plan either way.

docs/decisions/0081-pi-gate-fanout-one-runs-all-wave.md
```diff
@@ -0,0 +1,25 @@
...
+
+We rejected emitting N blocking calls with the conductor composing the join order (the shape Pi rejects, and the observed failure); emitting `tasks: [...]` (not available in this version); making Claude Code / Codex dispatch code-driven (their delivery stays agent-authored per the per-harness delivery table, so this decision changes no other harness's dispatch path); and adding gate-time detection of a conductor that bypasses the emitter entirely — the emitter-scoped refusal is what this record pins, and a bypass remains undetected, recorded as an explicit non-goal rather than an implied guarantee.
+
+## Consequences
+
+A multi-unit Pi round releases concurrently: on its own dogfood round, 6 dispatch units went out as two waves of 3 released by two calls, with each wave's reviewers landing their artifacts in parallel. Wall-clock stops scaling linearly with unit count, which removes the parent-deadline class that motivated the issue.
```
mfittko: **question** (`holistic`): Definition-of-done requires "a live Pi gate round completes a multi-unit fan-out in parallel waves ... consolidating normally at fan-in", but the only evidence is the ADR's Consequences prose ("on its own dogfood round, 6 dispatch units went out as two waves of 3 released by two calls"). No artifact in this round records that run (no keyed wave-plan/emit-plan pair, no fan-in ledger referencing it), and a static review cannot verify it. Is there a recorded wave plan (or gate ledger) evidencing the live parallel-wave round, or is this DoD row asserted from memory only? — judge: act
Recommendation: Point the DoD row at the concrete recorded artifact (the &lt;gate>-&lt;headSha>.wave-plan.json plus the corresponding fan-in ledger) or mark the live-round row as not-yet-verified for this head.

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `d9b667a5048d777b697b8e417d184c40180b21a7`
**Verdict:** clean

**Findings summary:** 11 angles reviewed; 15 findings (see per-angle breakdown below).

**Inline findings:** 2 (🔵 1 question · 🟠 1 medium) — on the diff, see Files changed. Angles: correctness, holistic.

Body-only findings — no anchorable changed line, so carried in full here (plain list, angle in brackets, `file:line` linked to the blob when known):

- 🟠 medium — The seam is documented but not wired into the conductor's operative runbook, so the change is not reachable by following the documented steps. Phase 2 step 4 still reads "Dispatch one collectable fresh-context review agent per emitted unit using its promptPath bytes verbatim" and enumerates only GATE-EXEC-FANOUT-DISPATCH-EMIT and GATE-EXEC-BRIEFING-PREFIX as the rules it follows — it never instructs running scripts/github/emit-wave-dispatch.mjs, and it does not list the new GATE-EXEC-FANOUT-WAVE-DISPATCH rule. The owner contract carries the same contradiction at skills/docs/gate-review-sub-loop-contract.md:257 ("Dispatch one independent, fresh-context review agent per emitted unit via the plain Agent tool") and :383 ("the conductor then dispatches one fresh-context review subagent per emitted unit"), with the Pi one-call clause appended only afterwards. A conductor executing step 4 literally reproduces exactly the per-unit serialization this PR exists to remove; the emitted wave body is only produced if the conductor independently notices the rule and runs the new script. Fix: add an explicit wave step to step 4 (run emit-wave-dispatch.mjs at the same key, then issue ONE subagent call per emitted wave) and list GATE-EXEC-FANOUT-WAVE-DISPATCH among the rules step 4 follows, and reconcile the two "one per emitted unit" sentences. [`skills/copilot-pr-followup/SKILL.md:415`](https://github.com/mfittko/dev-loops/blob/d9b667a5048d777b697b8e417d184c40180b21a7/skills/copilot-pr-followup/SKILL.md#L415) — _`deferred`_ — judge: _`act`_ _(holistic)_

**Clean (5):** config-drift, determinism, gate-evidence, link-check, scope

<details>
<summary>Suppressed low/nit findings (12) — below the inline severity floor</summary>


- `scripts/github/emit-wave-dispatch.mjs:535` **low** (`contract-surface`): maxConcurrentSource misreports the origin of the emitted maxConcurrent when --sequential overrides the recorded concurrency to 1. The USAGE documents maxConcurrentSource: "emit-plan|config" as the source of the reported maxConcurrent (lines 75, 583-584), but the override at line 535 runs AFTER maxConcurrentSource is set at lines 515/518, so the success payload can emit { maxConcurrent: 1, maxConcurrentSource: "emit-plan" } while the emit-plan recorded maxConcurrent 3. This is reachable whenever gates.requireFanoutEvidence is off and --sequential is passed (the refusal path is skipped): a coordinator reading the keyed wave plan then sees a concurrency attributed to the emit-plan, which did not produce it — only sequentialSource: "explicit-request" (line 542) disambiguates. — judge: act

- `scripts/github/emit-wave-dispatch.mjs:535` **low** (`correctness`): After the --sequential override sets maxConcurrent = 1, maxConcurrentSource is not updated, so the persisted wave plan can report maxConcurrent: 1 with maxConcurrentSource: "emit-plan" while the emit-plan actually recorded a value > 1. Reachable when gates.requireFanoutEvidence is off and --sequential is passed (the test at test/github/emit-wave-dispatch.test.mjs "an explicit --sequential without the config flag" asserts maxConcurrent=1 and sequentialSource="explicit-request" but never checks maxConcurrentSource). The documented vocabulary is "emit-plan|config" (USAGE line 75), neither of which describes the override, so the two provenance fields contradict each other. — judge: act

- `scripts/github/emit-wave-dispatch.mjs:494` **low** (`correctness`): The wave emitter resolves the config root from cwd (loadConfig({ repoRoot: cwd })), whose default is resolveRepoRoot(process.cwd()) (line 307), while the producer it consumes resolves config from the raw shell cwd (emit-fanout-dispatch.mjs:610 loadDevLoopConfig({ repoRoot: process.cwd() })). When the two steps of one round run from a non-root cwd, the producer can record a maxConcurrent resolved against a different config root than the consumer re-resolves, tripping the recorded-vs-resolved equality refusal (line 511) on a legitimate round. — judge: act

- `test/github/emit-wave-dispatch.test.mjs:482` **low** (`coverage`): Test name claims the write failure happens "mid-round", but seedEmitPlan(tmpDir) defaults to 3 units at maxConcurrent 3, which partitions into exactly ONE wave — so writeScript throws on the first and only call. The genuinely multi-wave cleanup path (wave-1.js already written, wave-2 write fails, both must be cleared) is untested, and the name over-promises what is asserted. — judge: act

- `scripts/github/emit-wave-dispatch.mjs:448` **low** (`coverage`): The non-ENOENT plan-read failure branch (corrupt/truncated emit-plan JSON → stderr + exit 2) has no test: the only plan-read test (test/github/emit-wave-dispatch.test.mjs:136) exercises the ENOENT refusal. A realistic interrupted-write plan is therefore unguarded against regression, and the documented "filesystem error → exit 2" contract is not pinned. — judge: act

- `scripts/github/emit-wave-dispatch.mjs:510` **low** (`coverage`): Only the ABSENT recorded maxConcurrent case is tested (test/github/emit-wave-dispatch.test.mjs:404 deletes the field). A malformed recorded value (0, -1, 2.5, or the string "3") silently takes the same fallback branch and re-partitions against the resolved config, which contradicts the doc's "a recorded value that disagrees refuses" posture — and no test pins that behavior either way. — judge: act

- `test/github/emit-wave-dispatch.test.mjs:742` **nit** (`coverage`): The empty-value usage loop enumerates --tmp-root/--cwd/--jq/--fields but omits --timeout-ms, so the timeoutArg === "" branch (scripts/github/emit-wave-dispatch.mjs:380) is untested even though it is the same input class the test exists to cover. — judge: act

- `scripts/github/emit-wave-dispatch.mjs:307` **low** (`holistic`): The USAGE claims the default --cwd is "the resolved repository root — the worktree root, via resolveRepoRoot — never the raw shell cwd" (line 61), but resolveDefaultCwd() returns resolveRepoRoot(process.cwd()), which only normalizes to the checkout the shell is already in. When the harness starts the command in the primary checkout (the exact GATE-EXEC-NO-CWD-DEPENDENCE scenario this file cites), the default is the PRIMARY checkout, not the round's linked worktree, and that value is written into the emitted subagent({ cwd }) call — dispatching every reviewer into the wrong tree. The config root (loadConfig({ repoRoot: cwd }), line 494) is likewise the wrong checkout. The --cwd flag exists as an override, but the default silently picks a wrong tree rather than refusing. Fix: refuse when --cwd is omitted and the resolved root is not the round's worktree, or correct the USAGE claim and require an explicit --cwd. — judge: act

- `scripts/github/emit-wave-dispatch.mjs:510` **low** (`input-validation`): A present-but-malformed recorded plan.maxConcurrent is silently ignored: Number.isInteger(plan.maxConcurrent) && plan.maxConcurrent >= 1 is false for a string ("3"), a float (2.5), or 0, so the else branch silently falls back to the live resolved concurrency and re-partitions the round against a config that may have changed. This is inconsistent with the same artifact's plan.count, which refuses fail-closed when malformed (!Number.isInteger(plan.count) at line 458), and it defeats the very drift guard the recorded-vs-resolved equality check exists for. — judge: act

- `scripts/github/emit-wave-dispatch.mjs:480` **low** (`input-validation`): unit.promptPath is taken verbatim from the emit-plan (only .trim()ed) and passed straight to readFile, with no confinement to the round's tmp/gate-context root. A traversal-like or foreign absolute path in the plan is accepted, and its bytes are inlined into a dispatchable workflow script. The plan is a local trusted artifact, so the exposure is limited, but this is the one plan field that reaches an arbitrary filesystem read and the round has a known root it could be checked against. — judge: reject

- `scripts/github/emit-wave-dispatch.mjs:252` **nit** (`input-validation`): Inconsistent normalization inside validateWaveDispatchPlan: the blank-key check trims (key.trim().length === 0, line 248) but the duplicate-key check compares the raw string (seen.has(key)), so " a" and "a " would be treated as distinct keys. The main() path pre-trims keys via waveDispatchKey, so this is latent for the CLI, but the exported validator is inconsistent for any other caller. — judge: act

- `docs/decisions/0081-pi-gate-fanout-one-runs-all-wave.md:1` **low** (`pr-description`): The PR body's ## In scope list names 7 artifacts but the diff touches 12 files. Unnamed: the new ADR docs/decisions/0081-pi-gate-fanout-one-runs-all-wave.md, the new test/contracts/gate-fanout-wave-dispatch-contract.test.mjs, and the modified test/github/emit-fanout-dispatch.test.mjs. The prior round (head 65654c2d) already raised the emit-fanout-dispatch test omission as a nit; the delta since then ADDED the ADR and the new contract test without naming either, so the scope contract under-enumerates the boundary by a wider margin than before. This is the delta instance of that finding. — judge: act
</details>


**Next action:** Draft gate clean at head d9b667a5. Proceed to Phase 4 fixer triage of the 14 judge-act findings, then re-gate at the new head and mark the PR ready for review.

scripts/github/emit-fanout-dispatch.mjs
mfittko: Addressed in eb6cf76. The wave emitter now treats a legitimately empty all-carried `--pending` plan (`{ ok: true, count: 0, units: [] }`) as a **zero-wave success** — it emits an empty `waves`/`calls` payload instead of the exit-1 refusal — while a plan whose recorded count is missing, malformed, or disagrees with its units still fails closed. The producer's usage guidance is also gated on `count > 0` and names the zero-unit path explicitly, so a conductor never follows the wave-step pointer into a false refusal. Regression coverage: `treats an ok:true count:0 plan as a zero-wave success` plus the `claims units it does not carry` / `records no valid count` refusal cases in test/github/emit-wave-dispatch.test.mjs.

docs/decisions/0081-pi-gate-fanout-one-runs-all-wave.md
mfittko: Answered in eb6cf76. No keyed `<gate>-<headSha>.wave-plan.json` / fan-in-ledger pair for the dogfood round is retained at this head, so the ADR's Consequences paragraph no longer asserts it as recorded evidence: it now states the 6-units-in-two-waves observation is a live-round observation not yet evidenced by a retained artifact pair, and that the DoD live-round row is **not-yet-verified** for this head until that pair is retained. No AC/DoD/Non-goal text was changed (docs/prose only, per the act).

scripts/github/emit-wave-dispatch.mjs
```diff
@@ -0,0 +1,665 @@
...
+
+  // Report WHY the round serialized, honestly: `sequential` mirrors the
+  // resolved `gates.fanout.sequential` config flag (never the derived
+  // maxConcurrent), while `sequentialSource` names the trigger.
+  let sequentialSource = null;
+  if (sequential === true) sequentialSource = "config-flag";
+  else if (requestedSequential) sequentialSource = "explicit-request";
+  else if (resolvedMaxConcurrent === 1) sequentialSource = "maxConcurrent:1";
+
+  const waves = zeroUnit ? [] : partitionWaves(keyed, maxConcurrent);
```
mfittko: **medium** (`correctness`): The wave partition has no primer exclusion, so on Pi the sanctioned path double-dispatches the round's designated primer reviewer. partitionWaves(keyed, maxConcurrent) partitions EVERY emit-plan unit, and keyed is built from all units; there is no primer marker anywhere (rg -ni primer scripts/github/emit-wave-dispatch.mjs and ... emit-fanout-dispatch.mjs both return zero hits). GATE-EXEC-PRIME's DEFAULT form is "One-reviewer-as-primer": "Dispatch one real reviewer first. It keeps its normal angle scope" — i.e. the primer IS one of the emitted units — and step 6 says "Release the remaining reviewers ... subject to Phase 2's concurrency bound". skills/copilot-pr-followup/SKILL.md:414 (Phase 1.5, MANDATORY) still requires that primer be dispatched first, while the new SKILL.md:415 text says on Pi to run emit-wave-dispatch.mjs and "issue ONE subagent call per emitted wave". A conductor obeying both must dispatch the primer standalone and then release wave 1, which re-contains the primer unit (the generated script is marked "do not edit by hand", so it cannot be dropped, and a runs.all wave starts all its items concurrently, so wave 1 itself cannot satisfy "dispatch one reviewer first"). Consequence: the round's primer angle is reviewed twice (the second run overwrites the same per-angle artifact path) and one of the wave's maxConcurrent slots is spent on the duplicate — the exact wall-clock/budget waste this PR exists to remove — while GATE-EXEC-PRIMER-EVIDENCE's "observed primer runs and reviewer runs" record becomes ambiguous about which run was the primer. Nothing in the new GATE-EXEC-FANOUT-WAVE-DISPATCH rule text (skills/docs/gate-review-sub-loop-contract.md:533) or in the SKILL reconciles the two. — judge: act
Recommendation: Make the emitter own the reconciliation: accept the primer unit's key (e.g. --primer-key &lt;key>) and exclude it from the partition, refusing (exit 1) when that key is absent from the plan or appears more than once — so wave 1 covers the remaining units and the primer keeps its own single-unit first call; or, if the intent is to leave the plan intact, state explicitly in GATE-EXEC-FANOUT-WAVE-DISPATCH and SKILL step 4 that the Pi form must be the "dedicated angle-less primer" (which is not an emitted unit) and that the default one-reviewer-as-primer form is not usable with the wave step. Pin the chosen behavior with a test (plan containing the primer key → wave 1 excludes it).

test/github/emit-wave-dispatch.test.mjs
```diff
@@ -0,0 +1,844 @@
...
+    // The legacy top-level shape this pi-subagents version rejects must never
+    // be emitted.
+    assert.equal(waveScriptCallShape(script).hasLegacyTasksInput, false);
+    // Prompt bytes are inlined verbatim, so the conductor never handles them.
+    assert.match(script, /PROMPT BYTES for pre-approval-gate-holistic/);
+    // The ready call body is the verified one-call shape.
+    assert.equal(plan.calls[0].async, false);
+    assert.equal(plan.calls[0].timeoutMs, DEFAULT_WAVE_TIMEOUT_MS);
+    assert.equal(plan.calls[0].workflowScriptPath, plan.waves[0].scriptPath);
+    assert.equal(typeof plan.calls[0].cwd, "string");
```
mfittko: **medium** (`coverage`): The default emitted-call cwd — the GATE-EXEC-NO-CWD-DEPENDENCE invariant this head explicitly re-derived (resolveDefaultCwd(tmpRoot) = the round's artifact root, scripts/github/emit-wave-dispatch.mjs:315,389) — is effectively unpinned: the only assertions are typeof plan.calls&#91;0].cwd === "string" and .length > 0. A regression back to process.cwd() (or any absolute path, including "/") keeps the suite green, and the consequence is reviewers silently dispatched into the shell's checkout instead of the round's worktree — contaminated/foreign findings, the exact failure the commit says it fixes. No test in the file asserts the default cwd's VALUE; only the explicit --cwd path is value-asserted (line 192). — judge: act
Recommendation: In the existing happy-path test, pin the default: assert.equal(plan.calls&#91;0].cwd, path.resolve(tmpDir)) (the fixture tmp dir is not a git checkout, so resolveRepoRoot(tmpDir) deterministically returns tmpDir). Add a second case asserting the default is NOT process.cwd() when the shell sits elsewhere (run main() with the round's --tmp-root under a different tree than the process cwd).

scripts/github/emit-wave-dispatch.mjs
```diff
@@ -0,0 +1,665 @@
...
+  // parallel execution was impractical. Any OTHER serialization silently
+  // degrades a gate that requires fan-out evidence, so it fails closed here
+  // instead of producing single-unit waves nobody justified. That includes the
+  // resolved concurrency collapsing to 1 for a reason other than the recorded
+  // flag (`gates.fanout.maxConcurrent: 1` is a valid config value and would
+  // otherwise serialize every round unnoticed), and an explicit --sequential
+  // request the config does not back.
+  const requestedSequential = argv.includes("--sequential");
+  const unjustifiedSerialization = maxConcurrent === 1 && !sequential;
+  if (!zeroUnit && requireFanoutEvidence && (unjustifiedSerialization || (requestedSequential && !sequential))) {
```
mfittko: **medium** (`holistic`): The new sanctioned Pi seam cannot express GATE-EXEC-DISPATCH-RETRY-BACKOFF's documented degradation. The emitter's only concurrency inputs are the emit-plan's recorded value and the freshly resolved config: it refuses (exit 1) when they disagree (emit-wave-dispatch.mjs:527-551) and refuses again when the resolved concurrency is 1 without gates.fanout.sequential (line 556). There is no --max-concurrent/backoffMaxConcurrent override anywhere in USAGE or argv parsing. For this repo resolveFanoutMaxConcurrent = 3, so backoffMaxConcurrent(3) (packages/core/src/loop/gate-fanin.mjs:66-69) = 1 — i.e. the contract's documented 429/5xx step 'halve the batch with backoffMaxConcurrent and recompute waves' (skills/docs/gate-review-sub-loop-contract.md:261, restated at :609-612) is both inexpressible through the wave emitter and actively refused by the new fail-closed serialization guard. A conductor that hits a 429 in a wave has no sanctioned way to retry at reduced concurrency and must either bypass the emitter (undetected, per the rule's own open gap) or abort the round. — judge: act
Recommendation: Add a bounded --max-concurrent &lt;n> override (n &lt;= the emit-plan's recorded value, and recorded in the wave plan as an explicit, justified degradation), or state in GATE-EXEC-FANOUT-WAVE-DISPATCH exactly how the backoff path re-partitions (e.g. re-run the emit step at the reduced effective concurrency so the recorded and resolved values agree).

scripts/github/emit-wave-dispatch.mjs
mfittko: **medium** (`holistic`): GATE-EXEC-PRIME's primer-then-release barrier is unreconciled with the wave seam. The default primer form is one-reviewer-as-primer: one real reviewer is dispatched first and 'the remaining reviewers' are released after its prefix write lands (skills/docs/gate-review-sub-loop-contract.md:199, :211). That primer IS one of the emitted dispatch units, but emit-wave-dispatch.mjs partitions EVERY keyed unit into waves (line 575) with no primer/exclusion concept anywhere in the module (no --exclude/--primer flag in USAGE, no primer handling). A conductor following skills/copilot-pr-followup/SKILL.md step 3 (line 414) then step 4 (line 415) therefore either (a) re-dispatches the primer unit inside wave 1 — two concurrent reviewers on the same per-angle artifact path and the same verify-fresh-review-context.mjs --scope &lt;gate>-&lt;angle> sentinel — or (b) drops the mandatory barrier by treating wave 1 as the primer release. The contract's own Phase 2 text (lines 254-261) and the new GATE-EXEC-FANOUT-WAVE-DISPATCH rule (line 533+) never mention the primer, and the DoD row explicitly requires 'preserving primer-then-release ordering (GATE-EXEC-PRIME)'. — judge: act
Recommendation: Reconcile the two rules explicitly: either make the dedicated angle-less &lt;gate>-prime primer the sanctioned Pi form (it is not an emitted unit, so the wave needs no exclusion), or give the emitter the primer unit's key to exclude and document the resulting flow (dispatch primer -> await its artifact -> re-run the emit step with --pending -> wave the remainder) in the wave rule and in SKILL step 4.

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `eb6cf76dd448caed09131ae706a8b7614823eb7f`
**Verdict:** clean

**Findings summary:** 19 findings at eb6cf76d: high 0, medium 4, low 12, nit 3 (all non-blocking; blockCleanOnFindingSeverities=[high]). The fan-in --out was withheld over the gate-comment render budget; the full per-angle findings live in the durable ledger (--findings-ledger) and the per-angle artifacts under tmp/gate-reviews/mfittko-dev-loops/pr-2352/draft_gate-eb6cf76d/.

**Inline findings:** 4 (🟠 4 medium) — on the diff, see Files changed. Angles: correctness, coverage, holistic.

<details>
<summary>Suppressed low/nit findings (15) — below the inline severity floor</summary>


- `scripts/github/emit-wave-dispatch.mjs:510` **low** (`config-drift`): Config-root anchor drift between the emit step and the wave step: emit-fanout-dispatch.mjs:618 resolves the round's config (and therefore the maxConcurrent it records in the emit-plan) against resolveRepoRoot(process.cwd()), while emit-wave-dispatch.mjs resolves the same config against cwd, whose default is resolveRepoRoot(path.dirname(tmpRoot)) (lines 389/315). Whenever the conductor follows the wave emitter's OWN documented remedy for a primary-checkout shell — --tmp-root &lt;worktree>/tmp (USAGE lines 68-76, GATE-EXEC-NO-CWD-DEPENDENCE) — the two anchors are DIFFERENT checkouts. If their .devloops fan-out config differs (e.g. the PR under review edits gates.fanout, or the primary checkout carries a local override), the recorded-vs-resolved guard refuses at line 530 (records maxConcurrent N but the resolved effective fan-out concurrency is M) with a remediation ('re-run emit-fanout-dispatch.mjs') that cannot change the recorded value, because the emitter ignores --tmp-root when choosing its config root. The PR's own comment at emit-fanout-dispatch.mjs:613-617 claims the two anchors coincide ('the checkout root the wave emitter's recorded-vs-resolved guard also resolves against'), which holds only when --tmp-root's parent equals the process cwd. — judge: act

- `docs/decisions/0081-pi-gate-fanout-one-runs-all-wave.md:5` **low** (`config-drift`): ADR-STATUS-VALUES drift: the record's acceptance evidence does not correspond to an accepting event. Status is Accepted — 2026-09-21 (&#91;issue 2350](https://github.com/mfittko/dev-loops/issues/2350)), but (a) the record's own commit is 2026-09-22 02:41 +0200 (00:41 UTC) — the round clock is 2026-09-22 — so the stated acceptance date precedes the record's existence by a day, and (b) 2350 is still OPEN (this PR closes it), so the linked artifact is not a merged PR / closed issue / commit. Sibling records follow the convention this one breaks: 0079 and 0080 both date to their own commit day and link the merged PR (&#91;PR 2319], &#91;PR 2326]). ADR-STATUS-VALUES requires the date the decision was accepted and a link to the accepting event, and permits Proposed only while the authoring PR is under review. — judge: act

- `scripts/github/emit-wave-dispatch.mjs:98` **low** (`contract-surface`): Documented exit-code contract disagrees with the implementation for the zero-unit round. The --help exit-1 list states the plan carries no units is a refusal, and the exit-0 line says only Emitted one ready wave script + call body per wave; but the implementation treats { ok: true, count: 0, units: &#91;] } as a zero-wave SUCCESS: the plan?.ok !== true guard (line 464) passes, zeroUnit (line 479) short-circuits the serialization guard (line 556), the shape validator (line 586) and the partition (line 575), so the run emits { ok: true, count: 0, waves: &#91;], calls: &#91;] } and exits 0. The owner rule GATE-EXEC-FANOUT-WAVE-DISPATCH in skills/docs/gate-review-sub-loop-contract.md also never states the zero-unit success — it only enumerates refusals — so the sanctioned all-carried round's wave-step behavior is documented ONLY in the sibling emitter's usage (A zero-unit all-carried pending round ({ ok: true, count: 0 }) dispatches nothing, so it needs no wave step). A conductor reading this CLI's own usage cannot distinguish the success from the documented refusal. — judge: act

- `scripts/github/emit-fanout-dispatch.mjs:618` **low** (`coverage`): Behavior change with zero coverage: this head changed the config root from process.cwd() to resolveRepoRoot(process.cwd()) so a non-root cwd normalizes to the checkout root the wave emitter's recorded-vs-resolved guard also resolves against. No test in test/github/emit-fanout-dispatch.test.mjs (or anywhere else) invokes the emitter from a subdirectory cwd — every runEmitCli(...) call passes cwd: repoRoot, and no test uses process.chdir. The subdir-cwd config divergence the change fixes (loadDevLoopConfig reads &lt;repoRoot>/.devloops with no upward walk, so a subdir cwd falls back to defaults and a later wave step refuses on the recorded-vs-resolved maxConcurrent mismatch) is therefore unverified. — judge: act

- `test/contracts/gate-fanout-wave-dispatch-contract.test.mjs:46` **low** (`coverage`): AC-6 names TWO surfaces — the skills/dev-loop/SKILL.md dispatch guidance AND the per-harness delivery table in skills/docs/gate-review-sub-loop-contract.md. The contract suite pins the rule's owner section and the SKILL.md paragraph, but nothing asserts the delivery-table row (skills/docs/gate-review-sub-loop-contract.md:463, "Code-driven Pi runs.all"). rg 'Code-driven' test/ returns no hits, so reverting that row to the old "The driver reads the file and directly supplies the spawned reviewer's prompt" wording passes every test in the suite — the half of AC-6 the PR itself calls out as the misreading this change fixes. — judge: act

- `scripts/github/emit-wave-dispatch.mjs:463` **low** (`coverage`): The plan?.ok !== true refusal branch is untested: every fixture in test/github/emit-wave-dispatch.test.mjs seeds ok: true, so no test covers a refusal-shaped emit-plan (ok: false with a matching count/units pair). Dropping that branch would still pass the count guards and dispatch a round whose sibling emitter recorded a refusal — a fail-open on the plan-validity contract. — judge: act

- `scripts/github/emit-wave-dispatch.mjs:556` **nit** (`coverage`): The !zeroUnit carve-out in the serialization guard is untested: no case exercises a zero-unit round combined with --sequential (or a config-level concurrency of 1) under requireFanoutEvidence: true, which is the one path where the guard is deliberately skipped. A regression dropping !zeroUnit would refuse an all-carried round. — judge: act

- `test/github/emit-wave-dispatch.test.mjs:817` **nit** (`coverage`): Test name grammar: "empty-value usage exits for --tmp-root, --cwd, --timeout-ms, --jq, and --fields exit 2" carries two verbs ("exits ... exit 2"); the name is otherwise accurate. — judge: act

- `scripts/github/emit-fanout-dispatch.mjs:618` **low** (`determinism`): The round's effective fan-out concurrency is anchored to two different config roots depending on the step, so the same round artifacts can resolve different maxConcurrent values and fail a legitimate round closed at the wave emitter's recorded-vs-resolved guard. The emit step now resolves config via resolveRepoRoot(process.cwd()) (this line), whose comment claims it aligns with "the checkout root the wave emitter's recorded-vs-resolved guard also resolves against". It does not: the wave emitter resolves config against its emitted-call cwd, which defaults to resolveRepoRoot(path.dirname(--tmp-root)) (scripts/github/emit-wave-dispatch.mjs:315/389) and is fed to loadConfig({ repoRoot: cwd }) at scripts/github/emit-wave-dispatch.mjs:510. The briefing's own warning is the trigger: the shell may sit in the primary checkout while --tmp-root points at the linked worktree. Then the emit step reads .devloops from the primary checkout's git root while the wave step reads it from the worktree, and any gates.fanout.maxConcurrent/sequential difference between the two checkouts makes plan.maxConcurrent !== resolvedMaxConcurrent and refuses exit 1 (scripts/github/emit-wave-dispatch.mjs:553). The emit step has no --cwd/--tmp-root-derived config anchor to fix this from the caller side, so the round's concurrency is not a pure function of the round's artifacts. — judge: act

- `test/github/emit-wave-dispatch.test.mjs:243` **low** (`determinism`): The cross-script integration test is not hermetic w.r.t. the ambient repo fan-out config, so it passes/fails on .devloops values it does not control. It drives the REAL emitFanoutMain (this line) with no config override (the emitter exposes no loadConfig seam), so emitted.maxConcurrent comes from the actual checkout's .devloops; the wave emitter is then stubbed at line 258 with configStub({ maxConcurrent: emitted.maxConcurrent }), whose defaults pin sequential: false and requireFanoutEvidence: true. If the ambient repo ever configures gates.fanout.sequential: true, the real emitter records maxConcurrent: 1 while the stub reports sequential: false, so the wave emitter's unjustified-serialization guard refuses exit 1 and the test's assert.equal(..., 0) fails on a legitimate config rather than on a code defect. Every other in-process test stubs config; only this one couples to environment. — judge: act

- `docs/decisions/0081-pi-gate-fanout-one-runs-all-wave.md:21` **low** (`holistic`): The DoD row 'A live Pi gate round completes a multi-unit fan-out in parallel waves, preserving primer-then-release ordering (GATE-EXEC-PRIME) and consolidating normally at fan-in' is not satisfied at this head. The ADR discloses this honestly ('not-yet-verified for the current head until that artifact pair is retained'), and this draft_gate round cannot supply it because the round was not dispatched through the new emitter. A live multi-unit wave plus its keyed &lt;gate>-&lt;headSha>.wave-plan.json and fan-in ledger is the only evidence that the emitted runs.all shape actually works under pi-subagents; until then the seam's central claim rests on the issue's dogfood observation alone. — judge: reject

- `scripts/github/emit-wave-dispatch.mjs:66` **low** (`holistic`): The USAGE --cwd text overstates the safety property: it says the default is 'the round's artifact root ... never the raw shell cwd', but resolveDefaultCwd derives that root from --tmp-root, whose own default is path.join(process.cwd(), "tmp"). With --tmp-root omitted (as the runbook's abbreviated invocations at skills/copilot-followup SKILL step 4 show) the emitted cwd is exactly the shell cwd's checkout resolved to its git toplevel. If the shell sits in the primary checkout while the round belongs to the PR worktree, the reviewers are dispatched into the primary checkout — the GATE-EXEC-NO-CWD-DEPENDENCE hazard the comment claims to prevent, and the tree the Reviewer source-read invariant forbids citing. In practice the round's gate-context is written via resolveGateArtifactTmpRoot (main-worktree tmp) while the emit step reads --tmp-root, so the mismatch normally surfaces earlier as a missing plan rather than as a silently wrong reviewer cwd. — judge: reject

- `scripts/github/emit-wave-dispatch.mjs:389` **low** (`input-validation`): --cwd — a new, USAGE-advertised option — is resolved but never validated for existence, and it is simultaneously the emitted child cwd and the config root the concurrency is resolved against: const cwd = path.resolve(cwdArg ?? resolveDefaultCwd(tmpRoot)) is passed straight to loadConfig({ repoRoot: cwd }) (line 525) and into calls&#91;].cwd. A nonexistent --cwd therefore silently resolves BUILT_IN_DEFAULTS: loadDevLoopConfig treats a missing &lt;repoRoot>/.devloops as a warning, not an error (packages/core/src/config/config.mjs:1539, warnOnMissing: true). Two consequences: (a) when the repo's configured value differs from the shipped default, the guard at line 533 fires with a message that blames the emit-plan and instructs "re-run emit-fanout-dispatch.mjs" — a remediation that cannot change the resolved root, since the wave step still reads --cwd, so the operator loops; (b) when the repo's configured value equals the shipped default (4), no guard fires and the emitter reports ok: true with calls&#91;].cwd pointing at a directory that does not exist, i.e. it emits a "ready" dispatch body that cannot launch. The sibling --tmp-root at least fails loudly (ENOENT on the plan); --cwd has no such backstop. — judge: act

- `scripts/github/emit-wave-dispatch.mjs:462` **low** (`input-validation`): The emit-plan's identity fields are never cross-checked against the requested round key, even though the same guard block already cross-checks count (line 469), count vs units.length (line 474) and maxConcurrent vs the resolved value (line 533). The emitter writes gate, headSha, repo and pr into the plan, and this script ignores all four: a plan carrying a different headSha/gate at this key would dispatch with exit 0, inlining promptPath bytes composed for that other head (prompt paths are keyed &lt;gate>-&lt;headSha>.dispatch-prompt-&lt;scope>.txt), so reviewers would run against the wrong round's briefing prefix while the round is recorded as this head's. Reachability is low (the sibling emitter is the only writer and writes the same key), but this is the same "is this plan really THIS round's?" guard family the block already implements for count/maxConcurrent, so the omission is an inconsistent coverage gap rather than a deliberate carve-out. — judge: reject

- `scripts/github/emit-fanout-dispatch.mjs:618` **nit** (`pr-description`): The PR body's ## In scope bullet for scripts/github/emit-fanout-dispatch.mjs enumerates exactly two changes ("mandatory artifact-write clause added to the emitted reviewer prompt suffix; usage text names the wave step"), but the diff to that file also carries a third, behaviour-relevant change: loadDevLoopConfig({ repoRoot: resolveRepoRoot(process.cwd()) }) replaces the raw process.cwd() config root (plus the matching import and its 4-line rationale comment). That normalization decides which .devloops the emitter reads, and .devloops is read at exactly &lt;repoRoot>/.devloops with no upward walk (scripts/loop/_repo-root-resolver.mjs), so from a subdir cwd the emitter previously resolved the fallback defaults and would now resolve the repo's configured gates.fanout.maxConcurrent — the same value the wave emitter's recorded-vs-resolved guard compares against. A reviewer reading the scope contract cannot see that the emitter's concurrency resolution changed. — judge: reject
</details>

**Next action:** Draft gate clean at head eb6cf76d (round 2). Proceed to Phase 4 fixer triage of the 15 judge-act findings, then re-gate at the new head and mark the PR ready for review.

scripts/github/emit-wave-dispatch.mjs
mfittko: Addressed in c7407deb. The emitter now owns the primer reconciliation instead of leaving it to the conductor: `--primer-key <key>` excludes the one-reviewer-as-primer unit from the remaining partition and emits it as its OWN single-unit FIRST wave, so awaiting `calls[0]` IS the GATE-EXEC-PRIME barrier and the remaining waves release after it — the primer is never re-dispatched inside wave 1, so no per-angle artifact path is double-written and no `maxConcurrent` slot is spent on a duplicate. It refuses (exit 1) when the key is absent from the emit-plan or matches more than one unit, and omitting the flag remains correct for the dedicated angle-less `<gate>-prime` primer (not an emitted unit). Documented in the `GATE-EXEC-FANOUT-WAVE-DISPATCH` owner rule, the dev-loop dispatch guidance, and copilot-pr-followup step 4, and pinned by `--primer-key emits the primer as its own first wave and excludes it from the rest` plus a refusal test.

test/github/emit-wave-dispatch.test.mjs
mfittko: Addressed in c7407deb. The happy-path test now pins the default emitted-call `cwd` by VALUE, not just its type: `assert.equal(plan.calls[0].cwd, path.resolve(tmpDir))`. The fixture tmp dir is not a git checkout, so `resolveRepoRoot(tmpDir)` deterministically returns tmpDir — a regression back to `process.cwd()` (or any other absolute path) now fails the suite. The explicit `--cwd` path keeps its own value assertion, and a new `--cwd` existence test covers the CLI boundary.

scripts/github/emit-wave-dispatch.mjs
mfittko: Addressed in c7407deb. The seam now expresses the documented halved-batch degradation: `--max-concurrent <n>` re-partitions this run's waves at most `n` units wide, bounded to `n <= the emit-plan's recorded maxConcurrent` (refusing exit 1 above it) and recorded in the wave plan as `maxConcurrentSource: "explicit-degradation"`. A `--max-concurrent 1` is therefore a recorded, justified degradation and no longer trips the fail-closed serialization guard, so a conductor that hits an exhausted 429/5xx retry can reduce concurrency through the sanctioned emitter instead of bypassing it or aborting. The resolved concurrency DEFAULT and its resolution order are unchanged (ng:1). Documented in USAGE and the `GATE-EXEC-FANOUT-WAVE-DISPATCH` rule; pinned by three tests (reduce, degrade-to-1, over-bound refusal).

scripts/github/emit-wave-dispatch.mjs
mfittko: Addressed in c7407deb. The primer-then-release barrier is now reconciled inside the seam rather than by prose: with `--primer-key <key>` the emitter emits the one-reviewer-as-primer unit as its OWN single-unit first wave (`calls[0]`), so awaiting it IS the barrier, and the remaining waves are partitioned and released only after it — never with the primer re-included in wave 1. This preserves the mandatory GATE-EXEC-PRIME ordering while keeping the conductor out of the prompt bytes. Documented in the `GATE-EXEC-FANOUT-WAVE-DISPATCH` owner rule and SKILL step 4, and pinned by the primer-wave test.

mfittko:
Superseded by a docs-only replacement (operator decision).

Why close rather than merge:
- The emitter's enforcement was **advisory only**. The repo's own gate contract states a conductor that bypasses `emit-wave-dispatch.mjs` and composes its own per-unit calls is NOT detected at gate time. So the fail-closed guard fired only on the already-compliant path, while the incident path (the serialization this issue records) stayed open.
- The new file's marginal contribution was rendering units into a `.js` file. `emit-fanout-dispatch.mjs` already resolves `maxConcurrent` (`resolveFanoutEffectiveConcurrency`), computes the wave partition, and emits verbatim prompt bytes per unit.
- Cost/benefit: three 11-angle gate rounds, each fixer push re-gating.

The honest minimal fix is doc-clarity only: name the one-call dispatch shape in the contract prose so "blocking joins" can no longer be read as N separate blocking calls.

Branch `issue-2350` is intentionally left in place for the record (not deleted).

PR #2360, opened 2026-09-22
chore(config): resolve the Pi pre-PR reviewer model to the Codex child model

mfittko:
## Objective

The Pi harness's pre-PR reviewer currently resolves to no model override, so the
pre-PR review pass runs on whatever model the session happens to use instead of
the strong reviewer model the Claude-Code harness already uses. The
pre-PR review contract (`skills/docs/pre-pr-review-contract.md`, rule
`PRE-PR-MODEL-CONFIG-RESOLVED`) requires that model to come from
`resolveRoleModel(config, { role: "pre-PR-reviewer", harness })` and forbids a
model literal in the phase's prose, tooling, or dispatch. `.devloops` is
therefore the only correct home for the opt-in.

## In scope

- `.devloops`: `models.tiers.pre-pr-strong` gains `pi: openai-codex/gpt-5.6-sol`.
  The Claude key keeps `fable`, and `models.roleTiers.pre-PR-reviewer` stays
  `pre-pr-strong`.
- `.devloops`: the adjacent rationale comment block is corrected. It previously
  claimed the Pi harness "has no entry here and resolves to null
  (inherit/default)", which the change makes false.
- `test/contracts/devloops-tier-config.test.mjs`: the drift guard that pinned the
  old `null`-on-Pi resolution is updated to pin the new value, with its title and
  comment reworded to match.
- `test/contracts/pre-pr-review-contract.test.mjs`: the prose guard that enforces
  "no model literal in the pre-PR phase's prose" is extended to also reject a
  resolved pre-PR reviewer model token. Its existing patterns match only
  Claude-family tokens, so without this the newly opted-in Pi token would be
  unguarded. The check is an exact comparison against the tokens this repo
  resolves, not a guessed `provider/model` shape: a shape pattern cannot tell a
  model id from an ordinary `dir/name-with-digit` path reference.

## Acceptance criteria

- [x] `resolveRoleModel(config, { role: "pre-PR-reviewer", harness: "pi" })` returns `openai-codex/gpt-5.6-sol` against this repo's `.devloops`.
- [x] `resolveRoleModel(config, { role: "pre-PR-reviewer", harness: "claude" })` still returns `fable`.
- [x] `loadDevLoopConfig` reports zero errors for the edited `.devloops`, so the tier keeps the sanctioned `models.tiers.<alias>.<harness>` shape.
- [x] The `.devloops` comment block no longer claims the Pi harness resolves to `null`, and its remaining rationale (why `fable` is the only accepted Claude token, why the model is never hardcoded) is intact.
- [x] The pre-PR phase prose guard rejects a resolved pre-PR reviewer model token, so a future hardcode of `openai-codex/gpt-5.6-sol` into `skills/docs/pre-pr-review-contract.md` or the local-implementation SKILL's pre-PR step fails the guard instead of passing silently.
- [x] The extended guard still admits the legitimate path references it scans alongside (it must not start failing on ordinary `dir/name` text), and a positive control proves the guard detects each resolved token rather than passing vacuously.

## Definition of done

- [x] `test/contracts/devloops-tier-config.test.mjs` pins both harness resolutions and passes.
- [x] `bun run verify` passes on the pushed head.
- [x] No model literal is introduced into the pre-PR phase's prose, tooling, or dispatch; the literals live only in `.devloops` (the config home) and in the contract tests that assert what that config resolves to.
- [x] `test/contracts/pre-pr-review-contract.test.mjs` passes with the resolved-token assertion and its positive control, proving the guard detects each resolved pre-PR reviewer token while still admitting the path references in the surfaces it scans.

## Scope note: the light-mode file cap

This PR was sanctioned as an issue-less `--lightweight` change under
`localImplementation.lightMode` (`maxFiles: 2`, `maxLines: 20`), and it adds a
third file, which exceeds that cap. The cap is a startup/create-time eligibility
gate that bounds review *depth*, and no depth discount is taken here: this diff
touches `test/contracts/**`, which is in the shipped risk-path denylist, so the
change is not eligible for any reduced tier or inline carve-out. The cap is also
never re-read after creation — `ready-for-review.mjs` and `pre-pr-ready-gate.mjs`
do not consult `maxFiles`, and the merge-gate `scopeUnderThreshold` re-verify
applies only to `inline_single_agent` verdicts. Exceeding the cap therefore does
not buy a lighter review, it buys a closer one.

The third file is in scope rather than deferred because the gap is one this diff
opens: opting into a new model id while relying on a guard that recognizes only
Claude-family ids would silently weaken the very rule the change exists to
satisfy (`PRE-PR-MODEL-CONFIG-RESOLVED`).

## Non-goals

- Not switching `models.roleTiers.pre-PR-reviewer` to the built-in `high` alias —
  that would also swap the Claude side from Fable to Opus.
- Not changing the Claude-harness token `fable` or its underlying model identity.
- Not changing any other tier, role mapping, gate, or the pre-PR contract prose.
- Not adding a Pi-specific doc for the opt-in; the contract already describes the
  mechanism generically for both harnesses.
- Not editing `CHANGELOG.md`; this change's commit type is `chore`, and the
  shipped entry for the pre-PR phase lives in the already-tagged
  `1.0.4-pre.0` section, which released notes are not rewritten.

## Open questions/risks

- The Pi-side value tracks the operator's Codex model identity. The config schema
  validates only that a harness value is a non-empty string, not that the Pi
  dispatch accepts it, so a token typo would surface at dispatch time rather than
  at config load.
- The literal now appears in two committed files. The contract test literal is
  judged acceptable because it asserts what the *config* resolves to, not what
  the phase dispatches — a drift guard that already pinned `"fable"` before this
  change.
- The extended prose guard compares against the tokens this repo resolves, so it
  guards whatever the opt-in currently names. It does not detect a model token
  that is hardcoded but not the repo's current opt-in (a stale token from an
  earlier config), which the Claude-family patterns cover only for the Claude
  vocabulary.
- The stale comment at `packages/core/test/config.test.mjs:6544` ("Mirrors this
  repo's .devloops opt-in: Fable on Claude only, Pi inherits.") is left as-is. The
  fixture deliberately omits the `pi` key to exercise the unset-harness path, its
  assertions stay correct, and the comment sits outside this PR's declared scope.

## Validation

- `bun run verify`: pass.


test/contracts/pre-pr-review-contract.test.mjs
```diff
@@ -99,11 +101,37 @@ test("no concrete model token is hardcoded in the phase prose or SKILL step (con
   const bareModel = /\b(fable|opus|sonnet|haiku)\b/i;
   const fullId = /claude-[a-z0-9]+-\d/i;
 
+  // A harness opt-in may resolve to a provider-qualified model id
+  // (`provider/model`). Assert against the model ids THIS repo actually resolves
+  // rather than a guessed id shape: a shape pattern cannot tell a model id from
+  // an ordinary `dir/name-with-digit` path reference, so it would both over-match
+  // real prose and miss a future id that does not fit the guess.
+  const { config, errors } = await loadDevLoopConfig({ repoRoot: process.cwd() });
```
mfittko: **medium** (`determinism`): The new resolved-token guard resolves its token set from process.cwd() instead of the file's own deterministic repo anchor, so under any cwd that is not this repo root the guard silently narrows to the built-in fallback token ('opus') and stops protecting the Pi token it was added for. Verified: loadDevLoopConfig({ repoRoot: &lt;dir without .devloops> }) returns errors: &#91;] and resolveRoleModel(pre-PR-reviewer) => &#91;'opus', null], so the resolvedModels.length > 0 sentinel at line 114-117 still passes (it is satisfied by built-in defaults via BUILTIN_ROLE_TIERS&#91;'pre-PR-reviewer']='high' -> BUILTIN_TIERS.high.claude='opus') and the positive control at line 122-127 passes against that substituted set. The failure is therefore silent, not loud: a hardcoded openai-codex/gpt-5.6-sol in the contract prose or SKILL pre-PR step would go undetected, which is exactly the gap this PR claims to close. The same test already computes the deterministic anchor two lines up (line 17 repoRoot via import.meta.url, used by readRepo for every file read), so the file mixes a stable anchor for file reads with a mutable global-process anchor for config resolution. Reachability: bun run verify from the repo root is unaffected; the degradation needs an invocation whose cwd differs (subdirectory, another checkout/worktree whose .devloops resolves different tokens, or a temp cwd left in-process by a sibling test file such as test/loop/consolidate-fanin.test.mjs, which does process.chdir into temp dirs). Note the repo treats cwd-dependent config resolution as a real hazard: test/loop/consolidate-fanin.test.mjs has a dedicated pin '--repo-root anchors config resolution regardless of process.cwd()'. — judge: act
Recommendation: Replace repoRoot: process.cwd() with the module-level repoRoot already computed at line 17 (the same anchor readRepo uses), and tighten the sentinel so a substituted built-in set cannot pass vacuously — e.g. assert the loaded config actually came from this repo (pin the expected claude/pi tokens or assert the alias pre-PR-reviewer maps to pre-pr-strong in config.models.roleTiers) before scanning the prose.

mfittko (commented):
### Gate review: `draft_gate`


**Reviewed head SHA:** `52c147133e15ef1fb97436f5e50f0858f7d71cf8`
**Verdict:** clean

**Findings summary:** 10 angles reviewed; 12 findings (see per-angle breakdown below).

**Inline findings:** 1 (🟠 1 medium) — on the diff, see Files changed. Angles: determinism.

Body-only findings — no anchorable changed line, so carried in full here (plain list, angle in brackets, `file:line` linked to the blob when known):

- 🔵 question — Unresolved visible-checkpoint-evidence question carried forward from the draft_gate round at head 4be2e924, which reported PR 2360 had 0 reviews, 0 issue comments and 0 review threads while a completed clean draft_gate round existed for head 606bc7a5 (main-anchored ledger tmp/gate-findings/mfittko-dev-loops/pr-2360/draft_gate-606bc7a532509d0601207cc50abb466af16321ad.json: verdict 'clean', executionMode 'fanout_fanin', loggedAt 2026-09-22T15:09:30Z). GATE-COMMENT-DRAFT-REQUIREMENTS (skills/docs/gate-review-comment-contract.md) requires a visible checkpoint verdict comment referencing the current head while the PR is draft, and scripts/github/detect-checkpoint-evidence.mjs fails closed on its absence, blocking the draft -> ready transition. This unit is bound not to poll PR state, so it can neither confirm nor refute the prior observation; the current round's fan-in is the write that is expected to satisfy it. Answer needed: is a visible draft_gate verdict comment posted for head 52c147133e15ef1fb97436f5e50f0858f7d71cf8 (and was the round-2 observation of zero comments at head 4be2e924 accurate)? — _`deferred`_ — judge: _`reject`_ _(gate-evidence)_

**Clean (2):** scope, contract-surface

<details>
<summary>Suppressed low/nit findings (10) — below the inline severity floor</summary>


- `test/contracts/pre-pr-review-contract.test.mjs:109` **low** (`correctness`): The new guard resolves the config under test from process.cwd() while the prose it scans comes from the import.meta.url-derived repoRoot (line 17). Invoked from any cwd that is not the repo root (e.g. cd test/contracts && bun test pre-pr-review-contract.test.mjs), loadDevLoopConfig finds no .devloops there and returns errors: &#91;] with built-in defaults, so resolvedModels silently degrades to &#91;"opus"]; the newly opted-in Pi token openai-codex/gpt-5.6-sol is then never compared against the contract or SKILL text and the guard passes while guarding nothing — exactly the Pi-side hardcode this change exists to prevent. All sanctioned invocations (bun run verify, scripts/run-bun-test.mjs spawning bun test without a cwd override) run from the repo root, so this is not operator-visible on the normal path; the sibling devloops-tier-config.test.mjs uses process.cwd() consistently for the whole file, whereas this file mixes the two anchors. — judge: act

- `test/contracts/pre-pr-review-contract.test.mjs:115` **low** (`coverage`): The new resolved-model guard's non-vacuity assertion (resolvedModels.length > 0, justified in-comment as 'for this guard to be meaningful') is satisfied by the built-in Claude fallback even when this repo's own .devloops opt-in is NOT loaded, so the guard silently degrades to guarding only the Claude token (opus) and the newly opted-in Pi token goes unguarded — the exact fail-open the PR exists to close. Evidence: line 109 loads config with repoRoot: process.cwd() while the same file already defines the authoritative repoRoot from import.meta.url (line 17, used by readRepo); with no .devloops in scope, resolveRoleModel(config, { role: 'pre-PR-reviewer', harness: 'claude' }) returns the built-in high-tier opus and harness: 'pi' returns null (BUILTIN_TIERS.high = { claude: 'opus', pi: null }, packages/core/src/config/config.mjs:34-37, pinned by packages/core/test/config.test.mjs:6555-6556), so resolvedModels is &#91;'opus'] and both the length assertion and the positive-control loop pass. The same degradation also occurs on any future config edit that drops the pi key from the tier, without any test failing. Nothing in the suite asserts that the guard's resolved-token set actually came from this repo's opt-in, so the 'proves the guard detects each resolved token' DoD row is contingent on ambient cwd + current config state. — judge: act

- `test/contracts/pre-pr-review-contract.test.mjs:134` **low** (`coverage`): Coverage gap against acceptance criterion 5 as literally worded ('a future hardcode of openai-codex/gpt-5.6-sol into skills/docs/pre-pr-review-contract.md ... fails the guard instead of passing silently'): the contract scan runs on contractRest (line 132 = contract with the PRE-PR-MODEL-CONFIG-RESOLVED paragraph removed), so the resolved-token assertion is skipped for exactly that paragraph. The exclusion's stated rationale — the paragraph 'MAY name the accepted-token vocabulary ... documenting the harness enum' — covers the Claude enum tokens (sonnet|opus|haiku|fable) it currently names (skills/docs/pre-pr-review-contract.md:93-95); the newly added provider-qualified Pi token has no such legitimate mention there, and that paragraph is the most likely place for a model literal to be introduced 'as documentation' while still being a hardcode of the phase's dispatch model. — judge: act

- `test/contracts/devloops-tier-config.test.mjs:43` **nit** (`coverage`): Test name quality: the renamed pin asserts the exact literal openai-codex/gpt-5.6-sol but names it 'the Codex child model' (title and the assertion message at line 55), so the name does not tell a reader which token is pinned or that it is a provider-qualified id — the failure output reads less informative than the assertion it guards. — judge: reject

- **low** (`gate-evidence`): The draft_gate round at head 4be2e924 produced consolidated findings but no final main-anchored ledger: tmp/gate-findings/ holds only a consolidation intermediate for that head (draft_gate-4be2e924....ledger.json, keys &#91;findings, overallVerdict], 11 findings, no repo/pr/gate/provenance/scopeDrift), while the main-anchored ledger exists only for head 606bc7a5. Per the findings-log anchoring rule the ledger must live at the main worktree so it survives worktree pruning and is readable at merge. This is consistent with an explained mid-flight supersession (the head moved to 52c14713 and the current round's carry-forward plan uses prevHead 606bc7a5, so the 4be2e924 round is retired and its findings are not silently lost). Flagged only so the missing ledger for a reviewed head reads as a recorded supersession rather than a missing-evidence gap. — judge: reject

- `test/contracts/pre-pr-review-contract.test.mjs:109` **low** (`holistic`): The new resolved-model guard derives its deny-list from loadDevLoopConfig({ repoRoot: process.cwd() }) while the same file already computes the checkout root as repoRoot (line 17) for every prose read. Because the guard is self-referential — it asserts only that the tokens *this* config resolves do not appear in the scanned prose — a cwd that is not the repo root makes it silently vacuous instead of failing: the load then yields the built-in tier (high -> opus on Claude, null on Pi; BUILTIN_TIERS/BUILTIN_ROLE_TIERS in packages/core/src/config/config.mjs), so resolvedModels = &#91;"opus"], the resolvedModels.length > 0 floor still passes, and the newly added openai-codex/gpt-5.6-sol protection (AC5/AC6) does not run at all. That is a fail-open degradation in the guard whose whole purpose is to catch a hardcoded resolved token. Reachability is off the sanctioned path only: scripts/verify.mjs and every test:* script spawn bun from the repo root (no cwd override in scripts/run-bun-test.mjs), so the shipped bun run verify runs this guard with cwd == repo root. Severity is low on that basis — real defect, no operator-visible consequence today. — judge: act

- `test/contracts/pre-pr-review-contract.test.mjs:118` **low** (`input-validation`): namesResolvedModel matches the config-resolved token case-sensitively (text.includes(model)), while every sibling check in the same test normalizes case: bareModel (line 101) is /i and fullId only lowercases the claude- prefix. A hardcoded token written in a different case (OpenAI-Codex/GPT-5.6-Sol, FABLE) is invisible to the new assertion and to bareModel/fullId (no fable|opus|sonnet|haiku word, no claude- prefix), so the guard normalizes the config token differently from the prose it scans and the newly added Pi token keeps a residual detection gap. Whitespace is not an issue: resolveTierMapping/trimmedOrNull already trims the resolved value before it reaches the predicate. — judge: reject

- **low** (`pr-description`): The body's 'Scope note: the light-mode file cap' paragraph carries gate-review process/verdict content, and its coverage claim no longer describes the round reviewing this head. It states the draft_gate 'resolved full_fanout / risk_path_touch and ran the complete untiered 20-angle pool'. That was true of the prior round at head 606bc7a5 (its ledger records 20 angles), but the current round resolved only 10 of the 20 candidate angles: tmp/gate-context/.../draft_gate-52c14713....json lists ci-guard, link-check, no-op, packaging-runtime, state-concurrency, pr-comments, contradiction-lens, code-conformance, semantic-drift and deslop as dropped ('detected categories (LOGIC_CHANGE, CONFIG_ONLY, TEST_ONLY) do not trigger this angle'). The paragraph is the body's justification for exceeding the sanctioned light-mode cap, so an overstated coverage claim inflates the evidence a reviewer reads when deciding whether the cap exception bought deeper review. Gate-dispatch mechanics and coverage counts belong in the gate verdict comment, not in the spec of record. — judge: reject

- **low** (`pr-description`): The '## Open questions/risks' section still carries a prior gate round's judge disposition: '... the draft_gate judge adjudicated the comment out of this PR's scope on the defer bar.' Judge/verdict status belongs in the gate verdict comment and the findings-log ledger (it is already delivered to reviewers as prior-round dispositions), not in the durable spec of record, which is the input the next reviewer reads before independently adjudicating. This is the same finding the pr-description angle raised against head 4be2e924; the sentence is unchanged at this head, so the finding is carried forward unresolved rather than newly introduced. — judge: reject

- `CHANGELOG.md:82` **nit** (`config-drift`): The ## 1.0.4-pre.0 entry (tagged v1.0.4-pre.0, unedited since the tag) states the pre-PR reviewer opt-in as 'this repo opts the Claude-Code harness into Fable ... while Pi inherits its default'. That is now false for the repo at HEAD: .devloops sets models.tiers.pre-pr-strong.pi = openai-codex/gpt-5.6-sol, so Pi no longer inherits. It is the only remaining surface that contradicts the new config. — judge: reject
</details>


**Next action:** Mark this PR ready for review, then apply the judge act list on the pre-PR review contract test and push after the ready flip.

test/contracts/pre-pr-review-contract.test.mjs
mfittko: Fixed in the working tree. The guard no longer derives its deny-list from `process.cwd()`: it now calls `loadDevLoopConfig({ repoRoot })`, the file-local `import.meta.url` anchor at line 17 that `readRepo` already uses for every prose read, so the resolved token set comes from this checkout rather than the ambient process directory.

The vacuity hole the finding names is closed as well — the guard now asserts that the loaded config actually carries this repo's opt-in (`config.models.tiers["pre-pr-strong"].pi`), so a substituted built-in fallback (`opus`/null) can no longer satisfy the `resolvedModels.length > 0` floor silently.

Verified locally: `bun test test/contracts/pre-pr-review-contract.test.mjs test/contracts/devloops-tier-config.test.mjs` → 7 pass, 0 fail.

The change lands as the next commit on this branch, pushed immediately after the ready-for-review flip; the draft-gate verdict is a one-time transition record, so that push does not invalidate it.

mfittko (commented):


**Reviewed head SHA:** `5262e5cb5dd636ad5c084a09039e9eb10ccc171b`
**Verdict:** clean
**Size-budget outcome:** pass
**Size-budget T1 slice:** not touched
**Size-budget waiver:** none

**Findings summary:** 6 angles reviewed; 8 findings (see per-angle breakdown below).

**Clean (2):** pr-checklist, renderer-security

<details>
<summary>Suppressed low/nit findings (8) — below the inline severity floor</summary>


- `test/contracts/pre-pr-review-contract.test.mjs:115` **low** (`acceptance-criteria`): The new non-vacuity assertion does not establish the property its comment claims, and it couples the contract test to one repo config key. It asserts config.models.tiers&#91;'pre-pr-strong'].pi is truthy with the message 'the guard must resolve this repo's .devloops opt-in, not a built-in fallback'. Verified against packages/core/src/config/config.mjs:1129-1144 (BUILTIN_ROLE_TIERS / BUILTIN_TIERS = { low: {claude:'sonnet',pi:null}, high: {claude:'opus',pi:null} }): with roleTiers&#91;'pre-PR-reviewer'] re-pointed to the built-in 'high' while the pre-pr-strong tier still carries its pi key, the assertion still passes while resolution falls back to opus/null — exactly the vacuity the comment says it prevents, and the Pi token then goes unguarded. Conversely, dropping only the pi key (a legitimate future config edit, e.g. reverting the Pi opt-in) fails this contract test with a non-vacuity message, duplicating the pin already owned by test/contracts/devloops-tier-config.test.mjs:54. — judge: reject

- `test/contracts/pre-pr-review-contract.test.mjs:161` **nit** (`acceptance-criteria`): assert.ok(step.length > 0, 'the SKILL pre-PR step scan surface must be non-empty') is not a real non-vacuity guard: when the LOCAL-PRE-PR-REVIEW-BEFORE-PUSH marker is absent, stepStart is -1 and skill.slice(-1, stepEnd) yields a single trailing character, which satisfies both this assert and the two token scans below. The practical hole is closed by the sibling test at lines 66-76 which asserts the marker index is found, so no reachable defect today. — judge: reject

- `packages/core/test/config.test.mjs:6544` **nit** (`contradiction-lens`): Stale claim now contradicted by the new config truth: the fixture comment says "Mirrors this repo's .devloops opt-in: Fable on Claude only, Pi inherits." and line 6549 repeats "Pi has no entry in the tier -> null (inherit/default), keeping it harness-agnostic". After this PR the repo's opt-in is Fable on Claude AND openai-codex/gpt-5.6-sol on Pi, so the 'Mirrors this repo' claim is false. Same drift in docs/decisions/0079-pre-pr-review-phase.md:15 ("This repo opts the Claude-Code harness into Fable" — now incomplete). The PR body discloses this and argues the fixture assertions stay correct (true: the fixture omits the pi key on purpose), so no behavior is wrong; only the mirrored-opt-in wording is. — judge: reject

- `test/contracts/pre-pr-review-contract.test.mjs:115` **low** (`holistic`): The new non-vacuity floor assert.ok(config?.models?.tiers?.&#91;"pre-pr-strong"]?.pi, "the guard must resolve this repo's .devloops opt-in, not a built-in fallback") conflates two independent conditions: config drift (does this repo's tier carry a pi value?) and guard vacuity (does the prose guard have any token to detect?). A schema-legal config where the Pi opt-in is explicitly null (ModelTierMapping permits null per harness: packages/core/src/config/config.mjs:97) still leaves the guard fully meaningful — resolvedModels keeps fable, and the line-122 resolvedModels.length > 0 floor already proves non-vacuity — yet this assertion fails with a message that misdescribes the cause ("not a built-in fallback"), so an operator legitimately reverting the Pi opt-in to null would get a prose-guard failure pointing at the wrong invariant. The config-value pin is already owned by test/contracts/devloops-tier-config.test.mjs:53, so this is a second, differently-worded failure site for one condition. — judge: reject

- `packages/core/test/config.test.mjs:6544` **nit** (`holistic`): The fixture comment // Mirrors this repo's .devloops opt-in: Fable on Claude only, Pi inherits. is now false for this repo (the tier sets pi: openai-codex/gpt-5.6-sol). The fixture's assertions stay correct because the synthetic config deliberately omits the pi key, so this is documentation staleness only — but it now contradicts the merged .devloops and will read as the repo's current policy to anyone opening the fixture. The PR declares it left as-is under open risks. — judge: reject

- `test/contracts/pre-pr-review-contract.test.mjs:115` **low** (`yagni`): The new non-vacuity assert on the raw config key config.models.tiers&#91;"pre-pr-strong"].pi is redundant with the pin that already exists in test/contracts/devloops-tier-config.test.mjs (resolveRoleModel(config, { role: "pre-PR-reviewer", harness: "pi" }) === "openai-codex/gpt-5.6-sol"), and it couples this contract guard to one tier alias name. Its stated purpose ("must resolve this repo's .devloops opt-in, not a built-in fallback") is already carried by the resolvedModels.length > 0 assert four lines below; the raw-key form adds no coverage but does add a false-failure path — a legitimate revert/remap of the opt-in (e.g. roleTiers pointed at the built-in high alias, which this PR explicitly leaves possible) drops models.tiers&#91;"pre-pr-strong"] and fails this contract test even though no model literal was hardcoded. — judge: reject

- `test/contracts/pre-pr-review-contract.test.mjs:151` **nit** (`yagni`): assert.ok(contractRest.length > 0, ...) and assert.ok(step.length > 0, ...) cannot fail on the surface they name: contractRest is contract minus one paragraph so it is never empty, and ruleParagraph already asserts the PRE-PR-MODEL-CONFIG-RESOLVED marker exists. The step variant is additionally weaker than it reads — when the SKILL marker is missing, slice(-1, stepEnd) yields either the last character (length 1) or an empty string depending on stepEnd, so it does not reliably detect the marker-absent case it appears to guard. — judge: reject

- `test/contracts/pre-pr-review-contract.test.mjs:145` **nit** (`yagni`): The Claude token vocabulary is now encoded twice in the same test: the pre-existing bareModel = /\b(fable|opus|sonnet|haiku)\b/i and the new inline !/^(sonnet|opus|haiku|fable)$/i filter. Two copies of the same enum drift independently. — judge: reject
</details>

**Next action:** Resolve the gate-authored finding threads, then merge under the standing authorization once the current-head evidence audit is green.

